PR #4, opened 2018-11-29
Backend crud

adamklepacz:
To start mongodb on localmachine you need to run 
`sudo mongod` in root directory of this project. 
It runs at `localhost:27017`

I've used Postman to test endpoints and MongoDB 
Compass app to check database entries.

Update:
This information are also included in Readme file.


src/server/controllers/item.controller.js
```diff
@@ -0,0 +1,36 @@
...
+// Create new item
+const itemCreate = function(req, resp) {
+  const item = new Item({
+    name: req.body.name,
+    isOrdered: req.body.isOrdered
+  });
+
+  item.save(err => {
+    if (err) return console.error(err);
+    return resp.send('Product created successfuly!');
```
marekrozmus: please set response status and for error also send response with proper status

src/server/controllers/item.controller.js
```diff
@@ -0,0 +1,36 @@
...
+  Item.findOneAndRemove(req.params.id, (err, product) => {
+    if (err) return console.error(err);
+    return resp.send(`${product.name} deleted succesfully!`);
+  });
+};
+
+module.exports = {
+  itemCreate,
+  getItemById,
+  deleteItemById
```
marekrozmus: please sort alphabetically

src/server/controllers/item.controller.js
```diff
@@ -0,0 +1,37 @@
...
+// Get item by given id
+const getItemById = function(req, resp) {
+  Item.findById(req.params.id, (err, product) => {
+    err ? resp.status(400).send(err.message) : resp.status(200).json(product);
+  });
+};
+
+// Delete item by given id
+const deleteItemById = function(req, resp) {
+  Item.find({ _id: req.params.id }).deleteOne((err, product) => {
```
marekrozmus: please check if findOneAndDelete or findByIdAndRemove could be used instead of calling two methods

src/server/controllers/items.controller.js
```diff
@@ -0,0 +1,28 @@
...
+
+// Get all the items
+const getAllItems = function(req, resp) {
+  Item.find({}, (err, items) =>
+    err ? resp.status(400).send(err.message) : resp.status(200).json(items)
+  );
+};
+
+// Get all ordered items
+const getOrderedItems = function(req, resp) {
```
marekrozmus: ordering/filtering should be parameter in getAllItems and both getOrderedItems and getAllUnordered should be removed

src/server/controllers/item.controller.js
```diff
@@ -0,0 +1,37 @@
+const Item = require('../models/item.model');
+
+// Create new item
+const itemCreate = function(req, resp) {
+  const item = new Item({
+    name: req.body.name,
```
jakub-bogacz: you can deconstruct req.body

src/server/controllers/item.controller.js
```diff
@@ -0,0 +1,37 @@
...
+    isOrdered: req.body.isOrdered
+  });
+
+  item.save(err => {
+    err
+      ? resp.status(400).send(err.message)
+      : resp.status(200).send('Product created successfuly!');
```
jakub-bogacz: You could return 201 Created status code (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201).
It would be good to set also Location header, and return newly created object in response (with new id).

src/server/controllers/item.controller.js
```diff
@@ -0,0 +1,37 @@
...
+    err ? resp.status(400).send(err.message) : resp.status(200).json(product);
+  });
+};
+
+// Delete item by given id
+const deleteItemById = function(req, resp) {
+  Item.find({ _id: req.params.id }).deleteOne((err, product) => {
+    err
+      ? resp.status(400).send(err.message)
+      : resp.status(200).send(`${product.n} item(s) deleted!`);
```
jakub-bogacz: you can use sendStatus(204) - No content.

src/server/index.js
```diff
@@ -1,10 +1,26 @@
 const express = require('express');
+const bodyParser = require('body-parser');
+const mongoose = require('mongoose');
+
+const item = require('./routes/item.route');
```
jakub-bogacz: I think that we don't need to duplicate 'route' part, so please name files just:
```
./routes/item
./routes/items
```

src/server/routes/item.route.js
```diff
@@ -0,0 +1,16 @@
+const express = require('express');
+
+const router = express.Router();
+
+const itemController = require('../controllers/item.controller');
```
jakub-bogacz: on require you can also deconstruct

src/server/routes/items.route.js
```diff
@@ -0,0 +1,16 @@
+const express = require('express');
+
+const router = express.Router();
+
+const itemsController = require('../controllers/items.controller');
```
jakub-bogacz: please deconstruct on require

.eslintrc
```diff
@@ -44,6 +46,7 @@
     "space-before-blocks": "error",
     "jsx-a11y/label-has-associated-control": 0,
     "jsx-a11y/label-has-for": 0,
+    "func-names": ["error", "never"],
```
marekrozmus: please sort alphabetically

src/server/index.js
```diff
@@ -1,10 +1,26 @@
...
+
+// Endpoint for all operations with /items
+app.use('/items', items);
 
-// Index endpoint
+// Root endpoint
 app.get('/', (req, resp) => resp.send('Hello World'));
 
-app.listen(8080, () => console.log('Listening on port 8080!'));
+app.listen(8080, () => console.info('Listening on ort 8080!'));
```
marekrozmus: 'p' is missing

src/server/controllers/items.js
```diff
@@ -0,0 +1,25 @@
+const Item = require('../models/item.model');
+
+// Get all the items
+const getAllItems = function(req, resp) {
+  switch (req.params.type) {
+    case 'ordered':
+      Item.find({ isOrdered: true }, (err, items) =>
```
marekrozmus: pull out this to some variable like: findConditions and call the Item.find(findConditions...) only once

PR #6, opened 2018-12-12
Ui messages

adamklepacz:
Ui messages for fetching items and adding new items.


src/client/App/App.js
```diff
@@ -0,0 +1,68 @@
...
+        <ProductsContainer isArchive products={archiveList} />
+        <Footer />
+        {fetchStatus === 'true' && <Preloader message="Fetching data..." />}
+      </div>
+    );
+  }
+}
+
+App.propTypes = {
+  fetchStatus: PropTypes.string,
```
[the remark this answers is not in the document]
adamklepacz: I've been talking with Kuba about this topic because we need 2 statuses: one for fetching data, second for adding new items. 
We decided that for representation three states of status which is pending, resolved and error we gonna use strings instead of booleans.

screenshot from 2018-12-18 10-46-05

But maybe there is a better way to store these statuses in the state. 

```
fetchStatus: boolean(true,false),
newItemStatus: boolean(true, false),

err: Object {
    status: boolean(true,false),
    message: String(eg. "Fetching went wrong", "Adding new item went wrong");
}
```

Something like this?? Two statuses properties and one err property as an object with some error status and error message??  Drop me a line what would be best solution.

src/client/App/actions.js
```diff
@@ -0,0 +1,18 @@
...
+// Action creators
+export const fetchItemsError = err => ({ type: FETCH_FAILED, err });
+export const recieveItems = json => ({ type: FETCH_ITEMS, items: json });
+
+// Dispatchers
+export const fetchItems = () => dispatch =>
+  fetch(`${ENDPOINT_URL}/items`, { credentials: 'same-origin' })
+    .then(resp => resp.json())
+    .then(json => {
+      setTimeout(() => dispatch(recieveItems(json)), 1500);
```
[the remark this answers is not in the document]
adamklepacz: This is only to simulate fetching/downloading experience for the user. 
Otherwise fetching is executed so fast that there is no preloader effect visible for the user.

src/client/App/App.js
```diff
@@ -0,0 +1,70 @@
...
+  items: PropTypes.arrayOf(PropTypes.object),
+
+  fetchItems: PropTypes.func
+};
+
+const mapStateToProps = state => ({
+  fetchStatus: state.uiStatus.fetchStatus,
```
marekrozmus: selectors for all items in mapStateToProps

src/client/App/App.js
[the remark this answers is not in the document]
marekrozmus: Please just create 'enum' with consts so that we don't have string comparison. You can easily make a typo in string comparison.

src/client/App/actions.js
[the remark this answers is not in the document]
marekrozmus: user don't need to see this if he/she got fast network :) for testing such stuff please use dev tools/network in chrome

src/client/components/InputBar/actions.js
```diff
@@ -0,0 +1,28 @@
...
+export const addItem = item => dispatch =>
+  fetch(`${ENDPOINT_URL}/item/create`, {
+    body: JSON.stringify(item),
+    headers: {
+      'Content-Type': 'application/json'
+    },
+    method: 'POST'
+  })
+    .then(resp => resp.json())
+    .then(json => {
```
marekrozmus: not needed braces

src/client/components/InputBar/InputBar.js
```diff
@@ -0,0 +1,95 @@
...
+    };
+
+    addItem(newItem)
+      .then(() => {
+        this.setState({
+          itemAuthor: '',
+          itemName: ''
+        });
+      })
+      .catch(err => console.error(err.message));
```
marekrozmus: shouldn't this error be shown in UI?

src/client/components/ProductsList/ProductsList.js
```diff
@@ -0,0 +1,70 @@
...
+        )}
+      </Fragment>
+    );
+  }
+}
+
+ProductsList.propTypes = {
+  toggle: PropTypes.func,
+  isArchive: PropTypes.bool,
+  products: PropTypes.arrayOf(PropTypes.object)
```
marekrozmus: sort props and separate functions

src/client/components/ProductsList/actions.js
```diff
@@ -0,0 +1,23 @@
...
+const toggle = (id, isOrdered) => dispatch => {
+  fetch(`${ENDPOINT_URL}/item/${id}/update`, {
+    body: JSON.stringify({ _id: id, isOrdered: !isOrdered }),
+    headers: {
+      'Content-Type': 'application/json'
+    },
+    method: 'PATCH'
+  })
+    .then(resp => resp.json())
+    .then(item => setTimeout(() => dispatch(toggleItem(item)), 500))
```
marekrozmus: remove setTimeout

src/client/components/InputBar/InputBar.js
adamklepacz: Error shown in UI is handled in actions by dispatching ` dispatch(addItemError(err));`

Here inside the component, I need this catch statement to not reset state if an error occurs, 
if I remove it all inputs are being reset anyway. 
Maybe I sholudn't console.log this error? - but if so, what to do inside this catch statement?

src/client/components/ProductsList/ProductsList.js
adamklepacz: should it always be like, properties first and then down below functions?

Even if this is not in alphabetical order??

src/client/components/ProductsList/actions.js
adamklepacz: I've been trying to implement this animation with external library react-transition-group but it doesn't work properly in our use case - also tested with Przemek. Items were blinking in an unpredictable order and UI gets poor because of this.

Current solution is super simple and works properly. Regarding to Dan Abramov post on Stack Overflow, using setTimeout while dispatching an action is not an anti-pattern: 
[https://stackoverflow.com/questions/35411423/how-to-dispatch-a-redux-action-with-a-timeout](url)

src/client/App/App.js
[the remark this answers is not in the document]
marekrozmus: Please create prop type for that enum

src/client/App/actions.js
```diff
@@ -0,0 +1,18 @@
...
+
+// Action creators
+export const fetchItemsError = err => ({ type: FETCH_FAILED, err });
+export const recieveItems = json => ({ type: FETCH_ITEMS, items: json });
+
+// Dispatchers
+export const fetchItems = () => dispatch =>
+  fetch(`${ENDPOINT_URL}/items`, { credentials: 'same-origin' })
+    .then(resp => resp.json())
+    .then(json => {
```
marekrozmus: unneeded braces

src/client/common/enums.js
```diff
@@ -0,0 +1,13 @@
+const messageType = Object.freeze({
```
marekrozmus: please make the first letter uppercase

src/client/components/ProductsList/ProductsList.js
```diff
@@ -0,0 +1,70 @@
...
+            <p>There are no items!</p>
+          </div>
+        ) : (
+          <ul className="products-list">
+            {products.slice(0, limit).map(item => (
+              <ProductsListItem
+                author={item.author}
+                id={item._id}
+                image={item.image}
+                archived={archived}
```
marekrozmus: sort props

src/client/reducers/itemsReducer.js
```diff
@@ -0,0 +1,62 @@
...
+  ADD_ITEM_FAILURE,
+  ADD_ITEM_SUCCESS
+} from '../components/InputBar/actions';
+import { statusType } from '../common/enums';
+
+const items = (state = itemsInitialState, action) => {
+  const { type } = action;
+  switch (type) {
+    case ADD_ITEM_SUCCESS:
+      return [...state, action.item];
```
marekrozmus: consider using 'payload' for action param

src/client/reducers/itemsReducer.js
```diff
@@ -0,0 +1,62 @@
...
+      return state;
+  }
+};
+
+/**
+ * Use string instead of booleans to indicate status of fetching
+ * items or adding a new item status. Eg. For pending state: "true",
+ * for resolved state: "false", for error: "error".
+ */
+const uiStatus = (state = status, action) => {
```
marekrozmus: please change 'status' to 'initialStatus'

PR #13, opened 2019-01-09
[EOC-17] Configure Webpack to ged rid of relative imports

alemikolo:


webpack.config.js
```diff
@@ -10,6 +10,9 @@ module.exports = {
     path: path.join(__dirname, outputDirectory),
     filename: 'bundle.js'
   },
+  resolve: {
+    modules: [path.resolve('./node_modules'), path.resolve('./src/client')]
```
adamklepacz: do we need `path.resolve('./node_modules')` if a webpack explicitly is looking into node_modules folder with no need of providing a full relative path ??

alemikolo: path.resolve('./node_modules') is necessary, without it, the webpack can't compile the app

package.json
```diff
@@ -47,6 +47,7 @@
     "eslint": "5.9.0",
     "eslint-config-airbnb": "17.1.0",
     "eslint-config-prettier": "3.3.0",
+    "eslint-import-resolver-webpack": "^0.10.1",
```
marekrozmus: please install only with -E option (exact version)

webpack.config.js
marekrozmus: please try following:
modules: [path.resolve(__dirname, 'src/client'), 'node_modules']

package.json
adamklepacz: or you can use `npm i <package-name> --save-exact` , it's equivalent to -E

.eslintrc
```diff
@@ -2,6 +2,13 @@
   "extends": ["airbnb", "plugin:prettier/recommended"],
   "plugins": ["dependencies"],
   "parser": "babel-eslint",
+  "settings": {
+    "import/resolver": {
+      "webpack": {
+        "config": "webpack.config.js"
+      }
+    }
+  },
```
adamklepacz: What this piece of code is doing? Just curious

alemikolo: The webpack configuration has been modified so that there are no long relative paths. But eslint doesn't see these settings and throws errors. This setting indicates the path to the webpack configuration where the rules for importing files are.

PR #14, opened 2019-01-09
[EOC-20] Update colors to Sandstream palette - move to other branch

adamklepacz:


src/client/components/Header.js
```diff
@@ -1,11 +1,8 @@
 import React from 'react';
 
 import AppIcon from '../assets/images/coffee-solid.svg';
-import {
-  COMPANY_NAME,
-  COMPANY_PAGE_URL,
-  PROJECT_NAME
-} from '../common/variables';
+import { COMPANY_PAGE_URL, PROJECT_NAME } from '../common/variables';
+import SandstreamLogo from '../assets/images/sandstream_logo.png';
```
marekrozmus: please change to something more general like CompanyLogo and also change file name

src/client/components/Header.js
```diff
@@ -19,7 +16,11 @@ const Header = props => (
       rel="noopener noreferrer"
       target="_blank"
     >
-      {COMPANY_NAME}
+      <img
+        alt="Sandstream logo"
```
marekrozmus: please change to "Company logo"

src/client/styles/variables/_colors.scss
```diff
@@ -1,17 +1,20 @@
 $green: #08ab6b;
 $light-green: #e1f5c4;
-$purple: #5562eb;
 $red: #fa0202;
 $red-light: #e0a5a5;
 $gray1: #646c73;
```
alemikolo: Color $gray1: #646c73 (used to text-normal) is similar to rgb(65, 64, 66)/#414042; which is defined in Sandstream colors book. It's a subtle change, but maybe it's better to use it with a stencil?

adamklepacz: I've tried and it's much more darker. So for now I leave it as it is.

PR #16, opened 2019-01-14
[EOC 13] voting feature thumb up

alemikolo:


src/client/reducers/itemsReducer.js
```diff
@@ -15,22 +22,18 @@ const items = (state = itemsInitialState, action) => {
     case FETCH_ITEMS:
       return action.items;
     case TOGGLE_ITEM: {
-      return state.map(item =>
-        item._id === action.item._id
-          ? { ...item, isOrdered: !item.isOrdered }
-          : item
-      );
+      const { _id, isOrdered, author, votes } = action.item;
+      return updateItem(state, _id, { isOrdered, author, votes });
```
marekrozmus: please revert toggle functionality

src/client/components/ProductsList/actions.js
```diff
@@ -2,22 +2,40 @@ import { ENDPOINT_URL } from '../../common/variables';
...
+    body: JSON.stringify(payload),
+    headers: { 'Content-Type': 'application/json' },
     method: 'PATCH'
   })
     .then(resp => resp.json())
-    .then(item => setTimeout(() => dispatch(toggleItem(item)), 500))
+    .catch(err => console.error(err));
+
+export const toggle = (id, isOrdered, updatedAuthor) => dispatch => {
+  const payload = {
```
marekrozmus: please revert to old toggle code

src/client/components/ProductsList/actions.js
```diff
@@ -2,22 +2,37 @@ import { ENDPOINT_URL } from '../../common/variables';
...
+export const vote = (id, votes) => dispatch => {
+  const payload = { _id: id, votes };
+  fetch(`${ENDPOINT_URL}/item/${id}/update`, {
+    body: JSON.stringify(payload),
+    headers: { 'Content-Type': 'application/json' },
+    method: 'PATCH'
+  })
+    .then(resp => resp.json())
+    .then(item => dispatch(voteForItem(item)))
+    .catch(err => console.error(err));
```
adamklepacz: wouldn't be better to  `console error(err.message)` to get more specific message when error occurs?

src/client/components/VotingBox.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+const VotingBox = ({ voteForItem, votesNumber, whetherUserVoted }) => (
+  <button
+    aria-label="Vote for item"
```
adamklepacz: do we need this? 
For now we are not supporting screen readers so I think it's not necessary

src/client/components/VotingBox.js
```diff
@@ -0,0 +1,29 @@
...
+    className={classNames('voting-box', {
+      'voting-box__voted': whetherUserVoted
+    })}
+    type="button"
+    onClick={e => {
```
adamklepacz: alphabetical order of props

src/client/components/VotingBox.js
```diff
@@ -0,0 +1,29 @@
...
+      e.stopPropagation();
```
adamklepacz: do we need to stop event propagation here?? Please check if necessary

src/client/components/VotingBox.js
```diff
@@ -0,0 +1,29 @@
...
+      voteForItem();
```
adamklepacz: not sure if we should call this here directly or just invoke as a reference?  `onClick={voteForItem}`

src/client/components/VotingBox.js
```diff
@@ -0,0 +1,29 @@
...
+    }}
+  >
+    <i className="fas fa-thumbs-up voting-box__icon" />
```
adamklepacz: consider using local SVG file, import it and place as `<VotingIcon />`
I did something like this in SortBox component - check it out.

src/client/styles/components/_VotingBox.scss
```diff
@@ -0,0 +1,30 @@
...
+  margin-right: 100px;
+  z-index: 3;
+  height: 40px;
+  padding: 0 10px;
+  background-color: $bg-color;
+  outline: none;
+  border: 2px solid lighten($gray4, 20%);
+  border-radius: 10px;
+  color: lighten($gray4, 20%);
+  font-size: 1.2em;
```
adamklepacz: consider using px instead of em's to make it consistent across all the project
Em's are hard to predict because they depend on parent fonts size. 
In my opinion, if I would use relative units I'll always use REM, cause they're much more predictable

src/client/components/VotingBox.js
alemikolo: Changed to local svg. In order not to modify the color of the image, I changed the button styles gently.

PR #17, opened 2019-01-16
[EOC-7] Filter products 

adamklepacz:
Filter products by showing all and by showing my products.
As we talked with @marekrozmus for now, only these two options.


src/client/components/ProductsContainer.js
```diff
@@ -1,27 +1,41 @@
...
+  { id: FilterOptionType.ALL_PRODUCTS, label: 'all products' },
+  { id: FilterOptionType.MY_PRODUCTS, label: 'my products' }
+];
+
 class ProductsContainer extends Component {
   state = {
     sortBy: SortOptionType.NAME,
-    sortOrder: SortOrderType.ASCENDING
+    sortOrder: SortOrderType.ASCENDING,
+
```
marekrozmus: please do not use new lines in object definition

src/client/components/ProductsContainer.js
```diff
@@ -53,17 +67,44 @@ class ProductsContainer extends Component {
     return sortOrder === SortOrderType.ASCENDING ? result : result.reverse();
   };
 
+  onFilterChange = filterBy => this.setState({ filterBy });
+
+  filterProducts = (products, filterBy) => {
+    let result = [...products];
```
marekrozmus: please remove this

src/client/components/ProductsContainer.js
```diff
@@ -53,17 +67,44 @@ class ProductsContainer extends Component {
...
+    const { currentUserId } = this.props;
+
+    switch (filterBy) {
```
marekrozmus: change to simple if and return collection without assigning to `result`

src/client/components/ProductsContainer.js
```diff
@@ -80,7 +121,15 @@ class ProductsContainer extends Component {
 
 ProductsContainer.propTypes = {
   archived: PropTypes.bool,
+  currentUserId: PropTypes.string.isRequired,
   products: PropTypes.arrayOf(PropTypes.object)
 };
 
-export default ProductsContainer;
+const mapStateToProps = state => ({
+  currentUserId: state.currentUser.id
```
marekrozmus: create selector for `currentUser`

src/client/components/ProductsContainer.js
```diff
@@ -80,7 +121,15 @@ class ProductsContainer extends Component {
...
+});
+
+export default connect(
+  mapStateToProps,
+  null
```
marekrozmus: this is not needed, you can invoke `connect` with only first parameter

src/server/controllers/item.js
```diff
@@ -3,9 +3,10 @@ const filter = require('../common/utilities');
 
 // Create new item
 const itemCreate = (req, resp) => {
-  const { author, isOrdered, name } = req.body;
+  const { author, authorId, isOrdered, name } = req.body;
```
alemikolo: new variable is authorId, maybe it would be good to change the name of the author variable to the authorName, this would be more self-documenting

PR #24, opened 2019-01-21
[EOC-32] Reorganize project stucture

adamklepacz:


client/src/app/styles/index.scss
```diff
@@ -16,21 +16,19 @@
...
-@import '../../modules/legacy/styles/components/ProductsContainer';
-@import '../../modules/legacy/styles/components/ProductsList';
-@import '../../modules/legacy/styles/components/InputBar';
-@import '../../modules/legacy/styles/components/UserBar';
-@import '../../modules/legacy/styles/components/SortBox';
-@import '../../modules/legacy/styles/components/VotingBox';
-@import '../../modules/legacy/styles/components/FilterBox';
-@import '../components/Toolbar/Toolbar';
+@import '../../common/components/AppLogo.scss';
+@import '../../app/components/Footer.scss';
```
alemikolo: Maybe it's better and cleaner to place in the each module's directory an index.scss file that would contain the import of scss files from a given module. Those module's main scss files could be imported in main index.scss from the app/styles directory.

client/src/modules/dashboard/index.jsx
```diff
@@ -1,11 +1,17 @@
...
+import ShoppingList from 'modules/shopping-list';
+
 const Dashboard = ({
   match: {
     params: { id }
   }
-}) => <div>Dashboard</div>;
+}) => (
+  <div>
+    <ShoppingList />
```
alemikolo: does full Shopping List should  be displayed in a dashboard? Shoping list module is better place to put it in..

adamklepacz: It's only a temporary solution. In the next task, you will move it to another view.

client/src/app/components/Layout/Layout.jsx
```diff
@@ -6,11 +6,11 @@ import { Redirect, Route, Switch, withRouter } from 'react-router-dom';
 import ShoppingList from 'modules/shopping-list';
 import Dashboard from 'modules/dashboard';
 import Cohort from 'modules/cohort';
-import AuthBox from 'app/components/AuthBox';
+import AuthBox from 'modules/authorization/AuthBox';
 import Toolbar from '../Toolbar/Toolbar';
-import { setCurrentUser } from 'app/components/Toolbar/actions';
+import { setCurrentUser } from 'app/components/Toolbar/model/actions';
```
alemikolo: setCurrentUser is used only in Layout, why it's model directory is in Toolbar?

client/src/app/components/Toolbar/components/UserBar/UserBar.jsx
```diff
@@ -3,8 +3,8 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import classNames from 'classnames';
 
-import { logoutCurrentUser } from './actions';
-import { getCurrentUser } from 'common/selectors';
+import { logoutCurrentUser } from './model/actions';
```
alemikolo: only wondering if actions related with user should be place in different places?

adamklepacz: actions referring to their components should be placed within this component directory - this is how we decided, so I hope it's in a good place 😃

client/src/app/styles/index.scss
adamklepacz: We decided to leave it as it is, casue we want to keep .scss file just next to .jsx file of component...
zrzut ekranu 2019-01-21 o 13 55 54

client/src/app/components/Toolbar/components/UserBar/UserBar.jsx
```diff
@@ -3,8 +3,8 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import classNames from 'classnames';
 
-import { logoutCurrentUser } from './actions';
-import { getCurrentUser } from '../selectors';
+import { logoutCurrentUser } from './model/actions';
```
marekrozmus: please move to authorization

PR #29, opened 2019-01-25
[EOC-36] save users data on signup to database

alemikolo:


server/config/auth.js
```diff
@@ -26,11 +28,11 @@ passport.deserializeUser((user, done) => {
 });
 
 const authenticate = passport.authenticate('google', {
-  scope: ['https://www.googleapis.com/auth/plus.login']
+  scope: ['https://www.googleapis.com/auth/plus.login', 'email', 'profile']
 });
 
 const authenticateCallback = passport.authenticate('google', {
-  failureRedirect: '/'
+  failureRedirect: FAILURE_URL
```
adamklepacz: why do we have to redirect to failure_url ??

server/controllers/user.js
```diff
@@ -0,0 +1,83 @@
...
+      done(null, false);
+    } else if (currentUser) {
+      done(null, currentUser);
+    } else {
+      new User({ ...user })
+        .save()
+        .then(newUser => {
+          done(null, newUser);
+        })
+        .catch(() => {
```
adamklepacz: should you catch(err) here??

server/controllers/user.js
```diff
@@ -0,0 +1,83 @@
...
+    name: name.givenName,
+    provider: 'google',
+    surname: name.familyName
+  };
+};
+
+// Get user by given id
+const getUserById = (req, resp) => {
+  User.findById(req.params.id, (err, user) => {
+    err ? resp.status(400).send(err.message) : resp.status(200).json(user);
```
adamklepacz: I would suggest using 404 code 'not found' here

server/controllers/user.js
```diff
@@ -0,0 +1,83 @@
...
+
+  User.findOneAndUpdate(
+    { _id: req.params.id },
+    newData,
+    { new: true },
+    (err, doc) => {
+      if (!doc) return resp.status(404).send('No user of given id');
+      return err
+        ? resp.status(400).send(err.message)
+        : resp.status(200).send(doc);
```
adamklepacz: should it be `resp.status(200).json(doc);` ??
what do we get when it's `send(doc)`??

server/index.js
```diff
@@ -24,8 +24,8 @@ app.use(
   session({
     secret: process.env.EXPRESS_SESSION_KEY,
     unset: 'destroy',
-    saveUninitialized: true,
-    resave: true
+    saveUninitialized: false,
+    resave: false
```
adamklepacz: just curious, what these changes are doing>>

server/config/auth.js
alemikolo: this code was removed

server/controllers/user.js
alemikolo: doc is an object

server/index.js
alemikolo: resave: true - forces the session to be saved back to the session store, even if the session was never modified...
saveUninitialized: true - forces a session that is "uninitialized" to be saved to the store.
Source: https://www.npmjs.com/package/express-session

server/controllers/userAuth.js
```diff
@@ -1,18 +1,15 @@
...
-
   // Set user cookies
   res.cookie(
     'user',
     JSON.stringify({
       name: req.user.displayName,
-      id: req.user.id,
-      avatar: req.user.photos[0].value
+      id: req.user.idFromProvider,
+      avatarUrl: req.user.avatarUrl
```
marekrozmus: alphabetical ordering

PR #32, opened 2019-01-29
[EOC-30] create new org form store to db

adamklepacz:


client/src/app/components/Toolbar/Toolbar.jsx
```diff
@@ -16,28 +16,42 @@ import UserBar from './components/UserBar';
...
 class Toolbar extends PureComponent {
   state = {
+    cohortFormVisbility: false,
     shoppingFormVisbility: false
   };
 
-  handleShoppingFormVisibilty = () => {
-    const { shoppingFormVisbility } = this.state;
-    this.setState({ shoppingFormVisbility: !shoppingFormVisbility });
+  handleFormVisibilty = isShoppingList => {
```
marekrozmus: should we check if this is list or cohort? Maybe we should show only one form - the one that was just clicked and hide all others? Clicking outside of form (if it is empty) should cliese the form popup.

client/src/modules/cohort/model/actions.js
```diff
@@ -1,9 +1,28 @@
...
+
+const fetchCohortsSuccess = data => ({
+  type: CohortActionTypes.FETCH_COHORTS_SUCCESS,
+  payload: data
 });
 
 // Dispatchers
-export const fetchCohorts = () => dispatch => dispatch(fetchCohortsSuccess());
+export const createCohort = (name, description) => dispatch =>
+  postData(`${ENDPOINT_URL}/cohorts/new-cohort`, { name, description })
```
marekrozmus: just `new` is enough - we got info about what will be created before /

client/src/modules/shopping-list/components/CreationForm.jsx
```diff
@@ -57,6 +57,7 @@ class CreationForm extends Component {
 }
 
 CreationForm.propTypes = {
+  isShoppingList: PropTypes.bool,
```
marekrozmus: remove this - the form should not know anything about shopping lists or cohorts - it should be also common component

server/models/cohort.model.js
```diff
@@ -0,0 +1,16 @@
+const mongoose = require('mongoose');
+
+const { Schema } = mongoose;
+
+const CohortSchema = new Schema(
+  {
+    adminIds: { type: String },
+    description: { type: String },
```
marekrozmus: no field is required?

server/routes/cohort.js
```diff
@@ -0,0 +1,15 @@
...
+
+const {
+  createCohort,
+  getCohortById,
+  getCohorts
+} = require('../controllers/cohort');
+
+router.get('/', getCohorts);
+router.get('/:id', getCohortById);
+router.post('/new-cohort', createCohort);
```
marekrozmus: remove `-cohort` from route

server/controllers/cohort.js
```diff
@@ -0,0 +1,41 @@
+const Cohort = require('../models/cohort.model');
+
+const createCohort = (req, resp) => {
+  const { description, name } = req.body;
+
+  const cohort = new Cohort({
+    name,
+    description
+  });
```
alemikolo: Shouldn't we save user id here (as adminId and memberId)?

adamklepacz: on creation we are saving ownerId and adminId

client/src/modules/cohort/model/actions.js
```diff
@@ -14,12 +14,11 @@ const fetchCohortsSuccess = data => ({
...
 
 // Dispatchers
-export const createCohort = (name, description, ownerId) => dispatch =>
+export const createCohort = (name, description, adminId) => dispatch =>
   postData(`${ENDPOINT_URL}/cohorts/create`, {
     name,
     description,
-    ownerId,
-    adminIds: ownerId
+    adminIds: adminId
```
alemikolo: shouldn't adminIds be an array? See in shoppingList controller also. Here string is saved... (Cohorts also...)

client/src/modules/cohort/model/actions.js
```diff
@@ -1,9 +1,32 @@
...
+
+const fetchCohortsSuccess = data => ({
+  type: CohortActionTypes.FETCH_COHORTS_SUCCESS,
+  payload: data
 });
 
 // Dispatchers
-export const fetchCohorts = () => dispatch => dispatch(fetchCohortsSuccess());
+export const createCohort = (name, description, adminId) => dispatch =>
+  postData(`${ENDPOINT_URL}/cohorts/create`, {
```
marekrozmus: frondend does not need to know db structure - it should only send data and server will take care of use that data to create correct db entry.

client/src/modules/shopping-list/model/actions.js
```diff
@@ -27,8 +27,12 @@ export const fetchProducts = () => dispatch =>
     .then(json => dispatch(recieveProducts(json)))
     .catch(err => dispatch(fetchProductsError(err)));
 
-export const createNewShoppingList = (name, description) => dispatch =>
-  postData(`${ENDPOINT_URL}/shopping-lists/new-list`, { name, description })
+export const createShoppingList = (name, description, adminId) => dispatch =>
+  postData(`${ENDPOINT_URL}/shopping-lists/create`, {
+    name,
+    description,
+    adminIds: adminId
```
marekrozmus: same here about data structure

server/controllers/cohort.js
```diff
@@ -0,0 +1,42 @@
+const Cohort = require('../models/cohort.model');
+
+const createCohort = (req, resp) => {
+  const { description, name, adminIds } = req.body;
```
marekrozmus: only one creator id should be send from front end

client/src/modules/cohort/model/actions.js
adamklepacz: it is saved as array, it's controlled by model

adamklepacz: zrzut ekranu 2019-02-04 o 12 20 13

PR #39, opened 2019-02-13
[EOC-34] edit list page

alemikolo:


client/src/common/components/ModalBox.jsx
```diff
@@ -0,0 +1,29 @@
...
+      )}
+      {children}
+    </div>
+  </div>
+);
+
+ModalBox.propTypes = {
+  children: PropTypes.node.isRequired,
+
+  onClose: PropTypes.func
```
adamklepacz: shoudn't be this required?

client/src/modules/shopping-list/model/reducer.js
```diff
@@ -16,6 +16,18 @@ const shoppingLists = (state = {}, action) => {
...
+        ...state[action.payload.listId]
+      };
+      updateList.name = action.payload.name || updateList.name;
+      updateList.description =
+        action.payload.description || updateList.description;
+      return {
+        ...state,
+        [action.payload.listId]: updateList
+      };
+    }
```
adamklepacz: maybe this would easier to reason about ??
```
      const previousName = { ...state[action.payload.listId] }.name;
      const previousDescription = { ...state[action.payload.listId] }
        .description;

      const updatedList = {
        ...state[action.payload.listId],
        name: action.payload.name || previousName,
        description: action.payload.description || previousDescription
      };

      return {
        ...state,
        [action.payload.listId]: updatedList
      };
```

server/controllers/item.js
```diff
@@ -56,10 +56,12 @@ const updateItem = (req, resp) => {
...
     (err, doc) => {
-      if (!doc) return resp.status(404).send('No item of given id');
+      if (!doc) {
+        return resp.status(404).send({ message: 'No item of given id' });
+      }
       return err
-        ? resp.status(400).send(err.message)
-        : resp.status(200).send(doc);
+        ? resp.status(400).send({ message: 'Updating item failed...' })
+        : resp.status(404).send({ message: 'Updating success' });
```
adamklepacz: Updating success with code 404??

client/src/modules/shopping-list/model/reducer.js
alemikolo: Good idea to make it more readable. I changed it similarly to your proposition but avoiding multiple spreading

client/src/common/components/DropdownMenu.jsx
```diff
@@ -77,7 +84,7 @@ DropdownMenu.propTypes = {
       icon: PropTypes.string,
       label: PropTypes.string,
 
-      callback: PropTypes.func
+      onClick: PropTypes.func
```
marekrozmus: this should be required - if it is undefined and item clicked then it will try to run undefined

client/src/modules/shopping-list/index.jsx
```diff
@@ -65,31 +99,39 @@ class ShoppingList extends Component {
...
         {showDialogBox && (
-          <DialogBox
-            onCancel={this.hideDialogBox}
-            onConfirm={this.deleteListHandler(listId, this.redirectHandler)}
-            message="Do you really want to delete the list?"
-          />
+          <ModalBox>
+            <DialogBox
+              onCancel={this.hideDialogBox}
+              onConfirm={this.deleteListHandler(listId, this.redirectHandler)}
```
adamklepacz: unneccessary second argument

PR #41, opened 2019-02-14
[EOC-55] Update authorization strategy

adamklepacz:


client/src/modules/shopping-list/index.jsx
```diff
@@ -73,7 +73,8 @@ class ShoppingList extends Component {
       },
       list
     } = this.props;
-    const listItems = list && list.products ? list.products : [];
+    const listItems =
+      list && list.products instanceof Array ? list.products : [];
```
marekrozmus: why do you need to use instanceof here?

PR #42, opened 2019-02-15
[EOC-59] redirect to login page when session lost

alemikolo:


server/middleware/authorize.js
```diff
@@ -1,9 +1,17 @@
...
       error: true,
       message: 'Unauthorized access.'
-    });
+    };
+
+    if (req.cookies['connect.sid']) {
+      res.clearCookie('connect.sid');
+      res.clearCookie('user');
+      error.message =
+        'Your session has ended, you will be redirect to the login page for 5s.';
```
adamklepacz: for 5s? shouldn't be "**in 5s**"??

adamklepacz:
zrzut ekranu 2019-02-15 o 09 44 52
I think also this issues should be fixed while session lost

PR #43, opened 2019-02-15
[EOC-57] Cleanup and closing forms toolbar 

adamklepacz:


client/src/app/components/Toolbar/Toolbar.jsx
```diff
@@ -19,27 +19,65 @@ import { createShoppingList } from 'modules/shopping-list/model/actions';
...
-    const { shoppingFormVisbility } = this.state;
+    const { shoppingFormVisibility } = this.state;
+
     this.setState({
-      shoppingFormVisbility: !shoppingFormVisbility,
-      cohortFormVisbility: false
+      cohortFormVisibility: false,
+      overlayVisibility: !shoppingFormVisibility,
+      shoppingFormVisibility: !shoppingFormVisibility,
+      userBarMenuVisibility: false
```
alemikolo: only wonder if this is scalable solution: If the button has its own handler then if we add more buttons we will have of course more handlers, and in each of them all the props  to change... If you remove button you have to edit all handlers...

client/src/app/components/Toolbar/Toolbar.jsx
```diff
@@ -58,91 +96,112 @@ class Toolbar extends PureComponent {
...
+            <div className="toolbar__right">
+              <UserBar
+                isMenuVisible={userBarMenuVisibility}
+                onClick={this.handleUserBarMenu}
+              />
+            </div>
           </div>
         </div>
-      </div>
+        {overlayVisibility && <Overlay type={OverlayStyleType.LIGHT} />}
```
alemikolo: Clever way to easily handle closing on click outside. However, I wonder if it is user friendly: when this form is visible and user decide to click something outside the toolbar (for example: list/cohort link), he won't get the effect he wants but only closing the form...

client/src/app/components/Toolbar/Toolbar.jsx
```diff
@@ -51,29 +52,32 @@ class Toolbar extends PureComponent {
 
   hideOverlayAndForm = () => {
     this.setState({
+      cohortFormVisibility: false,
       overlayVisibility: false,
-      cohortFormVisbility: false,
-      shoppingFormVisbility: false
+      shoppingFormVisibility: false,
+      userBarMenuVisibility: false
     });
```
alemikolo: there is no clean up on closing list form (cohort form also)

client/src/app/components/Toolbar/Toolbar.jsx
adamklepacz: every form handles independent part of api. any idea how to handle this form visibility and submission in another more scalable way?

client/src/app/components/Toolbar/Toolbar.jsx
adamklepacz: yes, I have noticed that also. Do you have any other solution for that? 

I have been trying with recognizing e.target on every element clicked outside the form, but there are also multiple elements like spans, inputs, and form itself which shouldn't cause form to disappear when clicked - so the IF condition excluding all of these form elements was too complicated and overloaded

alemikolo: I meant that maybe it would be good that each form have own  visibility handlers. They could work independetly. And it would be easy to remove/edit them or took to another place in code. Also you could add events on keypress (escape) and clickig outside only when current form is visible (not on didMount) to avoide invoke them when there is no need.
I wonder if we could use (To avoid using overlay):
- additional option in addeventlistener: capture  -  and set it to true
- detect if clicked element is a child o form to avoid invoke closing handler,

client/src/app/components/Toolbar/Toolbar.jsx
alemikolo: I wonder if we could use (To avoid using overlay):
- additional option in addeventlistener: capture  -  and set it to true
- detect if clicked element is a child of form to not close the form

client/src/app/components/Toolbar/Toolbar.jsx
```diff
@@ -24,21 +24,23 @@ import Overlay, { OverlayStyleType } from 'common/components/Overlay';
...
 
-  componentDidMount() {
-    document.addEventListener('click', this.clickListener);
-    document.addEventListener('keydown', this.escapeListener);
-  }
-
   componentWillUnmount() {
     document.removeEventListener('click', this.clickListener);
     document.removeEventListener('keydown', this.escapeListener);
   }
```
alemikolo: these listeners won't be removed because creationForms are always "mounted" in the toolbar (they are only not visible)

client/src/app/components/Toolbar/Toolbar.jsx
```diff
@@ -19,27 +19,73 @@ import { createShoppingList } from 'modules/shopping-list/model/actions';
...
 import { UserPropType } from 'common/constants/propTypes';
+import Overlay, { OverlayStyleType } from 'common/components/Overlay';
 
 class Toolbar extends PureComponent {
   state = {
-    cohortFormVisbility: false,
-    shoppingFormVisbility: false
+    cohortFormVisibility: false,
+    formDescription: '',
+    formTitle: '',
```
marekrozmus: these two should be in form, not here - please try react component methods to reset these values

client/src/app/components/Toolbar/Toolbar.jsx
```diff
@@ -19,27 +19,67 @@ import { createShoppingList } from 'modules/shopping-list/model/actions';
...
+  }
+
+  addEventListeners = () => {
+    document.addEventListener('click', this.clickListener);
+    document.addEventListener('keydown', this.escapeListener);
+  };
+
+  clickListener = event => {
+    const { className } = event.target;
+    className.length > 0 && className.includes('overlay')
```
marekrozmus: simple 'if' is enough - you don't need the 'else' in this case

client/src/app/components/Toolbar/Toolbar.jsx
```diff
@@ -58,91 +98,117 @@ class Toolbar extends PureComponent {
...
+            <div className="toolbar__right">
+              <UserBar
+                isMenuVisible={userBarMenuVisibility}
+                onClick={this.handleUserBarMenu}
+              />
+            </div>
           </div>
         </div>
-      </div>
+        {overlayVisibility && <Overlay type={OverlayStyleType.LIGHT} />}
```
marekrozmus: try with onClick , not event handlers

client/src/app/components/Toolbar/components/UserBar/UserBar.jsx
```diff
@@ -11,83 +11,85 @@ import UserIcon from 'assets/images/user-solid.svg';
...
-                <img
-                  alt="User Icon"
-                  className="user-bar__menu-icon"
-                  src={UserIcon}
-                />
-              </li>
-              <li className="user-bar__menu-item">
-                Profile settings
+        </button>
+
```
marekrozmus: no new lines

client/src/modules/shopping-list/index.jsx
```diff
@@ -120,22 +121,26 @@ class ShoppingList extends Component {
...
         )}
         {showUpdateFrom && (
-          <ModalBox onClose={this.hideUpdateForm}>
-            <CreationForm
-              type="modal"
-              label="Edit list"
-              onSubmit={this.updateListHandler(listId)}
-            />
-          </ModalBox>
+          <Overlay type={OverlayStyleType.MEDIUM}>
```
marekrozmus: overlay should be inside components

client/src/app/components/Toolbar/Toolbar.jsx
```diff
@@ -19,27 +19,60 @@ import { createShoppingList } from 'modules/shopping-list/model/actions';
...
+    document.removeEventListener('keydown', this.escapeListener);
+  }
+
+  addEventListeners = () => {
+    document.addEventListener('keydown', this.escapeListener);
+  };
+
+  escapeListener = event => {
+    if (event.code === 'Escape') {
+      this.hideOverlayAndForm();
```
marekrozmus: move to form and dialog components

.eslintrc
```diff
@@ -22,6 +22,7 @@
     "comma-dangle": ["error", "never"],
     "comma-spacing": ["warn"],
     "computed-property-spacing": ["error", "never"],
+    "jsx-a11y/click-events-have-key-events": "off",
```
marekrozmus: alphabetical ordering

.eslintrc
```diff
@@ -42,6 +43,7 @@
     "no-mixed-spaces-and-tabs": ["error", "smart-tabs"],
     "no-multi-spaces": ["error", { "ignoreEOLComments": false }],
     "no-multiple-empty-lines": ["error", { "max": 1 }],
+    "jsx-a11y/no-noninteractive-element-interactions": "off",
```
marekrozmus: alphabetical ordering

client/src/app/components/Toolbar/Toolbar.jsx
```diff
@@ -58,91 +66,100 @@ class Toolbar extends PureComponent {
...
+                  type="button"
+                >
+                  <CohortIcon />
+                  <img
+                    alt="Plus icon"
+                    className="toolbar__icon-plus"
+                    src={PlusIcon}
+                  />
+                </button>
+                <div className="toolbar__form">
```
marekrozmus: is this div needed if creation form is not rendered?

client/src/app/components/Toolbar/Toolbar.jsx
```diff
@@ -58,91 +66,100 @@ class Toolbar extends PureComponent {
...
+                  {shoppingFormVisibility && (
+                    <CreationForm
+                      label="Create new shopping list"
+                      onSubmit={this.handleShoppingListSubmission}
+                      type="menu"
+                      hideForms={this.hideForms}
```
marekrozmus: onHide or something similar

client/src/app/components/Toolbar/components/UserBar/UserBar.jsx
```diff
@@ -21,33 +22,45 @@ class UserBar extends Component {
   };
 
   handleShowMenu = () => {
```
marekrozmus: toggle

client/src/common/components/CreationForm.jsx
```diff
@@ -34,43 +47,55 @@ class CreationForm extends PureComponent {
...
     );
   }
 }
 
 CreationForm.propTypes = {
-  label: PropTypes.string.isRequired,
+  label: PropTypes.string,
   type: PropTypes.string,
 
+  hideForms: PropTypes.func,
```
marekrozmus: onHide or something similar

client/src/common/components/Overlay.jsx
```diff
@@ -0,0 +1,59 @@
...
+      </div>
+    );
+  }
+}
+
+Overlay.propTypes = {
+  children: PropTypes.node,
+  type: PropTypes.string.isRequired,
+
+  onVisbilityChange: PropTypes.func
```
marekrozmus: onClick

client/src/common/components/Overlay.jsx
```diff
@@ -0,0 +1,59 @@
...
+    document.removeEventListener('keydown', this.escapeListener);
+  }
+
+  onEvent = () => {
+    const { onVisbilityChange } = this.props;
+
+    onVisbilityChange(false);
+  };
+
+  escapeListener = event => {
```
marekrozmus: move keyboard handling to dialog and forms

client/src/app/components/Toolbar/components/UserBar/UserBar.jsx
```diff
@@ -9,45 +9,75 @@ import { UserPropType } from 'common/constants/propTypes';
...
 
   handleLogOut = () => {
     const { logoutCurrentUser } = this.props;
     logoutCurrentUser();
   };
 
-  handleShowMenu = () => {
-    const { hideMenu } = this.state;
-    this.setState({ hideMenu: !hideMenu });
+  toggleMenu = () => {
```
marekrozmus: this could be one line:
this.SetState(({ isVisible: oldValue ) => ({ isVisible: !oldValue }));

client/src/common/components/Overlay.jsx
```diff
@@ -0,0 +1,44 @@
...
+    const { children, type } = this.props;
+
+    return (
+      <div
+        className={classNames('overlay', {
+          'overlay--light': type === OverlayStyleType.LIGHT,
+          'overlay--dark': type === OverlayStyleType.DARK,
+          'overlay--medium': type === OverlayStyleType.MEDIUM
+        })}
+        onClick={this.onClick}
```
marekrozmus: could we call it directly from props?

client/src/app/components/Toolbar/Toolbar.jsx
```diff
@@ -58,91 +66,100 @@ class Toolbar extends PureComponent {
...
+                    src={PlusIcon}
+                  />
+                </button>
+                {cohortFormVisibility && (
+                  <div className="toolbar__form">
+                    <CreationForm
+                      label="Create new cohort"
+                      onSubmit={this.handleCohortSubmission}
+                      type="menu"
+                      onFormHide={this.hideForms}
```
marekrozmus: just onHide - we know what this will hide -> CreataionForm

PR #46, opened 2019-02-21
[EOC-62] change deletion to archivization and then delete permanently

alemikolo:


alemikolo:
archive_dialog
archive_view
delete_dialog

client/src/modules/shopping-list/index.jsx
```diff
@@ -25,17 +26,17 @@ class ShoppingList extends Component {
...
   componentDidMount() {
-    this.fetchProducts();
+    this.fetchData();
   }
 
-  fetchProducts = () => {
+  fetchData = () => {
     const {
-      fetchItemsFromGivenList,
+      fetchDataFromGivenList,
```
marekrozmus: maybe we could call it just: fetchListData(id)

client/src/modules/shopping-list/index.jsx
```diff
@@ -47,16 +48,19 @@ class ShoppingList extends Component {
...
-    this.hideDialogBox();
     const { deleteList } = this.props;
-    deleteList(id)
-      .then(this.redirectHandler)
-      .catch(noOp);
+    deleteList(id);
+  };
+
+  archiveListHandler = listId => () => {
+    this.hideDialogBox();
```
marekrozmus: this should be done after archiveList is finished

client/src/modules/shopping-list/index.jsx
```diff
@@ -87,33 +111,61 @@ class ShoppingList extends Component {
...
+            <InputBar />
+            <ProductsContainer
+              description={description}
+              name={name}
+              products={shoppingList}
+            />
+            <ProductsContainer archived products={archivedList} />
+          </div>
+        )}
+        {isArchived && (
```
marekrozmus: make separate component ArchivedList

client/src/modules/shopping-list/model/actions.js
```diff
@@ -10,19 +10,20 @@ import {
 import { ShoppingListActionTypes } from './actionTypes';
 import { MessageType as NotificationType } from 'common/constants/enums';
 import { createNotificationWithTimeout } from 'modules/notification/model/actions';
+import history from 'common/utils/history';
 
 // Action creators
```
marekrozmus: please remove this comment

client/src/modules/shopping-list/model/actions.js
```diff
@@ -79,18 +80,42 @@ const fetchShoppingListsMetaDataRequest = () => ({
...
+});
+const restoreListFailure = errMessage => ({
+  type: ShoppingListActionTypes.RESTORE_FAILURE,
+  payload: errMessage
+});
+const restoreListRequest = () => ({
+  type: ShoppingListActionTypes.RESTORE_REQUEST
+});
+
 // Dispatchers
```
marekrozmus: please remove this comment

client/src/modules/shopping-list/model/actions.js
```diff
@@ -79,18 +80,42 @@ const fetchShoppingListsMetaDataRequest = () => ({
...
-export const fetchItemsFromGivenList = listId => dispatch => {
-  dispatch(fetchProductRequest());
-  getData(`${ENDPOINT_URL}/shopping-lists/${listId}/products`)
+export const fetchDataFromGivenList = listId => dispatch => {
+  dispatch(fetchDataRequest());
+  getData(`${ENDPOINT_URL}/shopping-lists/${listId}/data`)
```
marekrozmus: consider returning the result - maybe we will need that in the place where it is called in the future

client/src/modules/shopping-list/model/actions.js
```diff
@@ -173,7 +199,52 @@ export const updateList = (listId, data) => dispatch => {
       createNotificationWithTimeout(
         dispatch,
         NotificationType.ERROR,
-        err.message || "Oops, we're sorry, updating lists failed..."
+        err.message || "Oops, we're sorry, updating list failed..."
+      );
+    });
+};
+
+export const archiveList = (listId, isArchived) => dispatch => {
```
marekrozmus: why we need here isArchived?

client/src/modules/shopping-list/model/actions.js
```diff
@@ -173,7 +199,52 @@ export const updateList = (listId, data) => dispatch => {
...
+      dispatch(archiveListFailure());
+      createNotificationWithTimeout(
+        dispatch,
+        NotificationType.ERROR,
+        err.message || "Oops, we're sorry, archiving list failed..."
+      );
+    });
+};
+
+export const restoreList = (listId, isArchived) => dispatch => {
```
marekrozmus: same here for isArchived

client/src/modules/shopping-list/model/reducer.js
```diff
@@ -28,14 +28,21 @@ const shoppingLists = (state = {}, action) => {
...
         ...state,
-        [action.payload.listId]: updatedShoppingList
+        [action.payload.listId]: archivedList
+      };
+    }
+    case ShoppingListActionTypes.RESTORE_SUCCESS:
+    case ShoppingListActionTypes.FETCH_DATA_SUCCESS: {
+      return {
+        ...state,
+        [action.payload.listId]: { ...action.payload.data }
```
marekrozmus: why do you spread the payload here?

adamklepacz:
zrzut ekranu 2019-02-22 o 09 36 10
After permanent deletion - other lists and cohorts cannot be re-fetched

client/src/modules/shopping-list/index.jsx
```diff
@@ -87,33 +111,61 @@ class ShoppingList extends Component {
       list
     } = this.props;
     const listItems = list && list.products ? list.products : [];
-    const archiveList = listItems.filter(item => item.isOrdered);
+    const archivedList = listItems.filter(item => item.isOrdered);
     const shoppingList = listItems.filter(item => !item.isOrdered);
     const description = list && list.description ? list.description : null;
+    const isArchived = list && list.isArchived ? list.isArchived : null;
```
adamklepacz: you can change it to `(list && list.isArchived) || list.isArchived

client/src/modules/shopping-list/index.jsx
```diff
@@ -87,33 +111,61 @@ class ShoppingList extends Component {
...
     const name = list && list.name ? list.name : null;
 
     return (
       <Fragment>
         <Toolbar>
-          <ToolbarItem mainIcon={<EditIcon />} onClick={this.showUpdateForm} />
-          <ToolbarItem mainIcon={<RemoveIcon />} onClick={this.showDialogBox} />
+          {!isArchived && (
+            <Fragment>
```
adamklepacz: do we really need fragment here as react only warns when he found two child elements at the top level of render method??

client/src/modules/shopping-list/index.jsx
```diff
@@ -78,6 +82,26 @@ class ShoppingList extends Component {
     this.hideUpdateForm();
   };
 
+  adjustDialogBox = (listId, isArchived) => {
```
adamklepacz: I would name it renderDialogBox as it method renders dialogBox conditionally, cause then in render method it's not clear when we see statement like this` {showDialogBox && this.adjustDialogBox(listId, isArchived)}`
This version would be more clear for someone reading render method 
` {showDialogBox && this.renderDialogBox(listId, isArchived)}`

client/src/modules/shopping-list/index.jsx
```diff
@@ -129,18 +181,17 @@ class ShoppingList extends Component {
...
   match: PropTypes.shape({
     params: PropTypes.shape({
       id: PropTypes.string
     })
   }).isRequired,
 
+  archiveList: PropTypes.func.isRequired,
   deleteList: PropTypes.func.isRequired,
-  fetchItemsFromGivenList: PropTypes.func.isRequired,
+  fetchDataFromGivenList: PropTypes.func.isRequired,
```
adamklepacz: fetchDataForCurrentList

client/src/modules/shopping-list/list.scss
```diff
@@ -0,0 +1,31 @@
+.archived-message {
+  position: relative;
+  width: 50%;
+  margin: 50px auto;
+  text-align: center;
+
+  &__header {
+    font-weight: bold;
```
adamklepacz: use digital values to be consistent across whole project

client/src/modules/shopping-list/list.scss
```diff
@@ -0,0 +1,31 @@
...
+    font-size: 20px;
+    color: $text-normal;
+  }
+
+  &__button {
+    display: block;
+    width: 150px;
```
adamklepacz: zrzut ekranu 2019-02-22 o 10 11 23
Always setup desired button width and height with padding properties.

client/src/modules/shopping-list/model/actions.js
```diff
@@ -173,7 +199,52 @@ export const updateList = (listId, data) => dispatch => {
...
+  dispatch(restoreListRequest());
+  patchData(`${ENDPOINT_URL}/shopping-lists/${listId}/update`, { isArchived })
+    .then(() => getData(`${ENDPOINT_URL}/shopping-lists/${listId}/data`))
```
adamklepacz: this seems a bit complicated for me. 
Can't you just return updated list when pathingData and than update reducer with returned data, or update reducer just using id of the list and toggle back isArchived property instead of replacing whole list object?

client/src/modules/shopping-list/model/reducer.js
```diff
@@ -28,14 +28,21 @@ const shoppingLists = (state = {}, action) => {
         [action.payload.listId]: updatedList
       };
     }
-    case ShoppingListActionTypes.FETCH_PRODUCTS_SUCCESS: {
-      const updatedShoppingList = {
-        ...state[action.payload.listId],
-        products: action.payload.products
+    case ShoppingListActionTypes.ARCHIVE_SUCCESS: {
+      const archivedList = {
+        _id: action.payload.listId,
```
adamklepacz: is this _id property needed to update??
I can see that you are updating using the key property `[action.payload.listId]` so maybe try update without this _id property

client/src/modules/shopping-list/model/reducer.js
```diff
@@ -28,14 +28,21 @@ const shoppingLists = (state = {}, action) => {
...
+        isArchived: action.payload.isArchived
       };
       return {
         ...state,
-        [action.payload.listId]: updatedShoppingList
+        [action.payload.listId]: archivedList
+      };
+    }
+    case ShoppingListActionTypes.RESTORE_SUCCESS:
```
adamklepacz: this case will be related with my previous comment about fetching back whole data instead of just toggling back isArchived property using listID. So maybe this case need to be replaced by separate case

server/controllers/shoppingList.js
```diff
@@ -8,7 +8,8 @@ const createNewList = (req, resp) => {
   const shoppingList = new ShoppingList({
     description,
     name,
-    adminIds: adminId
+    adminIds: adminId,
+    isArchived: false
```
adamklepacz: you can set this property in data model using 
isArchived: {type: String, default: false}

server/controllers/shoppingList.js
```diff
@@ -71,7 +72,7 @@ const deleteListById = (req, resp) => {
               "Oops we're sorry, an error occurred while deleting the list"
           })
         : resp.status(200).send({
-            message: `List with id: ${req.params.id} was successfully deleted!`
+            message: `List with id: ${doc.name} was successfully deleted!`
```
adamklepacz: List with id ` NAME? `

server/controllers/shoppingList.js
```diff
@@ -229,7 +236,7 @@ module.exports = {
   createNewList,
   deleteListById,
   getAllShoppingLists,
-  getProductsForGivenList,
+  getDataForGivenList,
```
adamklepacz: getDataForCurrentList

server/models/shoppingList.model.js
```diff
@@ -7,10 +7,11 @@ const ShoppingListSchema = new Schema(
   {
     adminIds: [String],
     description: { type: String },
-    products: [ItemSchema],
+    isArchived: { type: Boolean },
```
adamklepacz: add default: false here

adamklepacz:
consider also changing delete icon to archive icon from font awesome 
zrzut ekranu 2019-02-22 o 10 43 51

client/src/modules/shopping-list/index.jsx
```diff
@@ -8,13 +8,14 @@ import ProductsContainer from 'modules/shopping-list/components/ProductsContaine
 import { getShoppingList } from 'modules/shopping-list/model/selectors';
 import InputBar from 'modules/shopping-list/components/InputBar';
 import {
-  fetchItemsFromGivenList,
+  archiveList,
   deleteList,
+  fetchDataFromGivenList,
```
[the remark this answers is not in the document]
alemikolo: It's changed to fetchListData... we always fetch all data for current list only

client/src/modules/shopping-list/index.jsx
```diff
@@ -47,16 +48,19 @@ class ShoppingList extends Component {
...
+    const { archiveList } = this.props;
+    archiveList(listId, true);
   };
 
-  redirectHandler = () => {
-    const { history } = this.props;
-    history.push('/dashboard');
+  restoreListHandler = listId => () => {
+    const { restoreList } = this.props;
+    restoreList(listId, false);
```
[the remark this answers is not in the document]
alemikolo: archiveList action invokes "patchData" to change property isArchived in DB to true and remove all list's properites from redux (except _id an isArchive), restoreList action invokes "patchData" to change property isArchived in DB to false to but then it invokes getData to restore list's Data in Redux and display list content.

client/src/modules/shopping-list/index.jsx
alemikolo: Yes, we need it, without fragment we get error...

client/src/modules/shopping-list/index.jsx
alemikolo: You're right, but this method is already removed, because of new component ArchivedList which take over some functionalities

client/src/modules/shopping-list/index.jsx
alemikolo: changed to fetchListData

client/src/modules/shopping-list/model/actions.js
alemikolo: We decided to not return data when documents are updating in db, we return status and message. I didn't want change this behaviour and modify other list's action. List data is removed from redux after list archivization so we have to get it one more time.

client/src/modules/shopping-list/model/reducer.js
alemikolo: archiveList removed data from redux so we have to get it one more time

server/controllers/shoppingList.js
alemikolo: getListData

client/src/modules/shopping-list/model/reducer.js
alemikolo: we override full list data object with new data with id and isArchive

client/src/modules/shopping-list/model/reducer.js
```diff
@@ -28,14 +28,21 @@ const shoppingLists = (state = {}, action) => {
         [action.payload.listId]: updatedList
       };
     }
-    case ShoppingListActionTypes.FETCH_PRODUCTS_SUCCESS: {
-      const updatedShoppingList = {
-        ...state[action.payload.listId],
-        products: action.payload.products
+    case ShoppingListActionTypes.ARCHIVE_SUCCESS: {
+      const archivedList = {
```
marekrozmus: please use destructuring:
```
const { listId: _id, isArchived } = this.action.payload;
const archivedList = { _id, isArchived };
```

server/controllers/shoppingList.js
```diff
@@ -139,18 +140,22 @@ const getProductsForGivenList = (req, resp) => {
...
 
-      const { products } = documents[0];
-      resp.status(200).json(products);
+      const { _id, isArchived } = documents[0];
+
+      if (isArchived) {
+        return resp.status(200).json({ _id, isArchived });
+      }
+
+      resp.status(200).json(documents[0]);
```
adamklepacz: here, once you return object and once documents[0]. Can it be consistent and always return an object like in line 155?

alemikolo: documents[0] is an object

client/src/modules/shopping-list/components/ArchivedList.jsx
```diff
@@ -0,0 +1,76 @@
...
+    this.setState({ showDialogBox: false });
+  };
+
+  restoreListHandler = listId => () => {
+    const { restoreList } = this.props;
+    restoreList(listId);
+  };
+
+  deleteListHandler = id => () => {
+    const { deleteList } = this.props;
```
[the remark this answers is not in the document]
alemikolo: Because clicking on 'permanently delete' button shouldn't invoke delete action directly but display dialog to confirm this delete action... So you have to return function which holds a reference to list Id, and this returned function will be invoke when you click confirm button on the dialog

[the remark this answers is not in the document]
adamklepacz: zrzut ekranu 2019-02-25 o 10 34 15
zrzut ekranu 2019-02-25 o 10 34 24

Try this solution, it's more intuitive and common pattern for other developers to understand

PR #47, opened 2019-02-22
[EOC-54]  create cohort basic view

adamklepacz:
zrzut ekranu 2019-02-22 o 08 46 43
zrzut ekranu 2019-02-22 o 08 46 53

Basically, the flow goes like this.
On dashboard - fetching meta-data for cohorts. 
When we open given cohort - meta-data for shopping list is fetched, 
When we open shopping-list - products to shopping list are fetched.


client/src/modules/cohort/index.jsx
```diff
@@ -1,34 +1,89 @@
...
       }
     } = this.props;
 
+    fetchListMetaDataForCurrentCohort(id);
+  }
+
+  render() {
+    const { cohortDetails, lists } = this.props;
+    const description = cohortDetails && cohortDetails.description;
+    const name = cohortDetails && cohortDetails.name;
```
marekrozmus: instead of checking every field try following:
```
const { cohortDetails, lists } = this.props;

if (!cohortDetails) {
  return null;
}

const { name, description } = cohortDetails;
```

client/src/modules/cohort/index.jsx
```diff
@@ -1,34 +1,89 @@
...
+            )}
+          </div>
         </div>
       </Fragment>
     );
   }
 }
 
 Cohort.propTypes = {
+  cohortDetails: PropTypes.objectOf(PropTypes.string),
```
marekrozmus: please create more detailed prop type

client/src/modules/cohort/index.jsx
```diff
@@ -1,34 +1,89 @@
...
+  lists: PropTypes.objectOf(PropTypes.object),
   match: PropTypes.shape({
```
marekrozmus: yo are using it in many places - maybe just create prop type and store it commons?

client/src/modules/cohort/model/actions.js
```diff
@@ -53,17 +55,20 @@ export const createCohort = (name, description, adminId) => dispatch => {
     });
 };
 
-export const fetchCohorts = () => dispatch => {
-  dispatch(fetchCohortsRequest());
-  getData(`${ENDPOINT_URL}/cohorts`)
+export const fetchCohortsMetaData = () => dispatch => {
+  dispatch(fetchCohortsMetaDataRequest());
+  getData(`${ENDPOINT_URL}/cohorts/meta-data`)
```
marekrozmus: please always return promise

client/src/modules/cohort/model/selectors.js
```diff
@@ -1,3 +1,19 @@
...
 export const getCohorts = state => state.cohorts.data;
 export const getCohortsError = state => state.cohorts.errorMessage;
 export const getIsFetchingCohorts = state => state.cohorts.isFetching;
+export const getCohortDetails = (state, cohortId) => {
+  if (!_isEmpty(state.cohorts.data)) {
+    const { description, name } = _filter(
+      state.cohorts.data,
+      (value, key) => key === cohortId
+    )[0];
+
```
marekrozmus: could be:
```  
const cohort = _head(_filter(getCohorts(state), (_, key) => key === cohortId);
if (cohort) {
  const { description, name } = cohort;
  return { description, name };
} 
```

server/models/shoppingList.model.js
```diff
@@ -6,6 +6,7 @@ const ItemSchema = require('./item.model').schema;
 const ShoppingListSchema = new Schema(
   {
     adminIds: [String],
+    cohortId: { type: String, default: '' },
```
marekrozmus: Shouldn't all ids be of type Schema.Types.ObjectId?

client/src/modules/shopping-list/model/selectors.js
```diff
@@ -1,6 +1,14 @@
 import _pick from 'lodash/pick';
+import _filter from 'lodash/filter';
+import _keyBy from 'lodash/keyBy';
 
 export const getShoppingList = (state, listId) =>
   _pick(state.shoppingLists.data, listId)[listId];
 export const getShoppingLists = state => state.shoppingLists.data;
+export const getLists = (state, cohortId) =>
```
alemikolo: maybe this above method should be named getLists  (without 'shopping') and this one could have name like getListsForCohort? I know this is longer but we should remove 'shopping' word from code and there will have a naming conflict

server/controllers/shoppingList.js
```diff
@@ -78,20 +78,47 @@ const deleteListById = (req, resp) => {
...
 
 const getShoppingListsMetaData = (req, resp) => {
+  const { cohortId } = req.params;
+
+  if (cohortId) {
+    return ShoppingList.find(
+      {
+        cohortId
+      },
+      '_id name description cohortId',
```
alemikolo: what about authorization here?

client/src/modules/shopping-list/model/selectors.js
adamklepacz: it was getListForCurrentCohort previously 😄

server/controllers/shoppingList.js
```diff
@@ -149,8 +181,10 @@ const getProductsForGivenList = (req, resp) => {
         return resp.status(404).send({ message: err.message });
       }
 
-      const { products } = documents[0];
-      resp.status(200).json(products);
+      if (documents[0]) {
```
marekrozmus: do you need check it aslo here? Didn't first `if` checked if something was found?

server/models/shoppingList.model.js
```diff
@@ -1,11 +1,16 @@
...
 
 const { Schema } = mongoose;
+const { ObjectId } = Schema.Types;
 const ItemSchema = require('./item.model').schema;
 
 const ShoppingListSchema = new Schema(
   {
     adminIds: [String],
+    cohortId: {
+      type: ObjectId,
```
marekrozmus: what about other `ids` that are described as `String`?

PR #48, opened 2019-02-22
[EOC-63] Fill edit form with data

adamklepacz:


client/src/common/components/CreationForm.jsx
```diff
@@ -5,10 +5,15 @@ import classNames from 'classnames';
...
-    description: '',
-    title: ''
-  };
+  constructor(props) {
+    super(props);
+    const { initialTitle, initialDescription } = this.props;
+
+    this.state = {
+      description: initialDescription || '',
+      title: initialTitle || ''
```
marekrozmus: should this be `title` or rather `name`?

client/src/modules/shopping-list/index.jsx
```diff
@@ -120,6 +124,8 @@ class ShoppingList extends Component {
               label="Edit list"
               onSubmit={this.updateListHandler(listId)}
               type="modal"
+              defaultName={name}
+              defaultDescription={description}
```
marekrozmus: sort props

PR #50, opened 2019-02-27
[EOC-65] change toolbar content depending on user roles for lists

alemikolo:
screenshot 2019-02-27 11 16 01
screenshot 2019-02-27 11 15 31

### DIFF
--- a/client/src/modules/shopping-list/index.jsx
+++ b/client/src/modules/shopping-list/index.jsx
@@ -18,7 +18,8 @@ import CreationForm from 'common/components/CreationForm';
 import { EditIcon, ArchiveIcon } from 'assets/images/icons';
 import { noOp } from 'common/utils/noOp';
 import ArchivedList from 'modules/shopping-list/components/ArchivedList';
-import { RouterMatchPropType } from 'common/constants/propTypes';
+import { RouterMatchPropType, UserPropType } from 'common/constants/propTypes';
+import { getCurrentUser } from 'modules/authorization/model/selectors';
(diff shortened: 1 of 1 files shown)


client/src/modules/shopping-list/index.jsx
```diff
@@ -76,10 +77,19 @@ class ShoppingList extends Component {
...
-  }
+  };
+
+  checkIfAuthorized = () => {
+    const {
+      currentUser: { id: userId },
+      list
+    } = this.props;
+
+    return list && list.adminIds && list.adminIds.includes(userId);
```
adamklepacz: A method is doing two things - it is checking if list exists and checking adminIds properties.
Consider separating checkIfListExist to a separated method.

alemikolo: I think it will be too complicated... In this additional method you will have to return true or false (depending on whether the list exists or not) You will have to extract list from props there to do this. In checkIfAuthorize method you'll have to extract list once again because it will be needed to access to adminIds... 
What about this solution:
I can pass list to this method to avoid checking if list exist the:

  checkIfAuthorized = list => {
    const {
      currentUser: { id: userId }
    } = this.props;

    return list.adminIds && list.adminIds.includes(userId);
  };
render() {
...
this.checkIfAuthorized(list)
...
}

adamklepacz: so stay at the solution you have now.
Passign list as argument doesnt make sense

PR #52, opened 2019-02-27
[EOC-72] Prepare initial landing page

adamklepacz:


adamklepacz:
zrzut ekranu 2019-02-27 o 14 13 44
zrzut ekranu 2019-02-27 o 14 13 52

landing/index.html
```diff
@@ -1,1043 +1,1146 @@
...
+          <div class="col-lg-12">
+            <h1 class="wow fadeInDown">
+              End Of Coffee<br />
+              Office inventory tracking app
+            </h1>
+            <p class="wow fadeInDown">
+              End of coffee helps you track all the office inventory you need!
+            </p>
+            <div class="bnts wow fadeInUp">
+              <a href="#" class="btn2">Try Now</a>
```
alemikolo: only wonder: will it be good if we change href attribute to our localhost app (maybe via some variables...) We don't have a server but locally it will be great to be able to switch between landing page and app

landing/index.html
```diff
@@ -1,1043 +1,1146 @@
...
+            </div>
+          </div>
+
+          <div class="col-lg-4 col-md-4 col-sm-6 module">
+            <div class="f-box wow fadeInUp" data-wow-duration="3s">
+              <div class="f-img"><img src="images/icon5.png" alt="f5" /></div>
+              <a href="#">Responsive Design</a>
+              <p>
+                Lorem Ipsum is simply dummy text the printing and typeseing
+                industry Lorem Ipsum has been standard.
```
alemikolo: maybe something like this: use on all your devices, whenever you want and wherever you are...

landing/index.html
```diff
@@ -1,1043 +1,1146 @@
...
+                    <p></p>
                   </div>
-               </div>
-               
-               
+                  <div class="clear"></div>
+                </div>
+              </div>
+
+              <a href="#" class="acc wow fadeInUp">Sing up to app</a>
```
alemikolo: same as above: link to app?

landing/index.html
```diff
@@ -1,1043 +1,1146 @@
...
+              <li>Dashboard</li>
+              <li>List</li>
+              <li>Cohort</li>
+            </ul>
+            <div class="resp-tabs-container hor_1">
+              <div>
+                <p><!-- vertical Tabs --></p>
+                <img
+                  class="tab-screen-image"
+                  src="images/screen-tab-1.png"
```
alemikolo: what do you think about inserting current views of eoc?

landing/index.html
adamklepacz: I didn't want to hardcoded this - because then we will have to change it everywhere. 
This is just a static file, not going through anything so I can't use .env variables or something like this

landing/index.html
adamklepacz: App is not responsive so I decided to leave it as is to show Lukasz and maybe he will decide to remove this section

landing/index.html
adamklepacz: First, we need change the layout of app to match landing layout - than we can insert screens

PR #54, opened 2019-02-28
[EOC-66] Modify notification position and styling

adamklepacz:
zrzut ekranu 2019-02-28 o 08 47 42
zrzut ekranu 2019-02-28 o 08 38 45
zrzut ekranu 2019-02-28 o 08 48 53
zrzut ekranu 2019-02-28 o 08 56 38
zrzut ekranu 2019-02-28 o 08 52 25


client/src/modules/notification/Notification.scss
```diff
@@ -0,0 +1,14 @@
...
+  position: absolute;
+  left: 0;
+  bottom: 0;
+  padding: 50px 0;
+  display: flex;
+  justify-content: flex-start;
+  align-items: flex-end;
+
+  &__wrapper {
+    max-width: 1440px;
```
alemikolo: Where does this value come from? Maybe set it to 100%?

client/src/modules/notification/Notification.scss
```diff
@@ -0,0 +1,14 @@
+.notification {
+  position: absolute;
```
alemikolo: why absolute and not fixed? when app height is greater than screensize, notification is not visible - you have to scroll the screen

client/public/index.html
```diff
@@ -10,8 +10,8 @@
   <body>
     <div id="root"></div>
     <link
-      href="https://fonts.googleapis.com/css?family=Roboto:400,500,700&amp;subset=latin-ext"
+      href="https://fonts.googleapis.com/css?family=Montserrat:400,500,700&amp;subset=latin-ext"
```
alemikolo: Montserrat font is from our landing page but unfortunately Roboto was more readable in my opinion, maybe we should increase the font size

client/src/common/components/MessageBox.jsx
```diff
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 
 import { MessageType } from 'common/constants/enums';
```
alemikolo: Only wonder: will we use the message outside of notifications? Maybe we could unify naming (change message to notification or vice versa)

client/src/modules/notification/Notification.scss
adamklepacz: same width as app wrapper that is declared to 1440px

client/public/index.html
adamklepacz: I think we will implement such changes when we gonna customize app layout, cause now it doesn't make sense. The whole app  gone look  different later on

client/src/common/components/MessageBox.jsx
adamklepacz: Yes, we  are using them for example when no lists or no cohort are found.

alemikolo:
screenshot from 2019-02-28 15-15-43
Notification should be displayed above all elements...

PR #57, opened 2019-03-01
[EOC-38] Edit cohort

adamklepacz:
zrzut ekranu 2019-03-01 o 08 52 01
Dialog is made as it was in previous branches - old way. 
I'm currently working on the new "proper" way 🚀


client/src/modules/cohort/index.jsx
```diff
@@ -27,18 +33,53 @@ class Cohort extends PureComponent {
...
     const { name, description } = cohortDetails;
-
+    const { updateFormVisibility } = this.state;
     return (
       <Fragment>
-        <Toolbar />
+        <Toolbar>
+          <ToolbarItem mainIcon={<EditIcon />} onClick={this.showUpdateForm} />
+        </Toolbar>
+        {updateFormVisibility && (
```
alemikolo: proposal: pass updateFormVisibility as a e.g. isVisible prop to ModalForm and decide there  if creationForm should be render or not

client/src/modules/cohort/model/actions.js
```diff
@@ -70,3 +84,25 @@ export const fetchCohortsMetaData = () => dispatch => {
...
+        NotificationType.SUCCESS,
+        json.message
+      );
+    })
+    .catch(err => {
+      dispatch(updateCohortFailure(err));
+      createNotificationWithTimeout(
+        dispatch,
+        NotificationType.ERROR,
+        err.message || "Oops, we're sorry, updating list failed..."
```
alemikolo: message: updating cohort failed not list

client/src/modules/shopping-list/index.jsx
```diff
@@ -144,15 +143,13 @@ class ShoppingList extends Component {
...
-          <ModalBox onCancel={this.hideUpdateForm}>
-            <CreationForm
-              defaultDescription={description}
-              defaultName={name}
-              label="Edit list"
-              onSubmit={this.updateListHandler(listId)}
-              type="modal"
-            />
-          </ModalBox>
+          <ModalForm
```
alemikolo: consider my previous proposal: pass updateFormVisibility as a e.g. isVisible prop to ModalForm and decide there if creationForm should be render or not

server/controllers/cohort.js
```diff
@@ -50,8 +51,43 @@ const getCohortById = (req, resp) => {
...
+    },
+    dataToUpdate,
+    { new: true },
+    (err, doc) => {
+      if (!doc) {
+        return resp.status(404).send({
+          message: 'You have no permissions to perform this action'
+        });
+      }
+      return err
```
alemikolo: shouldn't we check err in separate if statement and use q quick return there?

client/src/modules/cohort/index.jsx
adamklepacz: This is rebuilt in the coming implementation so I think it doesn't make any sense to change it now.

server/controllers/cohort.js
```diff
@@ -73,14 +73,17 @@ const updateCohortById = (req, resp) => {
...
-          });
+
+      if (err) {
+        return resp.status(400).send({
+          message:
+            "Oops we're sorry, an error occurred while processing the cohort"
+        });
+      }
+
+      return resp.status(200).send({
```
alemikolo: Is this last return necessary? maybe just resp.status... is enough?

client/src/common/components/CreationForm.jsx
```diff
@@ -97,7 +97,7 @@ class CreationForm extends PureComponent {
               type="submit"
               value={defaultName ? 'Update' : 'Create'}
             />
-            {type === CreationFormType.MODAL && (
+            {onCancel && (
```
alemikolo: do we need nested condition here? What do you think about: {type === CreationFormType.MODAL && onCancel && (
)

client/src/modules/cohort/index.jsx
```diff
@@ -27,18 +33,62 @@ class Cohort extends PureComponent {
...
+
+  updateCohortHandler = cohortId => (name, description) => {
+    const { cohortDetails, updateCohort } = this.props;
+    const {
+      description: previousDescription,
+      name: previousName
+    } = cohortDetails;
+    const dataToUpdate = {};
+    name ? (dataToUpdate.name = name) : null;
+    description ? (dataToUpdate.description = description) : null;
```
marekrozmus: please use ternary operator only if return value is used
move these checks and `const dataToUpdate` inside the if in line 54 - we do not need to do those if they are the same

server/controllers/cohort.js
```diff
@@ -50,8 +51,46 @@ const getCohortById = (req, resp) => {
...
+    dataToUpdate,
+    { new: true },
+    (err, doc) => {
+      if (!doc) {
+        return resp.status(404).send({
+          message: 'You have no permissions to perform this action'
+        });
+      }
+
+      if (err) {
```
marekrozmus: move error checking to the top - in current case if we have error then we will send 404 from the first if

client/src/modules/cohort/index.jsx
```diff
@@ -27,18 +33,63 @@ class Cohort extends PureComponent {
...
+      name: previousName
+    } = cohortDetails;
+    const dataToUpdate = {};
+
+    if (
+      previousDescription !== dataToUpdate.description ||
+      previousName !== dataToUpdate.name
+    ) {
+      name && (dataToUpdate.name = name);
+      description && (dataToUpdate.description = description);
```
marekrozmus: please crete object and then filter out undefined keys:
```
const dataToUpdate = { description, name };
updateCohort(cohortId, _.pick(dataToUpdate, _.identity));
```

PR #63, opened 2019-03-07
[EOC-53] change toolbar content depending on user roles for cohort

alemikolo:
This branch started form EOC-3 so to review is last commit: 
Change toolbar content depending on user role.
screenshot 2019-03-07 12 21 55
screenshot 2019-03-07 12 22 37


client/src/modules/cohort/index.jsx
```diff
@@ -124,6 +124,15 @@ class Cohort extends PureComponent {
     return cohortDetails && cohortDetails.isArchived;
   };
 
+  checkIfAuthorized = () => {
+    const {
+      currentUser: { id: userId },
+      cohortDetails: cohort
+    } = this.props;
+
+    return cohort && cohort.adminIds && cohort.adminIds.includes(userId);
```
marekrozmus: data leak! adminIds should not be sent to client - cohots and lists right should be checked on backend and some flags `true/false` should be set on data which is send to client

client/src/modules/cohort/index.jsx
```diff
@@ -120,6 +120,11 @@ class Cohort extends PureComponent {
     return cohortDetails && cohortDetails.isArchived;
   };
 
+  checkIfAuthorized = () => {
+    const { cohortDetails: cohort } = this.props;
```
adamklepacz: maping to cohort is not necessary

client/src/modules/cohort/index.jsx
```diff
@@ -120,6 +120,11 @@ class Cohort extends PureComponent {
     return cohortDetails && cohortDetails.isArchived;
   };
 
+  checkIfAuthorized = () => {
```
adamklepacz: not sure is the name of this method is proper, 
cause you're checking if the user is admin and not if authorized.

client/src/modules/shopping-list/index.jsx
```diff
@@ -83,12 +82,8 @@ class ShoppingList extends Component {
   };
 
   checkIfAuthorized = () => {
-    const {
-      currentUser: { id: userId },
-      list
-    } = this.props;
-
-    return list && list.adminIds && list.adminIds.includes(userId);
+    const { list } = this.props;
```
adamklepacz: same about this method.
You are checking if user is admin, not if authorized.

server/common/utilities.js
```diff
@@ -6,4 +8,9 @@ const filter = f => object =>
     Object.entries(object).filter(([key, value]) => f(value, key, object))
   );
 
-module.exports = filter;
+const checkRole = (ids, userIdFromReq) => {
```
adamklepacz: second parameter - just `userId`

alemikolo: const userId is used below, function parameter can't be change to userId

server/controllers/cohort.js
```diff
@@ -126,13 +126,18 @@ const getCohortDetails = (req, resp) => {
...
 
       if (isArchived) {
         return resp.status(200).json({ _id, isArchived, name });
       }
 
-      resp.status(200).json(doc);
+      const isAdmin = checkRole(adminIds, req.user._id);
+      const data = isAdmin
+        ? { _id, isAdmin, isArchived, description, name }
+        : { _id, isArchived, description, name };
```
adamklepacz: this seems bit confusing, you can always return the same object of data.
Once with `isAdmin` set as `true`, once as `false` and there will be no need of this ternary operator here. 

You need to propably change `checkRole` utlility method

server/common/utilities.js
adamklepacz: I think also, that the first argument is a bit confusing, it should be named as `idsArray` to make this clear that we should pass array of ids here

PR #64, opened 2019-03-07
[EOC-67] New items should appear on top dashboard

adamklepacz:
Referring to what Marek said about sorting data in selectors I have created a new selector for items in the current list which is sorting items by the createdAt property to store newest one on the top. 

Anyway, lists and cohort are stored as collections of data, so sorting on collections is not possible.
I've tried to convert the collection to an array, sort lists and then using _keyBy lodash function make them collection once again, but _keyBy is not respecting the order of array and items are once again mixed. 
I have even asked on StackOverflow to sort collection - and they answer that sorting collection is not possible - only tricky way is to sort the array and then convert to the collection (but there is no guarantee about the order).

The other approach that works perfectly fine is adding a new list or cohort on the beginning of the previous state inside reducer. This solution works good for cohorts/ lists and there is no need of tricky sorting inside selectors.


client/src/modules/shopping-list/index.jsx
```diff
@@ -5,7 +5,10 @@ import { withRouter } from 'react-router-dom';
 
 import Toolbar, { ToolbarItem, ToolbarLink } from 'common/components/Toolbar';
 import ProductsContainer from 'modules/shopping-list/components/ProductsContainer';
-import { getList } from 'modules/shopping-list/model/selectors';
+import {
+  getList,
+  getItemsForCurrentList
```
alemikolo: I think that name getItems will be enough for this selector (or getListItems)...

client/src/modules/shopping-list/index.jsx
```diff
@@ -103,8 +107,8 @@ class ShoppingList extends Component {
     if (!list) {
       return null;
     }
-    const { cohortId, description, isArchived, name, products } = list;
-    const listItems = products || [];
+    const { cohortId, description, isArchived, name } = list;
+    const listItems = items || [];
     const archivedList = listItems.filter(item => item.isOrdered);
```
alemikolo: maybe this variables should contain word 'items' in name not 'list'

client/src/modules/shopping-list/index.jsx
```diff
@@ -176,7 +181,8 @@ ShoppingList.propTypes = {
 
 const mapStateToProps = (state, ownProps) => ({
   currentUser: getCurrentUser(state),
-  list: getList(state, ownProps.match.params.id)
+  list: getList(state, ownProps.match.params.id),
+  items: getItemsForCurrentList(state, ownProps.match.params.id)
```
alemikolo: See above: getItems....

client/src/modules/shopping-list/model/selectors.js
```diff
@@ -4,6 +4,18 @@ import _keyBy from 'lodash/keyBy';
 
 export const getList = (state, listId) =>
   _pick(state.lists.data, listId)[listId];
+export const getItemsForCurrentList = (state, listId) => {
```
alemikolo: getItems (or getSortedItems)...

client/src/modules/shopping-list/index.jsx
marekrozmus: I agree the current name is totally confusing - is this archived list or rather orderedItems?

client/src/modules/shopping-list/model/reducer.js
```diff
@@ -5,11 +5,12 @@ import { ItemActionTypes } from 'modules/shopping-list/components/InputBar/model
 
 const items = (state, action) => {
   switch (action.type) {
-    case ItemActionTypes.ADD_SUCCESS:
+    case ItemActionTypes.ADD_SUCCESS: {
```
marekrozmus: you don't need braces here

client/src/modules/shopping-list/model/selectors.js
```diff
@@ -4,6 +4,18 @@ import _keyBy from 'lodash/keyBy';
...
+  const currentList = _pick(state.lists.data, listId)[listId];
```
marekrozmus: This `state.lists.data` should be done with selector!

client/src/modules/shopping-list/model/selectors.js
```diff
@@ -4,6 +4,18 @@ import _keyBy from 'lodash/keyBy';
...
+  if (
+    currentList &&
+    Object.prototype.hasOwnProperty.call(currentList, 'products')
```
marekrozmus: why so complicated?

client/src/modules/shopping-list/index.jsx
```diff
@@ -108,9 +105,8 @@ class ShoppingList extends Component {
       return null;
     }
     const { cohortId, description, isArchived, name } = list;
-    const listItems = items || [];
-    const archivedList = listItems.filter(item => item.isOrdered);
-    const shoppingList = listItems.filter(item => !item.isOrdered);
+    const archivedItems = items ? items.filter(item => item.isOrdered) : [];
```
alemikolo: we have cohort and list archivization. And item ordering. Are these the same processes? If they aren't, we should use orderedItems not archiveItems. Or we should discuss this problem...

client/src/modules/shopping-list/model/selectors.js
```diff
@@ -4,15 +4,18 @@ import _keyBy from 'lodash/keyBy';
...
 export const getList = (state, listId) =>
   _pick(state.lists.data, listId)[listId];
-export const getItemsForCurrentList = (state, listId) => {
-  const currentList = _pick(state.lists.data, listId)[listId];
-  if (
-    currentList &&
-    Object.prototype.hasOwnProperty.call(currentList, 'products')
-  ) {
+export const getItems = (state, listId) => {
+  const currentList = getList(state, listId);
```
alemikolo: wonder if we need to use currentList? maybe 'list' is enough?

adamklepacz: in my opinion currentList is a good variable in this circumstance

PR #65, opened 2019-03-11
[EOC-61] Fix server issues

adamklepacz:
I did a bit of investigation before doing this task. I have found a nice article about error messages 
https://uxdworld.com/2018/05/30/how-to-write-good-error-messages/
Referring to the article all `err.message `sent from backend are replaced with custom messages, 
to make notifications understandable to the user. 

Previously using `err.message`,  we had something like this: 
Zrzut ekranu 2019-03-08 o 11 55 49
which was not understandable to a non-technical user. 
Now all the messages are easy to understand.

### DIFF
--- a/server/controllers/shoppingList.js
+++ b/server/controllers/shoppingList.js
@@ -17,7 +17,7 @@ const createList = (req, resp) => {
   list.save((err, doc) => {
     if (err) {
       return resp
-        .status(404)
+        .status(400)
         .send({ message: 'List not saved. Please try again.' });
     }
 
@@ -98,7 +98,7 @@ const getArchivedListsMetaData = (req, resp) => {
       if (err) {
         return resp.status(400).send({
           message:
-            'An error occurred while fetching the lists data. Please try again.'
+            'An error occurred while fetching the archived lists data. Please try again.'
         });
       }
 
@@ -160,7 +160,7 @@ const getListData = (req, resp) => {
     },
     (err, doc) => {
       if (err) {
-        return resp.status(404).send({
+        return resp.status(400).send({
           message:
             'An error occurred while fetching the lists data. Please try again.'
         });
@@ -175,7 +175,7 @@ const getListData = (req, resp) => {
           : resp.status(200).json(list);
       }
 
-      return resp.status(404).send({ message: 'List data not found.' });
+      resp.status(404).send({ message: 'List data not found.' });
     }
   );
 };
@@ -211,7 +211,7 @@ const updateShoppingListItem = (req, resp) => {
       if (err) {
         return resp.status(400).send({
           message:
-            'An error occurred while updating the lists data. Please try again.'
+            'An error occurred while updating the list data. Please try again.'
         });
       }
       const itemIndex = doc.products.findIndex(item => item._id.equals(itemId));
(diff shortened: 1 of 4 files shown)


server/controllers/shoppingList.js
```diff
@@ -236,10 +189,9 @@ const updateShoppingListItem = (req, resp) => {
    * are passed in the request
    *  */
   const propertiesToUpdate = {};
-  typeof isOrdered !== 'undefined'
-    ? (propertiesToUpdate['products.$.isOrdered'] = isOrdered)
-    : null;
-  voterIds ? (propertiesToUpdate['products.$.voterIds'] = voterIds) : null;
+  typeof isOrdered !== 'undefined' &&
+    (propertiesToUpdate['products.$.isOrdered'] = isOrdered);
+  voterIds && (propertiesToUpdate['products.$.voterIds'] = voterIds);
```
marekrozmus: Please do not use `&&` in this way - use simple `ifs`.
That kind of `&&` operator usage is ok for jsx.

server/controllers/shoppingList.js
```diff
@@ -205,24 +158,24 @@ const getListData = (req, resp) => {
...
       }
 
-      if (documents && documents.length > 0) {
-        const { cohortId, _id, isArchived } = documents[0];
-
-        if (isArchived) {
-          return resp.status(200).json({ cohortId, _id, isArchived });
-        }
+      if (doc) {
+        const list = _head(doc);
```
alemikolo: We look exactly for one list, so we can use findOne and use doc without lodash _head an or checking docs.length

server/controllers/shoppingList.js
```diff
@@ -205,24 +158,24 @@ const getListData = (req, resp) => {
...
-        return resp.status(200).json(documents[0]);
+        return isArchived
+          ? resp.status(200).json({ cohortId, _id, isArchived })
+          : resp.status(200).json(list);
       }
 
-      return resp
-        .status(404)
-        .send({ message: 'Products not found for given list id!' });
+      resp.status(404).send({ message: 'List data not found.' });
```
alemikolo: I think that we should check if doc exist first and use quick return if not. There will be more logic when we add checking roles, so all code for creating success response should be at the end...

server/controllers/shoppingList.js
```diff
@@ -28,111 +32,51 @@ const createList = (req, resp) => {
...
-            message:
-              "Oops we're sorry, an error occurred while deleting the list"
+      doc
+        ? resp.status(200).send({
+            message: `List "${doc.name}" successfully deleted.`
           })
-        : resp.status(200).send({
-            message: `List ${doc.name} was successfully deleted!`
-          });
+        : resp.status(404).send({ message: `List "${doc.name}" not found` });
```
alemikolo: If !doc we can't access doc.name

server/controllers/shoppingList.js
```diff
@@ -205,24 +158,22 @@ const getListData = (req, resp) => {
         { purchaserIds: req.user._id }
       ]
     },
-    (err, documents) => {
+    (err, doc) => {
       if (err) {
-        return resp.status(404).send({ message: err.message });
+        return resp.status(400).send({
+          message:
+            'An error occurred while fetching the lists data. Please try again.'
```
alemikolo: list data not lists - this method gets details of one list

PR #66, opened 2019-03-11
[EOC-56] create 404 page and set up redirection

alemikolo:
screenshot-localhost-3000-2019 03 11-12-43-13


client/src/app/components/Layout/Layout.jsx
```diff
@@ -13,6 +13,7 @@ import { UserPropType } from 'common/constants/propTypes';
 import { getCurrentUser } from 'modules/authorization/model/selectors';
 import Footer from '../Footer';
 import Notifications from 'modules/notification';
+import PageNotFound from 'common/components/PageNotFound';
```
adamklepacz: consider renaming to Page404

client/src/common/components/PageNotFound.jsx
```diff
@@ -0,0 +1,27 @@
+import React, { Fragment } from 'react';
+import { Link } from 'react-router-dom';
+
+import Toolbar from 'common/components/Toolbar';
+
+const PageNotFound = () => (
+  <Fragment>
```
adamklepacz: you don't need Fragment here

client/src/common/components/PageNotFound.scss
```diff
@@ -0,0 +1,33 @@
+.page-not-found {
+  position: absolute;
+  width: 100%;
+  height: 100%;
```
adamklepacz: you don't need these 3 properties

client/src/common/utils/fetchMethods.js
```diff
@@ -2,13 +2,19 @@ import {
...
 const handleFetchErrors = resp => {
   if (resp.status === 403) {
     setTimeout(() => {
       window.location = FRONTEND_URL;
     }, NOTIFICATION_TIMEOUT);
   }
+
+  if (resp.status === 404) {
+    history.replace('/page-not-found');
+  }
```
marekrozmus: add `return` just in case

PR #69, opened 2019-03-12
[EOC-76] Remove client server host from const

adamklepacz:
I removed const from login and logout logic from the backend. 
The only const that left on the backend is required for CORS middleware, where I have to set the origin to enable cors requests between our client app and server API. It's in app.js file.

Zrzut ekranu 2019-03-12 o 14 26 52

### DIFF
--- a/server/controllers/userAuth.js
+++ b/server/controllers/userAuth.js
@@ -9,6 +9,9 @@ const setUserAndSession = (req, res) => {
     })
   );
 
+  /**
+   * TODO Refferer is not always present on req. FIXIT
+   */
   const clientUrl = req.headers.referer;
   res.status(200);
   res.redirect(clientUrl);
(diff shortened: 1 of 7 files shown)


server/controllers/userAuth.js
```diff
@@ -10,7 +8,10 @@ const setUserAndSession = (req, res) => {
       name: req.user.displayName
     })
   );
-  res.redirect(FRONTEND_URL);
+
+  const clientUrl = req.headers.referer;
+  res.status(200);
```
marekrozmus: that is not needed - it is by default 200

server/controllers/userAuth.js
```diff
@@ -19,8 +20,8 @@ const logout = (req, res) => {
 
     res.clearCookie('connect.sid');
     res.clearCookie('user');
-    res.redirect(FRONTEND_URL);
   });
+  res.status(200).end();
```
marekrozmus: try just `res.send()`

adamklepacz:
Pull Request outdated.
I'm closing this.

PR #70, opened 2019-03-13
[EOC-56] fix and improve

alemikolo:


server/controllers/shoppingList.js
```diff
@@ -157,22 +158,30 @@ const getListData = (req, resp) => {
...
+            .exec()
+            .then(cohort => {
+              if (!cohort || cohort.isArchived) {
+                throw new ResponseError('List data not found.', 404);
+              }
+            });
+        }
+      },
+      err => {
+        if (err.name === 'CastError') {
```
adamklepacz: can you explain here a little bit?
Wy you checking is err.name is 'CastError' and not just checking if err exist?

server/controllers/shoppingList.js
```diff
@@ -194,13 +203,18 @@ const getListData = (req, resp) => {
...
     })
     .catch(err => {
-      resp.status(400).send({
-        message:
-          'An error occurred while fetching the list data. Please try again.'
-      });
+      if (err instanceof ResponseError) {
+        const { status, message } = err;
+        resp.status(status).send({ message });
+      } else {
```
adamklepacz: else not necessary here

server/controllers/shoppingList.js
```diff
@@ -194,13 +203,18 @@ const getListData = (req, resp) => {
...
+      resp.status(200).json(data);
     })
     .catch(err => {
-      resp.status(400).send({
-        message:
-          'An error occurred while fetching the list data. Please try again.'
-      });
+      if (err instanceof ResponseError) {
+        const { status, message } = err;
+        resp.status(status).send({ message });
```
adamklepacz: add return here

server/controllers/cohort.js
```diff
@@ -142,24 +143,30 @@ const getCohortDetails = (req, resp) => {
...
     })
     .catch(err => {
-      resp.status(400).send({
-        message:
-          'An error occurred while deleting the cohort. Please try again.'
-      });
+      if (err instanceof ResponseError) {
+        const { status, message } = err;
+        resp.status(status).send({ message });
+      } else {
```
adamklepacz: else not necessary

server/controllers/cohort.js
```diff
@@ -142,24 +143,30 @@ const getCohortDetails = (req, resp) => {
...
         .send({ message: `Cohort ${documentName} successfully deleted.` });
     })
     .catch(err => {
-      resp.status(400).send({
-        message:
-          'An error occurred while deleting the cohort. Please try again.'
-      });
+      if (err instanceof ResponseError) {
+        const { status, message } = err;
+        resp.status(status).send({ message });
```
adamklepacz: add return here

server/controllers/shoppingList.js
alemikolo: This code will be removed. It's unclear and not proper. It was added to handle this specific error which occurs when id is not valid (e.g. too short). Before that, in this situation catch method was handling this error, and was returning 400. This solution handle this specific error and return 404 and frontend app can react on it.
We should think about all actions which user can perform and create proper scenariouses of handling errors and when and how show to user proper message.

server/common/utilities.js
```diff
@@ -9,8 +9,10 @@ const filter = f => object =>
   );
 
 const checkRole = (idsArray, userIdFromReq) => {
-  const userId = mongoose.Types.ObjectId(userIdFromReq);
+  const userId = ObjectId(userIdFromReq);
   return idsArray.some(id => id.equals(userId));
 };
 
-module.exports = { checkRole, filter };
+const isIdValid = id => ObjectId.isValid(id);
```
adamklepacz: isValidMonogId

server/common/exceptions/HttpException.js
```diff
@@ -0,0 +1,9 @@
+class HttpException extends Error {
+  constructor(status, message) {
+    super(message);
+    this.status = status;
+    this.message = message;
```
adamklepacz: I think this not necessary here if you call super(message) from Error class

server/controllers/shoppingList.js
```diff
@@ -194,9 +208,13 @@ const getListData = (req, resp) => {
         ? { _id, cohortId, description, isAdmin, isArchived, name, products }
         : { _id, cohortId, description, isArchived, name, products };
```
adamklepacz: in previous PR this was diffrent, as I said earlier you can send isAdmin with value false or true and this ternary will be not need

PR #72, opened 2019-03-13
[EOC-89] Use common dialog in project

adamklepacz:
CreationForm and ModalForm removed and replaced by 
`<Dialog />` and` <FormDialog />.`
`<DropdownForm />` modified to user` <Form/> `instead of `<CreationForm />.`

I have also changed props name on the <Form /> to match convention from an article that Marek sent.
I've got some news that dropdowns will be replaced for dialogs in the future. So this implementation is just for the current use case.


client/src/common/components/Form.jsx
```diff
@@ -76,8 +76,8 @@ Form.propTypes = {
   defaultDescription: PropTypes.string,
   defaultName: PropTypes.string,
 
-  handleDescriptionChange: PropTypes.func.isRequired,
-  handleNameChange: PropTypes.func.isRequired,
+  onDescriptionChange: PropTypes.func.isRequired,
+  onNameChange: PropTypes.func.isRequired,
   onSubmit: PropTypes.func
```
alemikolo: whether onSubmit shouldn't be require?

client/src/modules/shopping-list/index.jsx
```diff
@@ -139,19 +139,19 @@ class ShoppingList extends Component {
...
             onConfirm={this.archiveListHandler(listId)}
           />
         )}
         {showUpdateForm && (
-          <ModalForm
+          <FormDialog
             defaultDescription={description}
             defaultName={name}
-            label="Edit list"
+            title="Edit list"
```
alemikolo: alphabetic order

client/src/modules/shopping-list/index.jsx
```diff
@@ -139,19 +139,19 @@ class ShoppingList extends Component {
         )}
         {isArchived && <ArchivedList listId={listId} name={name} />}
         {showDialogBox && (
-          <DialogBox
-            message={`Do you really want to archive the ${name} list?`}
+          <Dialog
+            title={`Do you really want to archive the ${name} list?`}
```
alemikolo: alphabetic order

client/src/common/components/DropdownForm.jsx
```diff
@@ -1,22 +1,81 @@
...
+              </button>
+            </div>
+          </div>
+          <Overlay onClick={onHide} type={OverlayStyleType.LIGHT} />
+        </Fragment>
+      )
+    );
+  }
+}
 DropdownForm.propTypes = {
```
marekrozmus: new line missing

client/src/common/components/Form.jsx
```diff
@@ -16,20 +16,20 @@ class Form extends PureComponent {
     const {
       target: { value }
     } = event;
-    const { handleNameChange } = this.props;
+    const { onNameChange } = this.props;
 
     this.setState({ name: value });
-    handleNameChange(value);
+    onNameChange && onNameChange(value);
```
marekrozmus: thos one is required in props so no need to check it here - you'll get error about missing props during component creation

client/src/common/components/Form.jsx
```diff
@@ -16,20 +16,20 @@ class Form extends PureComponent {
...
   handleDescriptionChange = event => {
     const {
       target: { value }
     } = event;
-    const { handleDescriptionChange } = this.props;
+    const { onDescriptionChange } = this.props;
 
     this.setState({ description: value });
-    handleDescriptionChange(value);
+    onDescriptionChange && onDescriptionChange(value);
```
marekrozmus: thos one is required in props so no need to check it here - you'll get error about missing props during component creation

PR #73, opened 2019-03-13
[EOC-79] Add title atributes to display tooltips

adamklepacz:
Zrzut ekranu 2019-03-13 o 14 13 08

### DIFF
--- a/client/src/common/components/Form.jsx
+++ b/client/src/common/components/Form.jsx
@@ -53,6 +53,7 @@ class Form extends PureComponent {
               className="form__input primary-input"
               onChange={this.handleNameChange}
               placeholder="Name"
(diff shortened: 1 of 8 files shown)


client/src/common/components/Form.jsx
```diff
@@ -53,6 +53,7 @@ class Form extends PureComponent {
               className="form__input primary-input"
               onChange={this.handleNameChange}
               placeholder="Name"
+              title="Please enter name here."
```
alemikolo: I'm not sure if those hints should be added to inputs (we have placehoders there). Maybe only to buttons/links on toolbar. But we will have to adjust our app to mobile and for people with impairments so we should check maybe it's right approach

PR #76, opened 2019-03-14
[EOC-98] History list on right two columns change name to done leave green tick on done items

alemikolo:
I changed item's styles a little bit to adjust its look according to Łukasz suggestions.
chrome-capture
chrome-capture (1)


client/src/modules/list/components/ItemsListItem.jsx
```diff
@@ -1,54 +1,72 @@
...
-      <span className="items-list__data">
-        <span>{name}</span>
-        <span className="items-list__author">{`Ordered by: ${authorName}`}</span>
-      </span>
-      {!archived && (
-        <VotingBox
-          voteForItem={voteForItem}
-          votesCount={votesCount}
-          whetherUserVoted={whetherUserVoted}
+class ItemsListItem extends PureComponent {
```
marekrozmus: change just to `ListItem`

client/src/app/styles/_mixins.scss
```diff
@@ -0,0 +1,5 @@
+@mixin medium-screens(){
+  @media only screen and (min-width: 800px) {
+    @content;
+  }
+}
```
adamklepacz: More flexible solution will be to define breakpoints as const in some sass file maybe _variables.scss 

```
 $small: 0px,
 $medium: 640px,
 $large: 1024px,
 $xlarge: 1200px,
 $xxlarge: 1440px,
```

and then write mixin
```
@mixin breakpoint($width) { 
   @media only screen and (min-width: $width) {
      @content;
    }
}
```

and then we will simply use 
```
@include breakpoint($medium) {
 ...some styles here
}
```

client/src/modules/list/components/ItemsContainer.scss
```diff
@@ -3,6 +3,11 @@
   background-color: $container-bg;
   margin-bottom: 20px;
   text-align: center;
+  width: 100%;
+
+  @include medium-screens {
+    width: 48%;
+  }
```
adamklepacz: and here we will have a more intuitive piece of code!
```
@include breakpoint($medium) {
   width: 48%;
}
```

client/src/app/styles/_variables.scss
```diff
@@ -0,0 +1,7 @@
+// Screen sizes
+$tiny: 360px;
+$small: 640px;
+$medium: 800px;
+$normal: 1024px;
+$large: 1200px;
+$xlarge: 1440px;
```
adamklepacz: start with $small and add extra $xxlarge variable.
convention of this variables is to keep as cloaths sizes so: s,m,l,xl, xxl etc

adamklepacz:
Zrzut ekranu 2019-03-15 o 13 38 25
Add some white space over arrow.

PR #77, opened 2019-03-18
[EOC 99] Restoring archived lists in cohorts

alemikolo:
chrome-capture


client/src/modules/cohort/index.jsx
```diff
@@ -122,8 +128,18 @@ class Cohort extends PureComponent {
     return cohortDetails && cohortDetails.isAdmin;
   };
 
+  handleArchivedListVisibility = id => () => {
+    const { areArchivedListVisible } = this.state;
+    const { fetchArchivedListsMetaData, removeArchivedLists } = this.props;
+    this.setState({ areArchivedListVisible: !areArchivedListVisible });
+    !areArchivedListVisible
+      ? fetchArchivedListsMetaData(id)
+      : removeArchivedLists();
```
adamklepacz: extract those 3 lines to other function like handleArchivedListsData = () =>
Cause right now this function is not handling only visibility but also data flow management.

client/src/modules/cohort/index.jsx
```diff
@@ -240,15 +263,18 @@ Cohort.propTypes = {
   match: RouterMatchPropType.isRequired,
 
   archiveCohort: PropTypes.func.isRequired,
+  fetchArchivedListsMetaData: PropTypes.func.isRequired,
   fetchCohortDetails: PropTypes.func.isRequired,
   fetchListsMetaData: PropTypes.func.isRequired,
+  removeArchivedLists: PropTypes.func.isRequired,
```
adamklepacz: removeArchivedListsMetaData ??

client/src/modules/cohort/index.jsx
```diff
@@ -240,15 +263,18 @@ Cohort.propTypes = {
...
   updateCohort: PropTypes.func.isRequired
 };
 
 const mapStateToProps = (state, ownProps) => ({
   cohortDetails: getCohortDetails(state, ownProps.match.params.id),
   currentUser: getCurrentUser(state),
-  lists: getCohortLists(state, ownProps.match.params.id)
+  lists: getCohortLists(state, ownProps.match.params.id),
+  archivedLists: getCohortArchivedLists(state, ownProps.match.params.id)
```
adamklepacz: sort alphabetically

client/src/modules/list/model/actionTypes.js
```diff
@@ -17,6 +17,7 @@ export const ListActionTypes = Object.freeze({
   FETCH_META_DATA_FAILURE: 'list/FETCH_META_DATA_FAILURE',
   FETCH_META_DATA_REQUEST: 'list/FETCH_META_DATA_REQUEST',
   FETCH_META_DATA_SUCCESS: 'list/FETCH_META_DATA_SUCCESS',
+  REMOVE_ARCHIVED: 'list/REMOVE_ARCHIVED',
```
adamklepacz: REMOVE_ARCHIVED_META_DATA

client/src/modules/cohort/index.jsx
```diff
@@ -122,8 +128,26 @@ class Cohort extends PureComponent {
     return cohortDetails && cohortDetails.isAdmin;
   };
 
+  handleArchivedListsVisibility = id => () => {
+    const { areArchivedListVisible } = this.state;
+    this.setState({ areArchivedListVisible: !areArchivedListVisible });
+    this.handleArchivedListsData(id);
+  };
```
adamklepacz: if we should follow our current approach there should be show and hide methods, but I would leave this toggle method just for that case, because this is toggling a view indeed.

client/src/modules/list/model/actionTypes.js
```diff
@@ -17,6 +17,7 @@ export const ListActionTypes = Object.freeze({
   FETCH_META_DATA_FAILURE: 'list/FETCH_META_DATA_FAILURE',
   FETCH_META_DATA_REQUEST: 'list/FETCH_META_DATA_REQUEST',
   FETCH_META_DATA_SUCCESS: 'list/FETCH_META_DATA_SUCCESS',
+  REMOVE_ARCHIVED_META_DATA: 'list/REMOVE_ARCHIVED_META_DATA',
```
adamklepacz: should here be request success and failure also?

client/src/modules/list/model/actions.js
```diff
@@ -115,6 +115,10 @@ const restoreListRequest = () => ({
   type: ListActionTypes.RESTORE_REQUEST
 });
 
+export const removeArchivedListsMetaData = () => ({
+  type: ListActionTypes.REMOVE_ARCHIVED_META_DATA
+});
+
```
adamklepacz: should here be success, request and failure??

alemikolo: Archived lists are removing from Redux only. We don't send a request to the server in this case. So I think there is no need to create failure, success and request actions.

client/src/modules/list/model/actionTypes.js
alemikolo: Archived lists are removing from Redux only. We don't send a request to the server in this case. So I think there is no need to create failure, success and request actions.

client/src/modules/cohort/index.jsx
alemikolo: You're right and I was wondering about it. But this is a classic toggle case. Here is one button for switching visibility and there is no other place where this method is invoke or where we can change visibility of archived cohorts.
When we are editing list or adding cohort we invoke e.g. 'hide' method in different places at different moments: on cancel button click, on escape button click, on edit method success. Invoking toggling method in those cases would be non-intuitive...

client/src/modules/list/model/selectors.js
```diff
@@ -22,7 +22,18 @@ export const getItems = (state, listId) => {
 export const getLists = state => state.lists.data;
 export const getCohortLists = (state, cohortId) =>
```
marekrozmus: please use `reselect` here

PR #84, opened 2019-03-21
[EOC-106] Data leak for voting

adamklepacz:
I've created a separated controller for voting cause it's much more complicated right now :) 


client/src/modules/list/components/ItemsList/ItemsList.jsx
```diff
@@ -41,24 +41,19 @@ class ItemsList extends Component {
   };
 
   voteForItem = item => () => {
-    const { _id, voterIds } = item;
+    const { _id, didCurrentUserVoted } = item;
```
alemikolo: there is always one user so maybe didUserVote should be enough, I think there is no need to use Voted if you use did?

client/src/modules/list/components/ItemsList/ItemsList.jsx
```diff
@@ -79,8 +74,8 @@ class ItemsList extends Component {
                 name={item.name}
                 toggleItem={this.toggleItem}
                 voteForItem={this.voteForItem(item)}
-                votesCount={item.voterIds.length}
-                whetherUserVoted={item.voterIds.includes(userId)}
+                votesCount={item.votesCount}
+                whetherUserVoted={item.didCurrentUserVoted}
```
alemikolo: maybe it will be easier to remember if it will be the same shorter name ? e.g. didUserVote={item.didUserVote}. If you will decide to change it - change it in all places.

server/controllers/list.js
```diff
@@ -111,18 +113,31 @@ const getArchivedListsMetaData = (req, resp) => {
   );
 };
 
+const checkIfCurrentUserVoted = (item, userId) =>
```
alemikolo: I wonder if  that method should be place in here. It is not a controller in fact. It probably could be defined as instance or static method for its model. Or maybe it could be placed in utilities folder like in user's helper methods case. 
https://mongoosejs.com/docs/2.7.x/docs/methods-statics.html
https://mongoosejs.com/docs/api.html#schema_Schema-method

server/controllers/list.js
```diff
@@ -111,18 +113,31 @@ const getArchivedListsMetaData = (req, resp) => {
...
+  item.voterIds.indexOf(userId) > -1;
+
+const itemToSend = (item, userId) => {
```
alemikolo: I wonder if  that method should be place in here. It is not a controller in fact. It probably could be defined as instance or static method for its model. Or maybe it could be placed in utilities folder like in user's helper methods case.
https://mongoosejs.com/docs/2.7.x/docs/methods-statics.html
https://mongoosejs.com/docs/api.html#schema_Schema-method

server/controllers/list.js
```diff
@@ -111,18 +113,31 @@ const getArchivedListsMetaData = (req, resp) => {
...
+  return {
+    ...rest,
+    didCurrentUserVoted: checkIfCurrentUserVoted(item, userId)
+  };
+};
+
 const addItemToList = (req, resp) => {
   const {
-    item: { name, isOrdered, authorName, authorId, voterIds },
+    item: { name, isOrdered, authorName, authorId },
```
alemikolo: It's not a main goal but maybe isOrdered should be removed here also (and from InputBar at the frontend). We could set it's default value to false in the model... there is no possibility to set this value during item cration.

server/controllers/list.js
```diff
@@ -144,12 +159,25 @@ const addItemToList = (req, resp) => {
...
 
       doc
-        ? resp.status(200).send(item)
+        ? resp.status(200).send(itemToSend(item, userId))
         : resp.status(404).send({ message: 'List  not found.' });
     }
   );
 };
 
+const getItemsForList = (userId, list) => {
```
alemikolo: I wonder if that method should be place in here. It is not a controller in fact. It probably could be defined as instance or static method for its model. Or maybe it could be placed in utilities folder like in user's helper methods case.
https://mongoosejs.com/docs/2.7.x/docs/methods-statics.html
https://mongoosejs.com/docs/api.html#schema_Schema-method

server/controllers/list.js
```diff
@@ -263,7 +284,53 @@ const updateListItem = (req, resp) => {
...
+        { purchaserIds: req.user._id }
+      ]
+    },
+    dataToUpdate,
+    { new: true },
+    (err, doc) => {
+      if (err) {
+        return resp.status(400).send({
+          message:
+            'An error occurred while updating the list data. Please try again.'
```
alemikolo: This controller is only for voting so maybe we could adjust messages to this action to be less generall.

client/src/modules/list/components/ItemsList/actions.js
```diff
@@ -54,11 +54,11 @@ export const toggle = (
...
 };
 
-export const vote = (itemId, listId, voterIds) => dispatch => {
+export const vote = (itemId, listId, didCurrentUserVoted) => dispatch => {
   dispatch(voteForItemRequest());
-  return patchData(`${ENDPOINT_URL}/lists/${listId}/update-item`, {
-    itemId,
-    voterIds
+  return patchData(`${ENDPOINT_URL}/lists/${listId}/vote-for-item`, {
+    didCurrentUserVoted,
```
marekrozmus: please do not use `true/false` flag - just create another api route e.g. `clear-vote` and `set-vote`

client/src/modules/list/components/ItemsList/ItemsList.jsx
marekrozmus: or just `isVoted`

server/controllers/list.js
marekrozmus: this should not have in name `current` if you pass `userId` as param. When thinking about function name try to use it in `if` statement and see how it reads. Currently it sound strange `if (checkIfCurrentUserVoted(...))`. Maybe just `if (userVoted(..))`?

server/controllers/list.js
marekrozmus: What for is the `for` in name? Could it be simpler like `getListItems` or even `listItems`?

server/controllers/list.js
```diff
@@ -190,15 +218,8 @@ const getListData = (req, resp) => {
...
-        _id,
-        adminIds,
-        cohortId,
-        description,
-        isArchived,
-        items,
-        name
-      } = list;
+      const { _id, adminIds, cohortId, description, isArchived, name } = list;
+      const items = getItemsForList(userId, list);
```
marekrozmus: Why do you get list items alredy here? If the list `isArchived` this data won't be needed.

server/controllers/list.js
```diff
@@ -263,7 +284,53 @@ const updateListItem = (req, resp) => {
...
+        return resp.status(400).send({
+          message:
+            'An error occurred while updating the list data. Please try again.'
+        });
+      }
+      const itemIndex = doc.items.findIndex(item => item._id.equals(itemId));
+      const item = doc.items[itemIndex];
+
+      doc
+        ? resp.status(200).json(itemToSend(item, userId))
```
marekrozmus: Maybe change name `itemToSend` with `responseWithItem`? And if we would have such other methods then we should caal them same like 'responseWith...`?

server/models/item.model.js
```diff
@@ -1,17 +1,19 @@
...
     authorId: { type: String, required: true },
+    authorName: { type: String, required: true },
     comment: { type: String },
     isOrdered: { type: Boolean, default: false },
     name: { type: String, required: true },
     purchaserId: { type: String },
     status: { type: String },
-    voterIds: [{ type: String }]
+    voterIds: [ObjectId],
+    votesCount: { type: Number, default: 0 }
```
marekrozmus: This should be counted in runtime - otherwise we can get out of sync.

server/controllers/list.js
```diff
@@ -157,27 +143,16 @@ const addItemToList = (req, resp) => {
             'An error occurred while adding a new item. Please try again.'
         });
       }
-
+      const listInstance = new List();
```
alemikolo: I think this was a bad idea to create instance's method (**I know that was my idea an my fault**) because in this case we don't operate on specific list's document (you have to create new one to use it)... Code after change looks strange. Maybe it would be better to move this 'helper' functions to utilities....

server/models/list.model.js
```diff
@@ -22,4 +23,28 @@ const ListSchema = new Schema(
   { timestamps: { createdAt: 'created_at' }, collection: 'lists' }
 );
 
+ListSchema.method(
```
alemikolo: I think this was a bad idea to create instance's method (**I know that was my idea an my fault**) because in this case we don't operate on specific list's document (you have to create new one to use it)... Code after change looks strange. Maybe it would be better to move this 'helper' functions to utilities....

server/routes/list.js
```diff
@@ -22,6 +22,7 @@ router.delete('/:id/delete', authorize, deleteListById);
 router.patch('/:id/update', authorize, updateListById);
 router.get('/:id/data', authorize, getListData);
 router.patch('/:id/update-item', authorize, updateListItem);
-router.patch('/:id/vote-for-item', authorize, voteForItem);
+router.patch('/:id/set-vote', authorize, voteForItem);
+router.patch('/:id/clear-vote', authorize, voteForItem);
```
alemikolo: Hmm, I know we discussed about it but it looks strange: we have two routes and one controller which check path (req.route.path.includes('set-vote')) and run appropriate logic. Maybe it would be better to use separate controllers? We have such cases: fetchListsMetaData and fetchArchiveListMetaData... and the same for cohort...

client/src/modules/list/components/ItemsList/ItemsList.jsx
```diff
@@ -41,24 +41,19 @@ class ItemsList extends Component {
...
-      currentUser: { id },
       match: {
         params: { id: listId }
       }
     } = this.props;
-    voterIds.includes(id)
-      ? vote(_id, listId, voterIds.filter(voterId => voterId !== id))
-      : vote(_id, listId, voterIds.concat(id));
+
+    vote(_id, listId, isVoted);
```
marekrozmus: Create separate method for voting and clearing the vote.

server/common/utils/utilities.js
```diff
@@ -0,0 +1,51 @@
+const { ObjectId } = require('mongoose').Types;
```
marekrozmus: naming of this file is strange - `utilities` in `utils`?

client/src/modules/list/components/ItemsList/ItemsList.jsx
```diff
@@ -41,24 +41,20 @@ class ItemsList extends Component {
...
+      setVote,
       match: {
         params: { id: listId }
       }
     } = this.props;
-    voterIds.includes(id)
-      ? vote(_id, listId, voterIds.filter(voterId => voterId !== id))
-      : vote(_id, listId, voterIds.concat(id));
+
+    isVoted ? clearVote(_id, listId, isVoted) : setVote(_id, listId, isVoted);
```
marekrozmus: - remove `isVoted` from function calls - this is not needed anymore
- please do not use ternary operator if return value is not needed
- or:
```
const action = isVoted ? clearVote : setVote;
action(_id, listId);
```

client/src/modules/list/components/ItemsList/ItemsList.jsx
```diff
@@ -79,8 +75,8 @@ class ItemsList extends Component {
                 name={item.name}
                 toggleItem={this.toggleItem}
                 voteForItem={this.voteForItem(item)}
-                votesCount={item.voterIds.length}
-                whetherUserVoted={item.voterIds.includes(userId)}
+                votesCount={item.votesCount}
+                whetherUserVoted={item.isVoted}
```
marekrozmus: change name of this prop

server/controllers/list.js
```diff
@@ -144,7 +151,7 @@ const addItemToList = (req, resp) => {
       }
 
       doc
```
marekrozmus: these all usage of ternary will have to be changed in other task

PR #85, opened 2019-03-25
[EOC-116] Archive list by link under list

alemikolo:
archive_button_under_list

### DIFF
--- a/client/src/modules/list/index.jsx
+++ b/client/src/modules/list/index.jsx
@@ -110,11 +110,6 @@ class List extends Component {
                 onClick={this.handleDialogContext(DialogContext.UPDATE)}
                 title="Edit list"
               />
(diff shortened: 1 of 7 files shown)


client/src/modules/list/index.jsx
```diff
@@ -128,6 +123,13 @@ class List extends Component {
               <InputBar />
             </ItemsContainer>
             <ItemsContainer archived items={orderedItems} />
+            <button
+              className="link-button"
+              onClick={this.handleDialogContext(DialogContext.ARCHIVE)}
+              type="button"
+            >
+              {`Archive the ${name} list`}
```
marekrozmus: surround `name` with `"` signs - this should be done everywhere we use list name, otherwise we get sometimes strange sentences

PR #86, opened 2019-03-25
[EOC-115] Privacy policy page

alemikolo:
Privacy Policy "page" is accessible via app also, because of point "Changes to this Privacy Policy" which informs user to review our privacy statement periodically to track the changes.

Login page:
login_page

Privacy Policy  accessible from login page:
privacy_policy_login

Privacy Policy  accessible from application:
privacy_policy_app

Landing page:
privacy_policy_cookies

Privacy Policy  accessible from landing page:
privacy_policy_landing


client/src/modules/privacy-policy/index.jsx
```diff
@@ -0,0 +1,153 @@
...
+  render() {
+    const { isCookieSet } = this.state;
+    const { currentUser } = this.props;
+
+    return (
+      <Fragment>
+        {currentUser && <Toolbar />}
+        <div className="privacy-policy">
+          <h1 className="privacy-policy__intro-heading">
+            Privacy &&nbsp;Terms
```
marekrozmus: replace `&nbsp` with some element with `nowrap` set

client/src/modules/privacy-policy/index.jsx
```diff
@@ -0,0 +1,153 @@
...
+          </h1>
+          <h2 className="privacy-policy__intro-subheading">
+            Check our Privacy Policy and&nbsp;Terms.
```
marekrozmus: replace `&nbsp` with some element with `nowrap` set

client/src/app/components/Footer.jsx
```diff
@@ -6,9 +6,15 @@ import { COMPANY_NAME, COMPANY_PAGE_URL } from 'common/constants/variables';
...
   <div className="footer">
     <div className="wrapper footer__wrapper">
-      <Link className="footer__text footer__link" to="/about">
-        About
-      </Link>
+      <span className="footer__text">
+        <Link className="footer__text footer__link" to="/about">
+          About
+        </Link>
+        {' | '}
```
adamklepacz: add this as a span with class .footer__text-divider

client/src/common/components/CookieConsentBox.jsx
```diff
@@ -0,0 +1,42 @@
...
+      expires={365}
+      location="bottom"
+      onAccept={onAccept}
+    >
+      We use cookies to ensure you get the best experience. By using this
+      website, you consent to use of these cookies.
+      {isAuthPage && (
+        <Fragment>
+          <br />
+          {'For more information, please see our '}
```
adamklepacz: check is those braces are needed?

client/src/common/components/CookieConsentBox.scss
```diff
@@ -37,6 +37,14 @@
     z-index: 11;
     margin: 0 16px 16px;
 
+    a {
```
adamklepacz: U can use className on `<Link/>` component

client/src/modules/authorization/AuthBox.jsx
```diff
@@ -63,23 +62,7 @@ class AuthBox extends PureComponent {
...
-              disableStyles
-              expires={365}
-              location="bottom"
-              onAccept={this.handleCookieAccept}
-            >
-              We use cookies to ensure you get the best experience. By using
-              this website, you consent to use of these cookies.
-            </CookieConsent>
-          </Fragment>
+          <CookieConsentBox isAuthPage onAccept={this.handleCookieAccept} />
```
adamklepacz: 👍

client/src/modules/privacy-policy/index.jsx
```diff
@@ -0,0 +1,153 @@
+/* eslint-disable react/jsx-one-expression-per-line */
```
adamklepacz: why did you enable ignoring this rule here?

client/src/modules/privacy-policy/index.jsx
```diff
@@ -0,0 +1,153 @@
...
+        {currentUser && <Toolbar />}
+        <div className="privacy-policy">
+          <h1 className="privacy-policy__intro-heading">
+            Privacy &&nbsp;Terms
+          </h1>
+          <h2 className="privacy-policy__intro-subheading">
+            Check our Privacy Policy and&nbsp;Terms.
+          </h2>
+          <div className="privacy-policy__info-wrapper">
+            <div className="privacy-policy__info">
```
adamklepacz: those infoboxes could be `<article>` instead of divs

client/src/modules/privacy-policy/index.jsx
```diff
@@ -0,0 +1,153 @@
...
+                for the content, privacy policies or reliability of such sites.
+              </p>
+            </div>
+          </div>
+          <div className="privacy-policy__info-wrapper">
+            <div className="privacy-policy__info">
+              <h2 className="privacy-policy__header">Contact Us</h2>
+              <p className="privacy-policy__content">
+                If you have any questions about this Privacy Policy or
+                practices, please contact us via a{' '}
```
adamklepacz: @marekrozmus when doing a space should we use` &nbsp;`  ??

landing/js/cookie-consent.js
```diff
@@ -3,6 +3,7 @@ $(window).on('load', () => {
   const cookieBar = $('[data-id="cookie-bar"]');
   const cookieButton = $('[data-id="cookie-button"]');
   const body = $('body');
+  const privacy = $('.privacy-policy');
```
adamklepacz: const privacyPage

landing/js/cookie-consent.js
```diff
@@ -40,7 +41,9 @@ $(window).on('load', () => {
     init: () => {
       if (!cookie.checkIfSet('cookie-consent')) {
         cookieMessage.open();
-        bodyElement.showOverlay();
+        if (privacy.length === 0) {
```
adamklepacz: check if you have to check length or just if(privacyPage) would be enough?

client/src/common/components/CookieConsentBox.jsx
alemikolo: I don't know if they are needed but there is no space after word 'our' without braces, and I couldn't find another solution.... this is problem that occures with more complicated text in jsx files. See correction in  privacy policy index.jsx file and my commment

client/src/modules/privacy-policy/index.jsx
alemikolo: with this rule we can't insert a components or html tags into the text. Eslint error occures. There is a lot of complicated texts now, some of them need special characters, spaces, sometimes we have to insert links into them. When you fix Eslit error a prettier error occures. I have no idea how to solve it in different way...

landing/js/cookie-consent.js
alemikolo: It wouldn't be enough because privacyPage is a collection (selecting via class attribute) so we have to check length

PR #92, opened 2019-03-27
[EOC-113] Add database seeding for dev

adamklepacz:
Usage of seeding is described in Readme. If not clear, comment with proposals :)


server/seed/seedCohorts.js
```diff
@@ -0,0 +1,48 @@
...
+
+const userId = process.env.USER_ID;
+const initialCohorts = [
+  {
+    _id: '5c9a12eb46e555ba24992a2c',
+    adminIds: [userId],
+    description: '',
+    isArchived: false,
+    memberIds: [],
+    name: 'Cohorta 1'
```
marekrozmus: remove extra 'a'

server/seed/seedCohorts.js
```diff
@@ -0,0 +1,48 @@
...
+  if (existingCohorts.length > 0) {
+    console.log(`Skip seeding -> collection ${Cohort.modelName} not empty`);
+    return;
+  }
+
+  console.log('Seeding cohorts... 🌱 🌱');
+  let counter = 0;
+  // eslint-disable-next-line no-restricted-syntax
+  for (const cohort of initialCohorts) {
+    const newCohort = new Cohort({ ...cohort });
```
marekrozmus: you are spreading and creating object again - use just `cohort`

server/seed/seedLists.js
```diff
@@ -0,0 +1,111 @@
...
+    console.log(`Skip seeding -> collection ${List.modelName} not empty`);
+    return;
+  }
+
+  console.log('Seeding lists... 🌱 🌱');
+  let counter = 0;
+  // eslint-disable-next-line no-restricted-syntax
+  for (const list of initialLists) {
+    const newList = new List({
+      ...list
```
marekrozmus: just `list`

server/seed/seedLists.js
```diff
@@ -0,0 +1,110 @@
+/* eslint-disable no-await-in-loop */
+/* eslint-disable no-console */
+const List = require('../models/list.model');
+require('dotenv').config();
+
+const userId = process.env.USER_ID;
+const initialLists = [
```
alemikolo: the same proposal like in cohorts case - move to separate file/folder...

server/seed/seedCohorts.js
```diff
@@ -0,0 +1,48 @@
...
+/* eslint-disable no-console */
+
+const Cohort = require('../models/cohort.model');
+require('dotenv').config();
+
+const userId = process.env.USER_ID;
+const initialCohorts = [
+  {
+    _id: '5c9a12eb46e555ba24992a2c',
+    adminIds: [userId],
```
alemikolo: Only this one property is constant (userId). Could we generate properties automatically? (e.g. generate cohort id - random 24 chars string) and other properties via some function or loop. And after that would get cohort from db and add list for them in the same way. I mentioned it because when we change models etc. we would have to edit whole jsons. If data would be generated automatically there will be one place to change it.

server/seed/generateLists.js
```diff
@@ -0,0 +1,86 @@
+const userId = process.env.USER_ID;
+
+const generateLists = cohortId => [
```
alemikolo: Could we generate the lists in similar way as cohorts?

server/seed/seedCohorts.js
```diff
@@ -0,0 +1,32 @@
...
+  // eslint-disable-next-line no-unused-vars
+  let counter = 0;
+  // eslint-disable-next-line no-restricted-syntax
+  for (const cohort of cohorts) {
+    const newCohort = new Cohort(cohort);
+
+    await newCohort.save();
+    counter += 1;
+  }
+  return cohorts;
```
marekrozmus: new line before `return`

PR #93, opened 2019-03-27
[EOC-119] On list creation select private public add lock icon for private list

adamklepacz:
Lists on dashboard are private by default.
List in cohorts: 
-when private - only visible to owner, 
-when public- visible to all cohort members

Zrzut ekranu 2019-03-27 o 14 05 05

### DIFF
--- a/client/src/common/components/CardItem.jsx
+++ b/client/src/common/components/CardItem.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-import { RegularStarIcon, SolidStarIcon } from 'assets/images/icons';
+import { RegularStarIcon, SolidStarIcon, LockIcon } from 'assets/images/icons';
 
 export const CardColorType = {
   BROWN: 'card/BROWN',
@@ -14,6 +14,7 @@ const CardItem = ({
   color,
   description,
   isFavourite,
+  isPrivate,
   name,
   onCardClick,
   onFavClick
@@ -27,7 +28,10 @@ const CardItem = ({
     onClick={onCardClick}
     role="figure"
   >
-    <h3 className="card-item__heading">{name}</h3>
+    <header className="card-item__header">
+      {isPrivate && <LockIcon />}
+      <h3 className="card-item__heading">{name}</h3>
+    </header>
     <p className="card-item__description">{description}</p>
     <button className="card-item__star" onClick={onFavClick} type="button">
       {isFavourite ? <SolidStarIcon /> : <RegularStarIcon />}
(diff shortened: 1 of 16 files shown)


client/src/common/components/CardItem.jsx
```diff
@@ -27,7 +28,10 @@ const CardItem = ({
     onClick={onCardClick}
     role="figure"
   >
-    <h3 className="card-item__heading">{name}</h3>
+    <header className="card-item__header">
```
alemikolo: Proposal: there is not enough space there, especially in tiles view, especially when the name will be long, maybe move lock icon to some bottom corner?

client/src/modules/cohort/index.jsx
```diff
@@ -128,6 +134,13 @@ class Cohort extends PureComponent {
       : removeArchivedListsMetaData();
   };
 
+  handleListType = isPrivate => {
```
alemikolo: This one form field is handle (manage) in different way. Could we pass it directly to handleListCreation link in name/description case?

client/src/common/components/GridList.jsx
```diff
@@ -91,6 +91,7 @@ class GridList extends PureComponent {
                   color={color}
                   description={item.description}
                   isFavourite={item.isFavourite}
+                  isPrivate={item.isPrivate}
```
marekrozmus: Stop using `is` prefix on props names.

client/src/modules/cohort/index.jsx
```diff
@@ -128,6 +134,13 @@ class Cohort extends PureComponent {
...
+    if (isPrivate === ListType.PRIVATE) {
+      this.setState({ isListPrivate: true });
+    }
+    this.setState({ isListPrivate: false });
```
marekrozmus: You are setting first to true and always to false. `else` is missing?

client/src/common/components/CardItem.jsx
adamklepacz: Client wanted to have it next to title

client/src/modules/cohort/index.jsx
adamklepacz: not really, cause then we gonna create a new list in this function and then also setting list type in the same function.
Referring to a single responsibility principle, a function should do only one thing. That's why it's separated.

client/src/common/components/GridList.jsx
adamklepacz: Zrzut ekranu 2019-04-01 o 16 41 48
I can't use only `private` word cause eslint says it is `private is a reserved word in strict mode`
How I can name it?

client/src/modules/cohort/index.jsx
adamklepacz: I don't think else is necessary here. There are only two cases public and private so why would I declare public case in else ??

marekrozmus: It will be called - after `if` block next lines will be called. Change to `this.setState({ isListPrivate: isPrivate === ListType.PRIVATE });`

PR #94, opened 2019-03-29
[EOC-120] Members component ui backend

adamklepacz:
Members component. Displaying members, adding new member when you're cohort owner. When you're not owner Notification is displayed.

Zrzut ekranu 2019-03-29 o 13 35 00
Zrzut ekranu 2019-03-29 o 13 35 07


server/controllers/cohort.js
```diff
@@ -248,13 +250,98 @@ const removeFromFavourites = (req, resp) => {
...
+    if (err) {
+      return resp.status(400).send({
+        message: "Can't get members data. Please try again."
+      });
+    }
+
+    if (doc) {
+      const { memberIds, ownerIds } = doc;
+
+      User.find({ _id: [...memberIds, ownerIds] }, (err, docs) => {
```
alemikolo: ownerIds has to be spreaded ...ownerIds

server/controllers/cohort.js
```diff
@@ -248,13 +250,98 @@ const removeFromFavourites = (req, resp) => {
...
+      Cohort.findOneAndUpdate(
+        { _id: cohortId, memberIds: { $nin: [newMemberId] } },
+        { $push: { memberIds: newMemberId } }
+      )
+        .exec()
+        .then(doc => {
+          if (doc) {
+            return resp
+              .status(200)
+              .json({ avatarUrl, displayName, email, newMemberId });
```
alemikolo: isOwner: false should be sent here, because it's needed and in other cases is sent e.g getMembers controller. Create some helper method responseWithUser like in other cases

client/src/modules/members/model/reducer.js
```diff
@@ -0,0 +1,35 @@
...
+
+const members = (state = {}, action) => {
+  switch (action.type) {
+    case MembersActionTypes.FETCH_SUCCESS:
+      return { ...state, ...action.payload };
+    case MembersActionTypes.CLEAR_DATA:
+      return {};
+    case MembersActionTypes.ADD_SUCCESS: {
+      const { avatarUrl, displayName, email, newMemberId } = action.payload;
+      return { ...state, [newMemberId]: { avatarUrl, displayName, email } };
```
alemikolo: If you change newMemberId to _id on backend, and add to sending data isOwner (see below), you could do:
```
case MembersActionTypes.ADD_SUCCESS: {
  const { _id } = action.payload;
  return { ...state, [_Id]: { ...action.payload } };
```

client/src/modules/members/model/reducer.js
```diff
@@ -0,0 +1,35 @@
...
+
+const isFetching = (state = false, action) => {
+  switch (action.type) {
+    case MembersActionTypes.ADD_SUCCESS:
+    case MembersActionTypes.ADD_FAILURE:
+    case MembersActionTypes.FETCH_SUCCESS:
+    case MembersActionTypes.FETCH_FAILURE:
+      return false;
+    case MembersActionTypes.FETCH_REQUEST:
+    case MembersActionTypes.ADD_REQUEST:
```
alemikolo: alphabetic order

client/src/modules/members/model/reducer.js
```diff
@@ -0,0 +1,35 @@
...
+      return { ...state, [newMemberId]: { avatarUrl, displayName, email } };
+    }
+    default:
+      return state;
+  }
+};
+
+const isFetching = (state = false, action) => {
+  switch (action.type) {
+    case MembersActionTypes.ADD_SUCCESS:
```
alemikolo: alphabetic order

client/src/modules/members/model/actions.js
```diff
@@ -0,0 +1,74 @@
...
+      dispatch(fetchMembersFailure());
+      createNotificationWithTimeout(
+        dispatch,
+        NotificationType.ERROR,
+        err.message || "Oops, we're sorry, fetching members failed..."
+      );
+    });
+};
+
+export const clearMembers = () => dispatch => dispatch(clearMembersRequest());
```
alemikolo: 1. I'm not sure that we have to dispatch clearMemberRequest() - we simply can invoke this action creator directly like in list's case:
```
export const removeArchivedListsMetaData = () => ({
  type: ListActionTypes.REMOVE_ARCHIVED_META_DATA
});
```
2. This action creator doesn't have to have request in it's name because there is no request here. If you want to leave this additional word change it to success, because on success we invoke actions which purpose is to modify store.

client/src/modules/cohort/index.jsx
```diff
@@ -55,6 +61,7 @@ class Cohort extends PureComponent {
       .then(() => {
         if (!this.checkIfArchived()) {
           fetchListsMetaData(id);
+          fetchCohortMembers(id);
```
marekrozmus: fetch users with cohort data

client/src/modules/members/index.jsx
```diff
@@ -0,0 +1,100 @@
...
+
+class MembersBox extends PureComponent {
+  state = {
+    isFormVisible: false
+  };
+
+  componentWillUnmount() {
+    const { clearMembers } = this.props;
+
+    clearMembers();
```
marekrozmus: remove this action - memebers should be per list/cohort

client/src/common/components/Members/index.jsx
```diff
@@ -0,0 +1,100 @@
...
+      params: { id }
+    }
+  } = ownProps;
+  return { users: getCohortMembers(state, id) };
+};
+
+export default withRouter(
+  connect(
+    mapStateToProps,
+    null
```
marekrozmus: uneeded `null`

server/common/utils/index.js
```diff
@@ -51,11 +51,30 @@ const responseWithItem = (item, userId) => {
...
+    isOwner: ownerIds.indexOf(user._doc._id.toString()) > -1
   }));
 
+const responseWithUser = data => {
+  const { avatarUrl, displayName, newMemberId } = data;
+  return {
+    _id: newMemberId,
+    avatarUrl,
+    displayName,
+    isOwner: false
```
marekrozmus: why is this always `false`?

server/controllers/cohort.js
```diff
@@ -122,41 +126,57 @@ const getCohortDetails = (req, resp) => {
...
+        message:
+          'An error occurred while fetching the cohort data. Please try again.'
+      })
+    );
 };
 
 const deleteCohortById = (req, resp) => {
   let documentName = '';
-  Cohort.findOne({ _id: req.params.id, ownerIds: req.user._id })
+  Cohort.findOne({ _id: req.params.id, owners: req.user._id })
```
marekrozmus: why these changes?

server/models/cohort.model.js
```diff
@@ -5,12 +5,12 @@ const { ObjectId } = Schema.Types;
...
   {
-    ownerIds: [ObjectId],
     description: { type: String },
     favIds: [ObjectId],
     isArchived: { type: Boolean, default: false },
-    memberIds: [ObjectId],
-    name: { type: String, required: true }
+    members: [{ type: ObjectId, ref: 'User' }],
+    name: { type: String, required: true },
+    owners: [{ type: ObjectId, ref: 'User' }]
```
marekrozmus: these are still ids

server/common/utils/index.js
adamklepacz: because when adding a new member to cohort, he is NOT the owner by default.

client/src/common/components/Members/components/MembersForm.jsx
```diff
@@ -0,0 +1,75 @@
...
+      </form>
+    );
+  }
+}
+
+MembersForm.propTypes = {
+  onAddNew: PropTypes.func.isRequired
+};
+
+export default MembersForm;
```
alemikolo: How can we close/hide this form?

client/src/modules/cohort/index.jsx
```diff
@@ -128,6 +130,16 @@ class Cohort extends PureComponent {
       : removeArchivedListsMetaData();
   };
 
+  handleAddNewMember = email => {
```
alemikolo: can we handle this feature inside membersBox to keep all members actions in common/members components? e.g removing, changing roles are perform inside MemberBox... You will have access to the cohortId or listId there via route params...

server/common/utils/index.js
```diff
@@ -51,11 +51,30 @@ const responseWithItem = (item, userId) => {
...
+    };
+  });
+
+const responseWithUsers = (users, ownerIds) =>
+  users.map(user => ({
+    ...user._doc,
+    isOwner: ownerIds.indexOf(user._doc._id.toString()) > -1
   }));
 
+const responseWithMember = (data, ownerIds) => {
```
alemikolo: names: responseWithMembers but above is respnseWithUsers

client/src/common/components/Members/components/MembersForm.jsx
adamklepacz: this component is integral part of MembersBox and should not be used independent anywhere, so it don't have any show hide methods.

client/src/modules/cohort/index.jsx
adamklepacz: I don't think it is a good idea cause MemberBox is a generic component and it shouldn't now about business logic of cohort or lists. If I do your proposal I will have to fire addCohortMember inside MembersBox component.

I think that adding cohort member should be handled within the cohort module.

PR #100, opened 2019-04-04
[EOC-91] inline edit name description for cohort

adamklepacz:
chrome-capture

Update of title and description is perfomed, when click event is captured away from the element  which is currently in the edition or when "Enter" key is pressed on the keyboard.

Update fields are hiding when the click event is captured outside element which is currently in the edition.

### DIFF
--- a/client/src/modules/cohort/components/CohortHeader.jsx
+++ b/client/src/modules/cohort/components/CohortHeader.jsx
@@ -53,6 +53,7 @@ class CohortHeader extends PureComponent {
     ) {
       this.setState({ isDescriptionTextareaVisible: false });
       this.handleDescriptionUpdate(id)();
+      return;
     }
 
     if (
@@ -83,19 +84,15 @@ class CohortHeader extends PureComponent {
     }
   };
 
-  handleNameInputVisibility = () => {
-    const { isNameInputVisible } = this.state;
+  handleNameInputVisibility = () =>
+    this.setState(({ isNameInputVisible }) => ({
+      isNameInputVisible: !isNameInputVisible
+    }));
 
-    this.setState({ isNameInputVisible: !isNameInputVisible });
-  };
-
-  handleDescriptionTextareaVisibility = () => {
-    const { isDescriptionTextareaVisible } = this.state;
-
-    this.setState({
+  handleDescriptionTextareaVisibility = () =>
+    this.setState(({ isDescriptionTextareaVisible }) => ({
       isDescriptionTextareaVisible: !isDescriptionTextareaVisible
-    });
-  };
+    }));
 
   handleNameChange = event => {
     const {
@@ -114,9 +111,15 @@ class CohortHeader extends PureComponent {
   };
 
   handleNameUpdate = cohortId => () => {
-    const { updateCohort } = this.props;
+    const { updateCohort, details } = this.props;
     const { name } = this.state;
     const nameToUpdate = name.trim();
+    const { name: previousName } = details;
+
+    if (previousName === name) {
+      this.setState({ isNameInputVisible: false });
+      return;
+    }
 
     if (nameToUpdate.length >= 1) {
       updateCohort(cohortId, { name });
@@ -125,8 +128,14 @@ class CohortHeader extends PureComponent {
   };
 
   handleDescriptionUpdate = cohortId => () => {
-    const { updateCohort } = this.props;
+    const { updateCohort, details } = this.props;
     const { description } = this.state;
+    const { description: previousDescription } = details;
+
+    if (previousDescription === description) {
+      this.setState({ isDescriptionTextareaVisible: false });
+      return;
+    }
 
     updateCohort(cohortId, { description });
     this.setState({ isDescriptionTextareaVisible: false });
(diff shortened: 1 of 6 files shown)


client/src/modules/cohort/components/CohortHeader.jsx
```diff
@@ -0,0 +1,213 @@
...
+      }
+    } = this.props;
+
+    if (
+      isDescriptionTextareaVisible &&
+      !this.descriptionTextarea.current.contains(event.target)
+    ) {
+      this.setState({ isDescriptionTextareaVisible: false });
+      this.handleDescriptionUpdate(id)();
+    }
```
marekrozmus: add return or else if

client/src/modules/cohort/components/CohortHeader.jsx
```diff
@@ -0,0 +1,213 @@
...
+        : this.handleNameUpdate;
+
+      action(id)();
+    }
+  };
+
+  handleNameInputVisibility = () => {
+    const { isNameInputVisible } = this.state;
+
+    this.setState({ isNameInputVisible: !isNameInputVisible });
```
marekrozmus: `this.setState(({ isNameInputVisible }) => ({ isNameInputVisible: !isNameInputVisible });`

client/src/modules/cohort/components/CohortHeader.jsx
```diff
@@ -0,0 +1,213 @@
...
+    this.setState({ description: value });
+  };
+
+  handleNameUpdate = cohortId => () => {
+    const { updateCohort } = this.props;
+    const { name } = this.state;
+    const nameToUpdate = name.trim();
+
+    if (nameToUpdate.length >= 1) {
+      updateCohort(cohortId, { name });
```
marekrozmus: do not update if same as original

client/src/modules/cohort/components/CohortHeader.jsx
```diff
@@ -0,0 +1,213 @@
...
+  handleNameInputVisibility = () => {
+    const { isNameInputVisible } = this.state;
+
+    this.setState({ isNameInputVisible: !isNameInputVisible });
+  };
+
+  handleDescriptionTextareaVisibility = () => {
+    const { isDescriptionTextareaVisible } = this.state;
+
+    this.setState({
```
alemikolo: same as above:
this.setState(({ isDescriptionTextareaVisible }) => ({ isDescriptionTextareaVisible: ! isDescriptionTextareaVisible });

client/src/modules/cohort/components/CohortHeader.jsx
```diff
@@ -0,0 +1,213 @@
+import React, { Fragment, PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import { withRouter } from 'react-router-dom';
+import { connect } from 'react-redux';
+
+import { CohortIcon } from 'assets/images/icons';
+import { updateCohort } from '../model/actions';
+import { RouterMatchPropType } from 'common/constants/propTypes';
+
+class CohortHeader extends PureComponent {
```
alemikolo: Proposal: This feature will appear in list also. Do we want to have two separate components CohortHeader and ListHeader with almost the same logic? Can we move some common logic to some more general Header component? We could do this while working on ListHeader... but the the will be more to do...

client/src/modules/cohort/components/CohortHeader.jsx
```diff
@@ -0,0 +1,213 @@
...
+      isNameInputVisible,
+      name
+    } = this.state;
+
+    return (
+      <header className="cohort-header">
+        <div className="cohort-header__top">
+          <CohortIcon />
+          {isNameInputVisible ? (
+            <input
```
alemikolo: inputs should have name attribute

client/src/modules/cohort/components/CohortHeader.jsx
```diff
@@ -0,0 +1,213 @@
...
+            <h1
+              className="cohort-header__heading"
+              onClick={this.handleNameInputVisibility}
+            >
+              {name}
+            </h1>
+          )}
+        </div>
+        {isDescriptionTextareaVisible ? (
+          <textarea
```
alemikolo: textarea should have name attribute

client/src/modules/cohort/components/CohortHeader.jsx
adamklepacz: It's not so general, because those two areas look different in the list and in the cohort. 
And I don't want to mix it up with list cause list have much more options like private public, and there can be more changes in future in list view. So we would perhaps change this back to ListHeader and Cohortheader in future, because of coming changes and ideas.

adamklepacz:
> General comments:
> 
> * I run branch this eoc-91 branch locally and I realized that this add description button is too visible'... Łukasz mentioned that he's wondering if this information is necessary so maybe add description button should look like e.g link-button (archive list). Now it looks like too much important,
> * editing header is not accessible via keyboard - you can't tab on it so you can't edit it, if you decide to change it - read first the article: [tabindex](https://developers.google.com/web/fundamentals/accessibility/focus/using-tabindex), because some usecases are antipattern
> * when you click header the the text header jumps - could it be done without this jumping? In other sites the text doesn't jump. I don't know if this is right and proper solution but maybe don't toggle between h1 and input but use input inside header and only toggle input's classes: edited, not-edited? The text won't jump and additionally input will be always visible an will be accessible via keyboard...

This needs to be done in separated tasks...it's harder then I thought.

PR #102, opened 2019-04-05
[EOC-126] Display archive lists below active lists on dashboard

alemikolo:
chrome-capture (1)

### DIFF
--- a/client/src/modules/dashboard/index.jsx
+++ b/client/src/modules/dashboard/index.jsx
@@ -26,7 +26,7 @@ import { Routes } from 'common/constants/enums';
 
 class Dashboard extends Component {
   state = {
-    areArchivedListVisible: false,
+    showArchivedLists: false,
     dialogContext: ''
   };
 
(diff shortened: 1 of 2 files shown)


client/src/modules/dashboard/index.jsx
```diff
@@ -59,9 +65,27 @@ class Dashboard extends Component {
...
+    this.handleArchivedListsData();
+  };
+
+  handleArchivedListsData = id => {
+    const { areArchivedListVisible } = this.state;
+    const {
+      fetchArchivedListsMetaData,
+      removeArchivedListsMetaData
+    } = this.props;
+    !areArchivedListVisible
```
adamklepacz: as Marek said, no teranry when nothing is return 
do something like. And try start with positive condition first not negative `!`
```
const action = areAchivedListVisible ? removeArchivedListsMetaData : fetchArchivedListsMetaData
action();
```

client/src/modules/dashboard/index.jsx
```diff
@@ -59,9 +65,27 @@ class Dashboard extends Component {
 
   hideDialog = () => this.handleDialogContext(null)();
 
+  handleArchivedListsVisibility = () => {
+    this.setState(({ areArchivedListVisible }) => ({
+      areArchivedListVisible: !areArchivedListVisible
+    }));
+    this.handleArchivedListsData();
```
marekrozmus: this method uses `areArchivedListVisible` from state. To be sure that value will be set in state already use `setState` callback param to call this method

client/src/modules/dashboard/index.jsx
```diff
@@ -21,6 +26,7 @@ import { Routes } from 'common/constants/enums';
 
 class Dashboard extends Component {
   state = {
+    areArchivedListVisible: false,
```
marekrozmus: maybe change to simple `showArchivedLists`?

client/src/modules/dashboard/index.jsx
```diff
@@ -59,9 +65,31 @@ class Dashboard extends Component {
 
   hideDialog = () => this.handleDialogContext(null)();
 
+  handleArchivedListsVisibility = () => {
```
adamklepacz: unecessary braces

PR #106, opened 2019-04-08
[EOC-125] inline edit title description for list

adamklepacz:
In this pull request I have also fixed a bug in cohort header ---> adding description or name with spaces only.

Zrzut ekranu 2019-04-08 o 12 17 25

### DIFF
--- a/client/src/common/components/DescriptionTextarea.jsx
+++ b/client/src/common/components/DescriptionTextarea.jsx
@@ -9,14 +9,14 @@ class DescriptionTextarea extends PureComponent {
   }
 
   componentDidMount() {
-    document.addEventListener('keydown', this.handleKeyPress);
+    document.addEventListener('keydown', this.onKeyPress);
     document.addEventListener('mousedown', this.handleClick);
 
     this.descriptionTextarea.current.focus();
   }
 
   componentWillUnmount() {
-    document.removeEventListener('keydown', this.handleKeyPress);
+    document.removeEventListener('keydown', this.onKeyPress);
     document.removeEventListener('mousedown', this.handleClick);
   }
 
@@ -29,19 +29,19 @@ class DescriptionTextarea extends PureComponent {
     handleClick(event, isClickedOutside);
   };
 
-  handleKeyPress = event => {
-    const { handleKeyPress } = this.props;
+  onKeyPress = event => {
+    const { onKeyPress } = this.props;
 
-    handleKeyPress(event);
+    onKeyPress(event);
   };
 
   render() {
-    const { description, handleDescriptionChange } = this.props;
+    const { description, onDescriptionChange } = this.props;
     return (
       <textarea
         className="desc-textarea primary-textarea"
         name="description"
-        onChange={handleDescriptionChange}
+        onChange={onDescriptionChange}
         ref={this.descriptionTextarea}
         type="text"
         value={description}
@@ -54,8 +54,8 @@ DescriptionTextarea.propTypes = {
   description: PropTypes.string.isRequired,
 
   handleClick: PropTypes.func.isRequired,
-  handleDescriptionChange: PropTypes.func.isRequired,
-  handleKeyPress: PropTypes.func.isRequired
+  onDescriptionChange: PropTypes.func.isRequired,
+  onKeyPress: PropTypes.func.isRequired
 };
 
 export default DescriptionTextarea;
--- a/client/src/common/components/DescriptionTextarea.jsx
+++ b/client/src/common/components/DescriptionTextarea.jsx
@@ -1,0 +1,61 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+
+class DescriptionTextarea extends PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.descriptionTextarea = React.createRef();
+  }
+
+  componentDidMount() {
+    document.addEventListener('keydown', this.handleKeyPress);
+    document.addEventListener('mousedown', this.handleClick);
+
+    this.descriptionTextarea.current.focus();
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keydown', this.handleKeyPress);
+    document.removeEventListener('mousedown', this.handleClick);
+  }
+
+  handleClick = event => {
+    const { handleClick } = this.props;
+    const isClickedOutside = !this.descriptionTextarea.current.contains(
+      event.target
+    );
+
+    handleClick(event, isClickedOutside);
+  };
+
+  handleKeyPress = event => {
+    const { handleKeyPress } = this.props;
+
+    handleKeyPress(event);
+  };
+
+  render() {
+    const { description, handleDescriptionChange } = this.props;
+    return (
+      <textarea
+        className="desc-textarea primary-textarea"
+        name="description"
+        onChange={handleDescriptionChange}
+        ref={this.descriptionTextarea}
+        type="text"
+        value={description}
+      />
+    );
+  }
+}
+
+DescriptionTextarea.propTypes = {
+  description: PropTypes.string.isRequired,
+
+  handleClick: PropTypes.func.isRequired,
+  handleDescriptionChange: PropTypes.func.isRequired,
+  handleKeyPress: PropTypes.func.isRequired
+};
+
+export default DescriptionTextarea;
--- a/client/src/common/components/DescriptionTextarea.jsx
+++ b/client/src/common/components/DescriptionTextarea.jsx
@@ -10,23 +10,23 @@ class DescriptionTextarea extends PureComponent {
 
   componentDidMount() {
     document.addEventListener('keydown', this.onKeyPress);
-    document.addEventListener('mousedown', this.handleClick);
+    document.addEventListener('mousedown', this.onClick);
 
     this.descriptionTextarea.current.focus();
   }
 
   componentWillUnmount() {
     document.removeEventListener('keydown', this.onKeyPress);
-    document.removeEventListener('mousedown', this.handleClick);
+    document.removeEventListener('mousedown', this.onClick);
   }
 
-  handleClick = event => {
-    const { handleClick } = this.props;
+  onClick = event => {
+    const { onClick } = this.props;
     const isClickedOutside = !this.descriptionTextarea.current.contains(
       event.target
     );
 
-    handleClick(event, isClickedOutside);
+    onClick(event, isClickedOutside);
   };
 
   onKeyPress = event => {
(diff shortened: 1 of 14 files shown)


client/src/modules/cohort/components/CohortHeader.jsx
```diff
@@ -6,6 +6,7 @@ import { connect } from 'react-redux';
 import { CohortIcon } from 'assets/images/icons';
 import { updateCohort } from '../model/actions';
 import { RouterMatchPropType } from 'common/constants/propTypes';
+import { whiteSpaceOnly } from 'common/utils/helpers';
```
alemikolo: space is always "white" character, if there would be more white characters, this method could be named whiteCharsOnly, but in this case maybe it's better to change it's name to spacesOnly?

client/src/modules/cohort/components/CohortHeader.jsx
```diff
@@ -128,22 +125,33 @@ class CohortHeader extends PureComponent {
...
+      details,
+      match: {
+        params: { id }
+      }
+    } = this.props;
     const { description } = this.state;
     const { description: previousDescription } = details;
 
-    if (previousDescription === description) {
+    if (previousDescription.trim() === description.trim()) {
```
alemikolo: is this necessary to trim previous values, which should be trimmed already?

client/src/modules/cohort/components/CohortHeader.jsx
```diff
@@ -188,7 +196,7 @@ class CohortHeader extends PureComponent {
           />
         ) : (
           <Fragment>
-            {description && (
+            {description.trim() && (
```
alemikolo: maybe use teranry operator here:  description.trim() ? (<p><p/>) : (<button></button>)

client/src/modules/cohort/components/CohortHeader.jsx
```diff
@@ -197,9 +205,9 @@ class CohortHeader extends PureComponent {
                 {description}
               </p>
             )}
-            {!description && (
+            {!description.trim() && (
```
alemikolo: is this necessary to trim description value in render? Could we do this before it's stored in state?

client/src/modules/list/components/ListHeader.jsx
```diff
@@ -0,0 +1,237 @@
...
+
+  handleDescriptionChange = event => {
+    const {
+      target: { value }
+    } = event;
+
+    this.setState({ description: value });
+  };
+
+  handleKeyPress = event => {
```
alemikolo: this handler is invoking every time you press enter or escape regardless of whether we are in edit mode or not, check the same issue in cohortHeader

client/src/modules/list/components/ListHeader.jsx
```diff
@@ -0,0 +1,237 @@
...
+    if (code === 'Enter' || code === 'Escape') {
+      const action = isDescriptionTextareaVisible
+        ? this.handleDescriptionUpdate
+        : this.handleNameUpdate;
+
+      action();
+    }
+  };
+
+  handleClick = event => {
```
alemikolo: this handler is invoking every time you click and anywhere you click, check the same issue in cohortHeader

client/src/modules/list/components/ListHeader.jsx
```diff
@@ -0,0 +1,237 @@
...
+    this.setState(({ isDescriptionTextareaVisible }) => ({
+      isDescriptionTextareaVisible: !isDescriptionTextareaVisible
+    }));
+
+  handleNameChange = event => {
+    const {
+      target: { value }
+    } = event;
+
+    this.setState({ name: value });
```
alemikolo: trim the name here, before you save it in state, and remove trimming from render method

client/src/modules/list/components/ListHeader.jsx
```diff
@@ -0,0 +1,237 @@
...
+  };
+
+  handleDescriptionChange = event => {
+    const {
+      target: { value }
+    } = event;
+
+    this.setState({ description: value });
```
alemikolo: trim the description here, before you save it in state, and remove trimming from render method

client/src/modules/cohort/components/CohortHeader.jsx
adamklepacz: I meant space as `przestrzeń` not as space key. 

https://developer.mozilla.org/pl/docs/Web/CSS/white-space

white space means basically all white characters

client/src/modules/cohort/components/CohortHeader.jsx
adamklepacz: https://eslint.org/docs/rules/no-nested-ternary

it would be nesting ternary inside ternary.

client/src/modules/list/components/ListHeader.jsx
adamklepacz: what about spaces in names? 
If I want to set ` My Cohort `. 
If I trim it here I can only get `MyCohort`

client/src/modules/list/components/ListHeader.jsx
adamklepacz: same as above

client/src/modules/list/components/ListHeader.jsx
alemikolo: The trim() method removes whitespace from both sides of a string.
[trim](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim)

client/src/modules/cohort/components/CohortHeader.jsx
adamklepacz: yes it's necessary

client/src/modules/list/components/ListHeader.jsx
adamklepacz: Exactly :) 
When writing `Name<space>` space is at the end so I can't perform adding space and the second word.

I got NameUpdated instead of Name Updated.

client/src/common/utils/helpers.js
```diff
@@ -0,0 +1,5 @@
+/**
+ *  Function checks if string contains only spaces, line breaks,
+ *  white spaces and returns true if so.
+ */
+export const whiteSpaceOnly = string => !string.replace(/\s/g, '').length;
```
marekrozmus: maybe just use: https://lodash.com/docs/4.17.11#trim and isEmpty?

client/src/modules/list/components/ListHeader.jsx
```diff
@@ -0,0 +1,238 @@
...
+import { updateList } from 'modules/list/model/actions';
+import { RouterMatchPropType } from 'common/constants/propTypes';
+import { whiteSpaceOnly } from 'common/utils/helpers';
+
+class ListHeader extends PureComponent {
+  constructor(props) {
+    super(props);
+
+    const { details } = this.props;
+    const { description, name } = details;
```
marekrozmus: this could be one line if you dont use `details`:
```const { details: { description, name } } = this.props;```

File: client/src/modules/cohort/components/CohortHeader.jsx [hunk 11]
```diff
@@ -6,16 +6,18 @@ import { connect } from 'react-redux';
...
 import { updateCohort } from '../model/actions';
 import { RouterMatchPropType } from 'common/constants/propTypes';
+import { whiteSpaceOnly } from 'common/utils/helpers';
 
 class CohortHeader extends PureComponent {
   constructor(props) {
     super(props);
 
     const { details } = this.props;
     const { name, description } = details;
```
marekrozmus: if `details` is not used it can be one line:
```const { details: { name, description } } = this,.props;```

File: client/src/modules/list/components/ListHeader.jsx [hunk 12]
```diff
@@ -0,0 +1,238 @@
...
+            className="cohort-header__desc-textarea primary-textarea"
+            name="description"
+            onChange={this.handleDescriptionChange}
+            ref={this.descriptionTextarea}
+            type="text"
+            value={description}
+          />
+        ) : (
+          <Fragment>
+            {description && (
```
marekrozmus: why not use ternary operator here? You've got true and false values checked and the return value is used.

client/src/common/utils/helpers.js
adamklepacz: it doesn't work with lodash _isEmpty

client/src/modules/list/components/ListHeader.jsx
adamklepacz: Eslint says, do not nest ternary 
https://eslint.org/docs/rules/no-nested-ternary

Should I put ternary here anyway?

client/src/common/components/DescriptionTextarea.jsx
```diff
@@ -0,0 +1,61 @@
...
+    );
+  }
+}
+
+DescriptionTextarea.propTypes = {
+  description: PropTypes.string.isRequired,
+
+  handleClick: PropTypes.func.isRequired,
+  handleDescriptionChange: PropTypes.func.isRequired,
+  handleKeyPress: PropTypes.func.isRequired
```
marekrozmus: didn't we agree that these will have prefix `on`?

client/src/common/components/NameInput.jsx
```diff
@@ -0,0 +1,60 @@
...
+    );
+  }
+}
+
+NameInput.propTypes = {
+  name: PropTypes.string.isRequired,
+
+  handleClick: PropTypes.func.isRequired,
+  handleKeyPress: PropTypes.func.isRequired,
+  handleNameChange: PropTypes.func.isRequired
```
marekrozmus: same here with prefix

client/src/modules/list/components/ListHeader.jsx
```diff
@@ -0,0 +1,225 @@
...
+        )}
+      </Fragment>
+    );
+  };
+
+  renderName = () => {
+    const { name, isNameInputVisible } = this.state;
+
+    return (
+      <Fragment>
```
alemikolo: Consider small refactoring:
```
renderDescription = () => {
    const { description, isDescriptionTextareaVisible } = this.state;

    return isDescriptionTextareaVisible ? (
      <DescriptionTextarea
        description={description}
        handleDescriptionChange={this.handleDescriptionChange}
        handleClick={this.handleClick}
        handleKeyPress={this.handleKeyPress}
      />
    ) : (
      this.someHelperRenderMethod()
    );
  };

```

client/src/common/utils/helpers.js
marekrozmus: _isEmpty(_trim(value)) does not work?

PR #110, opened 2019-04-15
[EOC-108] Add item details

adamklepacz:
**Be aware that this branch contains EOC-104.**

Zrzut ekranu 2019-04-15 o 12 40 29
Zrzut ekranu 2019-04-15 o 12 40 40

I have added Save button and there is no support for ENTER or ESCAPE, because:
-we have multiple fields and it may be confusing to the user which enter updates which fields,
-updating textarea by enter press is not a good idea in my opinion, because what about the case when the user wants to write description where is an empty line, like so:
``` 
Lorem ipsum dolor sit amet.Lorem ipsum dolor sit amet.
Lorem ipsum dolor sit amet.Lorem ipsum dolor sit amet.

P.S Please rembember about this"
```

Referring to updating textarea filed by hitting enter, we should also remove update by enter press from "description field" of list and cohort (it's gonna be simple to implement). - Of course, if you agree with my opinion 😄 
I did some research about forms but didn't found any solution about how to handle updates in forms, inputs, etc. But checking in other apps, eg. Trello.com, on fields where there is a textarea like description fields, they use BUTTON to save the value.
**This task also includes a solution for EOC 1️⃣ 0️⃣ 9️⃣, editing fields.**


client/src/modules/list/components/ListItem.jsx
```diff
@@ -41,16 +60,118 @@ class ListItem extends PureComponent {
...
+    const { itemDescription } = this.state;
+    const {
+      addItemDescription,
+      data: { _id: itemId },
+      match: {
+        params: { id: listId }
+      }
+    } = this.props;
+
+    addItemDescription(listId, itemId, itemDescription);
```
alemikolo: on one click save button, we may have to API calls at the same time to update one item: addItemDescription, addItemLink. I Wonder if it could be done with one request like in creating a list/cohort case? 
What if both failed will we receive two notifications?

client/src/modules/list/model/reducer.js
```diff
@@ -37,6 +34,39 @@ const items = (state, action) => {
...
         )
       };
+    case ItemActionTypes.ADD_DESCRIPTION_SUCCESS: {
+      const {
+        payload: { description, itemId }
+      } = action;
+
+      return {
+        ...state,
+        items: state.items.map(item =>
```
alemikolo: What do you think about storing items collection in redux as an object which properties that are items' identifiers like in list and cohort case? We will have quick access to the specific item and we won't have to map overall collection?

client/src/common/utils/helpers.js
```diff
@@ -0,0 +1,15 @@
...
+  const pattern = new RegExp(
+    '^(https?:\\/\\/)?' +
+      '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' +
+      '((\\d{1,3}\\.){3}\\d{1,3}))' +
+      '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' +
+      '(\\?[;&a-z\\d%_.~+=-]*)?' +
+      '(\\#[-a-z\\d_]*)?$',
+    'i'
+  );
+  return !!pattern.test(string);
```
marekrozmus: ? doesn't `test` method returns boolean already? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test

client/src/modules/list/components/ListItem.jsx
```diff
@@ -41,16 +60,118 @@ class ListItem extends PureComponent {
     // Adding new comment will be handled in next tasks
   };
 
+  handleDataUpdate = () => {
+    const { itemDescription, link } = this.state;
+    const {
+      data: { description: previousDescription, link: previousLink }
+    } = this.props;
+    const isLinkUpdated = !areStringsEqual(previousLink, link);
+    const isDescriptionUpdated = !areStringsEqual(
```
marekrozmus: why not using: https://lodash.com/docs/4.17.11#isEqual? Why creating own function for comparison?

client/src/modules/list/components/ListItem.jsx
```diff
@@ -41,16 +60,118 @@ class ListItem extends PureComponent {
...
+      previousDescription,
+      itemDescription
+    );
+
+    if (isLinkUpdated || isDescriptionUpdated) {
+      this.setState({ areFieldsUpdated: true });
+      return;
+    }
+
+    this.setState({ areFieldsUpdated: false });
```
marekrozmus: from 126 to 131 coudl be replaced with:
`this.setState({ areFieldsUpdated: isLinkUpdated || isDescriptionUpdated });

client/src/modules/list/model/actions.js
```diff
@@ -453,3 +480,53 @@ export const changeRole = (listId, userId, role) => dispatch => {
       );
     });
 };
+
+export const addItemDescription = (listId, itemId, description) => dispatch => {
```
marekrozmus: maybe single route for updating `details` instead of these two?

server/routes/list.js
```diff
@@ -39,5 +41,7 @@ router.patch('/:id/remove-member', authorize, removeMember);
 router.patch('/:id/change-to-owner', authorize, changeToOwner);
 router.patch('/:id/change-to-member', authorize, changeToMember);
 router.patch('/:id/add-member', authorize, addMember);
+router.patch('/:id/add-item-description', authorize, addItemDescription);
+router.patch('/:id/add-item-link', authorize, addItemLink);
```
marekrozmus: single route for `update-details`?

client/src/common/utils/helpers.js
adamklepacz: yes, it returns, but the `isUrlValid` function is surrounded by braces so I need to type return here

client/src/modules/list/model/reducer.js
adamklepacz: yes, we can...but I dont have time now to change this everywhere in reducers actions etc....
We can make separate task to refactore this

client/src/modules/list/model/actions.js
```diff
@@ -174,6 +175,19 @@ const changeRoleSuccess = (listId, userId, isOwner) => ({
...
+const addItemDetailsSuccess = (listId, itemId, data) => ({
+  type: ItemActionTypes.ADD_DETAILS_SUCCESS,
+  payload: { listId, itemId, data }
+});
+
+const addItemDetailsRequest = () => ({
+  type: ItemActionTypes.ADD_DETAILS_REQUEST
+});
+
+const addItemDetailsFailure = () => ({
```
marekrozmus: add? should be update

client/src/modules/list/model/reducer.js
```diff
@@ -203,6 +229,8 @@ const isFetching = (state = false, action) => {
     case ListActionTypes.UPDATE_FAILURE:
     case ListActionTypes.UPDATE_SUCCESS:
       return false;
+    case ItemActionTypes.ADD_DETAILS_FAILURE:
+    case ItemActionTypes.ADD_DETAILS_REQUEST:
```
marekrozmus: change types also from ADD to UPDATE

server/controllers/list.js
```diff
@@ -699,6 +682,43 @@ const addMember = (req, resp) => {
     });
 };
 
+const updateItemDetails = (req, resp) => {
+  const { description, link, itemId } = req.body;
+  const { id: listId } = req.params;
+
+  const propertiesToUpdate = {};
+  if (description !== undefined) {
```
marekrozmus: `if (description)` should be enough

client/src/common/utils/helpers.js
marekrozmus: remove `!!`

PR #112, opened 2019-04-16
[EOC-143] - Change responsive behaviour on mobile

adamklepacz:
On mobile, box is displayed below users list.
On the desktop, it is as it was previously.

Zrzut ekranu 2019-04-16 o 10 55 15
Zrzut ekranu 2019-04-16 o 10 55 21

### DIFF
--- a/client/src/common/components/Members/index.jsx
+++ b/client/src/common/components/Members/index.jsx
@@ -2,7 +2,7 @@ import React, { PureComponent } from 'react';
 import { connect } from 'react-redux';
 import _map from 'lodash/map';
 import PropTypes from 'prop-types';
-import { Manager } from 'react-popper';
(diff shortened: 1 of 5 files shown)


client/src/common/components/Members/index.jsx
```diff
@@ -16,8 +17,26 @@ import { Routes } from 'common/constants/enums';
...
+
+  componentDidMount() {
+    window.addEventListener('resize', _debounce(this.handleResize, 100));
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.handleResize);
+  }
+
+  handleResize = () => {
```
marekrozmus: ```this.SetState({ isMobile: window.outerWidth < 400 })```

PR #118, opened 2019-04-19
[EOC-142] Cloning item

alemikolo:
1.  If cloning the last item causes that the items count is above the limit of displayed items, this cloned item is being unmounted. Since there is still async action in progress we get warning in the browser. To prevent it I reject pending promises in componentWillUnmount. This solution is based on (https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html) and (https://medium.com/trabe/avoid-updates-on-unmounted-react-components-2fbadab17ad2)

2. Because of mess with items actions (some of them was defined in the list's actions file, some in other files), I made some cleanup. I created Items directory in list/components and moved there (via git mv) files and directories related to items. I created a model directory and put in it action.js, actionTypes.js reducer.js files.

cloning_item


client/src/common/components/CardItem.scss
```diff
@@ -31,6 +31,8 @@
     svg {
       height: 12px;
       width: 12px;
+      min-height: 12px;
+      min-width: 12px;
```
adamklepacz: is this necessary when height and width are 'hard set' above??

client/src/common/components/Preloader.jsx
```diff
@@ -12,34 +12,60 @@ export const PreloaderSize = {
...
+
+  componentWillUnmount() {
+    clearTimeout(this.timer);
+  }
+
+  enableAnimation = () => this.setState({ displayAnimation: true });
+
+  render() {
+    const { message, size, theme } = this.props;
+
```
adamklepacz: remove a line of the break here

client/src/common/components/Toolbar/components/UserBar/UserBar.scss
```diff
@@ -94,6 +94,7 @@
   }
 
   &__menu-logout {
+    font-family: inherit;
```
adamklepacz: why do you need this?? 
Did font was different from other elements on the app??

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -130,17 +147,45 @@ class ListItem extends PureComponent {
...
+      .then(() => {
+        this.setState({ pending: false });
+        this.removePendingPromise(abortableCloning);
+      })
+      .catch(err => {
+        if (!(err instanceof AbortPromiseException)) {
+          this.setState({ pending: false });
+          this.removePendingPromise(abortableCloning);
+        }
+      });
```
adamklepacz: why this have to be done this way with 'abortable promises'
can it just be done like in any other actions ??

`cloneItem(listId, itemId).then(()=> //do something).catch(err)`

client/src/modules/list/components/Items/ListItem/ListItem.scss
```diff
@@ -153,6 +153,15 @@
     padding-bottom: 16px;
   }
 
+  &__cloning {
+    position: relative;
+    display: flex;
+    justify-content: flex-start;
+    align-items: center;
+    min-height: 45px;
+    border-bottom: 1px solid #e6e7e8;
```
adamklepacz: use variables

client/src/modules/list/components/Items/model/reducer.js
```diff
@@ -0,0 +1,58 @@
...
+        )
+      };
+    case ItemActionTypes.VOTE_SUCCESS:
+      return {
+        ...state,
+        items: state.items.map(item =>
+          item._id === action.payload.item._id
+            ? {
+                ...action.payload.item,
+                voterIds: action.payload.item.voterIds
```
adamklepacz: we need to check voting reducer cause I can see voterIds here. And we have removed IDS from front layer supposedly 😃

server/controllers/list.js
```diff
@@ -705,13 +705,62 @@ const updateItemDetails = (req, resp) => {
...
+
+      newItemId = item._id;
+      list.items.push(item);
+
+      return list.save();
+    })
+    .then(list =>
+      resp.status(200).send({
+        message: 'Item successfully cloned.',
+        item: responseWithItem(list.items.id(newItemId))
```
adamklepacz: I'm wondering do we need to send saved item, or we can just copy existing one from redux state and send new itemID only??

client/src/common/components/CardItem.scss
alemikolo: Yes, it's necessary. Without these rules, if you add a long list title, the ellipsis is added but lock icon is very small (both in Chrome and Firefox):
Firefox_Screenshot_2019-04-23T05-37-32 262Z

client/src/common/components/Toolbar/components/UserBar/UserBar.scss
alemikolo: Without this rule, the logout button has a different font like the whole app. It's strange because it should inherit font-family by default, but it doesn't.

client/src/modules/list/components/Items/ListItem/ListItem.jsx
alemikolo: If cloning the last item causes that the items count is above the limit of displayed items, this cloned item is being unmounted. Since there is still async action in progress we get warning in the browser. To prevent it I reject pending promises in componentWillUnmount. We want to do these actions only when the component is mounted so we have to check: if we created AbortPromiseException we know that component is unmounted and these actions shouldn't be invoked.

server/controllers/list.js
alemikolo: We need new itemId and 'created-at', isVoted and VotesCount. Of course, I could send only these two first fields and set isVoted and VotesCount myself in cloneItem action but I think that now is more uniformly...

client/src/modules/list/components/Items/model/reducer.js
alemikolo: You're right - it's unnecessary.  It looks like this is a souvenir from the past...

client/src/common/components/CardItem.scss
```diff
@@ -31,6 +31,8 @@
     svg {
       height: 12px;
       width: 12px;
+      // min-height: 12px;
+      // min-width: 12px;
```
marekrozmus: commentd out code added

client/src/common/components/Preloader.jsx
```diff
@@ -12,34 +12,59 @@ export const PreloaderSize = {
...
-    </div>
-    <span className="preloader__text">{message}</span>
-  </div>
-);
+class Preloader extends PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = { displayAnimation: false };
+    this.timer = setTimeout(this.enableAnimation, 250);
```
marekrozmus: shouldn't this be rather in component did mount?

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -33,14 +37,27 @@ class ListItem extends PureComponent {
...
 
   componentDidUpdate() {
     this.checkIfFieldsUpdated();
   }
 
+  componentWillUnmount() {
+    this.pendingPromises.map(promise => promise.abort());
+  }
+
+  appendPendingPromise = promise => {
```
marekrozmus: why not `add` but `append`?

client/src/modules/list/components/Items/model/actionTypes.js
```diff
@@ -1,13 +1,16 @@
 export const ItemActionTypes = Object.freeze({
-  UPDATE_DETAILS_FAILURE: 'item/UPDATE_DETAILS_FAILURE',
-  UPDATE_DETAILS_REQUEST: 'item/UPDATE_DETAILS_REQUEST',
-  UPDATE_DETAILS_SUCCESS: 'item/UPDATE_DETAILS_SUCCESS',
   ADD_FAILURE: 'item/ADD_FAILURE',
   ADD_REQUEST: 'item/ADD_REQUEST',
   ADD_SUCCESS: 'item/ADD_SUCCESS',
+  CLONE_FAILURE: 'item/CLONE__FAILURE',
+  CLONE_REQUEST: 'item/CLONE__REQUEST',
+  CLONE_SUCCESS: 'item/CLONE__SUCCESS',
```
marekrozmus: why double `_`?

client/src/modules/list/components/Items/model/actions.js
```diff
@@ -0,0 +1,190 @@
...
+  itemId,
+  listId,
+  updatedAuthor
+) => dispatch => {
+  dispatch(toggleItemRequest());
+  return patchData(`${ENDPOINT_URL}/lists/${listId}/update-item`, {
+    authorName: updatedAuthor,
+    itemId,
+    isOrdered: !isOrdered,
+    listId
```
marekrozmus: do you need to pass all that data to update `isOrdered`?
`itemId` and `isOrdered` should be enough - `listId` is already in api path

alemikolo: I removed "listId". I removed updatedAuthor from passing it to backend (it's not used there now) but I left it in action. This "toggling item" functionality will be refactored in the future task "EOC-151 Fix item toggling (restoring should change author)" because for now it only changes "isOrdered" field. We should fix it to store new User Id when an item is restored, so "updatedAuthor" variable will be needed then.

alemikolo:
3. I added preloader animation to cloning item functionality. I added some delay to it by the way. In my first solution, I placed this delay inside preloader component but it had some UX's disadvantages - short flashes of content: the content was displayed next was disappearing and next, the preloader was displaying). If the network was fast, the preloader wasn't rendered but the content still flashed. 
In the current approach, I set "pending" prop after some time. If async action performed faster I clear timeout. Why 1 second: (https://ux.stackexchange.com/a/104782)

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -33,14 +37,27 @@ class ListItem extends PureComponent {
...
 
   componentDidUpdate() {
     this.checkIfFieldsUpdated();
   }
 
+  componentWillUnmount() {
+    this.pendingPromises.map(promise => promise.abort());
+  }
+
+  addPendingPromise = promise => {
```
adamklepacz: unecessary braces

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -130,17 +147,45 @@ class ListItem extends PureComponent {
...
+    const {
+      cloneItem,
+      data: { _id: itemId },
+      match: {
+        params: { id: listId }
+      }
+    } = this.props;
+
+    this.setState({ pending: true });
+    const abortableCloning = makeAbortablePromise(cloneItem(listId, itemId));
```
adamklepacz: add line break before and after const

client/src/modules/list/components/Items/model/actions.js
```diff
@@ -0,0 +1,188 @@
...
+    });
+};
+
+export const cloneItem = (listId, itemId) => dispatch => {
+  dispatch(cloneItemRequest());
+  return patchData(`${ENDPOINT_URL}/lists/${listId}/clone-item`, {
+    itemId
+  })
+    .then(resp => resp.json())
+    .then(data => {
```
adamklepacz: follow the same pattern as we did in previous functions 
so `json`   -->
`dispatch(cloneItemSuccess(listId, data.item));`

server/controllers/list.js
```diff
@@ -705,13 +705,62 @@ const updateItemDetails = (req, resp) => {
...
+    $or: [{ ownerIds: userId }, { memberIds: userId }]
+  })
+    .exec()
+    .then(list => {
+      if (!list) {
+        throw new BadRequestException('List data not found.');
+      }
+
+      const { description, link, name } = list.items.id(itemId);
+
```
adamklepacz: remove break line

server/controllers/list.js
```diff
@@ -705,13 +705,62 @@ const updateItemDetails = (req, resp) => {
...
+    })
+    .then(list =>
+      resp.status(200).send({
+        message: 'Item successfully cloned.',
+        item: responseWithItem(list.items.id(newItemId))
+      })
+    )
+    .catch(err => {
+      if (err instanceof BadRequestException) {
+        const { status, message } = err;
```
adamklepacz: add line break

PR #120, opened 2019-04-26
[EOC-160, EOC-158, EOC-159, EOC-155] -  User roles for lists

adamklepacz:
Includes: 

- EOC-158 - Implementation of setting member role in lists in db
- EOC-159 - Implementation of setting viewer role in private lists in db
- EOC-155 - Implementation of setting owner role in lists in db
- EOC-160 - Implementation of setting viewer role in shared list in db

Flow: 
- Dashboard: 
 
     - Private list on dashboard - while adding new member, he becomes **viewer** by default -then we can change his role  to '**member**' or '**owner**'.
    - All added users are guest by default.

- Cohort Private list: 
    -  Person who creates, becomes an **owner** by default and also has **member** and **viewer** 
  rights.
    - List is not visible to cohort members.
    - Person added to private list is **viewer** by default, an can be changed to **member** and **owner**. 
    - User from outside cohort is a **guest** and has **viewer** role by default.

- Cohort public list:
    - Person that creates, becomes an **owner** by default, and also has **member** and **viewer** rights.
    - All cohort members becomes public list **viewers**, by deafult.
    -  When user removed from cohort, and has been viewer in public lists - is automaticlly removed frm that lists.
    - When user removed from cohort, but has been **member** or **owner** in public lists - he lefts in this lists with his roles, but now he has isGuest property.
     - User from outside cohort is a **guest** and has **viewer** role by default.

Zrzut ekranu 2019-04-26 o 11 05 01
Zrzut ekranu 2019-04-26 o 11 04 46
Zrzut ekranu 2019-04-26 o 11 04 52


server/controllers/cohort.js
```diff
@@ -147,24 +149,24 @@ const getCohortDetails = (req, resp) => {
...
-        memberIds: membersData
+        memberIds: membersCollection
       } = doc;
 
       if (isArchived) {
         return resp.status(200).json({ _id, isArchived, name });
       }
 
-      const owners = ownerIds.map(owner => owner.id);
+      const ownerIds = ownersCollection.map(owner => owner.id);
```
alemikolo: ownerIds could be an array with ids (no need populate ownerIds), we don't need owners objects and mapping them to ids

server/controllers/cohort.js
```diff
@@ -147,24 +149,24 @@ const getCohortDetails = (req, resp) => {
...
+      const isOwner = checkIfArrayContainsUserId(ownerIds, req.user._id);
       const members = responseWithCohortMembers(
-        [...membersData, ...ownerIds],
-        owners
+        [...membersCollection, ...ownersCollection],
```
alemikolo: this could be only members (owners are already there)

server/controllers/list.js
```diff
@@ -22,28 +21,53 @@ const {
...
+    Cohort.findOne({ _id: cohortId }).then(cohort => {
+      const { memberIds } = cohort;
+      list.viewersIds = memberIds;
 
-    resp
-      .status(201)
-      .location(`/lists/${doc._id}`)
-      .send(responseWithList(doc, userId));
-  });
+      return list.save((err, doc) => {
```
alemikolo: Don't mix using callbacks and promises with queries, or you may end up with duplicate operations.
(https://mongoosejs.com/docs/queries.html)

server/controllers/list.js
```diff
@@ -22,28 +21,53 @@ const {
...
   });
 
-  list.save((err, doc) => {
-    if (err) {
-      return resp
-        .status(400)
-        .send({ message: 'List not saved. Please try again.' });
-    }
+  if (cohortId && !isListPrivate) {
+    Cohort.findOne({ _id: cohortId }).then(cohort => {
```
alemikolo: where is catch? - what if something goes wrong?

server/controllers/list.js
```diff
@@ -191,10 +215,11 @@ const getListData = (req, resp) => {
 
   List.findOne({
     _id: listId,
-    $or: [{ ownerIds: userId }, { memberIds: userId }]
+    $or: [{ ownerIds: userId }, { memberIds: userId }, { viewersIds: userId }]
```
alemikolo: we can check only viewers (owners and members are already there)

server/controllers/list.js
```diff
@@ -216,10 +240,12 @@ const getListData = (req, resp) => {
               );
             }
 
-            const { memberIds, ownerIds } = cohort;
-            return [...memberIds, ...ownerIds];
+            const { memberIds } = cohort;
+            return memberIds;
```
alemikolo: what if something went wrong here do we need catch after this then?

server/controllers/list.js
```diff
@@ -191,10 +215,11 @@ const getListData = (req, resp) => {
 
   List.findOne({
     _id: listId,
-    $or: [{ ownerIds: userId }, { memberIds: userId }]
+    $or: [{ ownerIds: userId }, { memberIds: userId }, { viewersIds: userId }]
   })
     .populate('memberIds', 'avatarUrl displayName _id')
     .populate('ownerIds', 'avatarUrl displayName _id')
+    .populate('viewersIds', 'avatarUrl displayName _id')
```
alemikolo: if we populate viewerIds there is no need to populate meberIds and ownerIds these arrays will be used to check roles only, all necessary data will be inluded in populated viewersIds collection, and there will be no need to map over these collections and get ids from them.

server/controllers/list.js
```diff
@@ -239,27 +266,29 @@ const getListData = (req, resp) => {
           .json({ cohortId, _id, isArchived, isPrivate, name });
       }
 
-      const allMembers = isPrivate
-        ? [...memberIds, ...ownerIds]
-        : uniqueMembers(cohortMembers, [...memberIds, ...ownerIds]);
+      const ownerIds = ownersCollection.map(owner => owner.id);
```
alemikolo: See above:  this could be array of ids not whole users objects

server/controllers/list.js
```diff
@@ -569,29 +593,76 @@ const changeToOwner = (req, resp) => {
...
+  const { userId } = req.body;
+  const {
+    user: { _id: ownerId }
+  } = req;
+
+  List.findOne({
+    _id: listId,
+    ownerIds: { $in: [ownerId] }
+  })
+    .populate('cohortId', 'memberIds ownerIds viewersIds')
```
alemikolo: viewerIds in cohort?

server/controllers/list.js
```diff
@@ -569,29 +593,76 @@ const changeToOwner = (req, resp) => {
...
+    .exec()
     .then(doc => {
       if (!doc) {
         throw new BadRequestException("Can't set user as a list's member.");
       }
 
-      const { cohortId: cohort, isPrivate, ownerIds } = doc;
-      const isNotCohortMember = !checkIfCohortMember(cohort, userId);
+      const { ownerIds, memberIds } = doc;
+      const userIsNotAMember = !memberIds.some(id => id.equals(userId));
```
alemikolo: You wrote separate helper for that checking. Couldn't be used here?

server/controllers/list.js
```diff
@@ -569,29 +593,76 @@ const changeToOwner = (req, resp) => {
...
 
-      if (isPrivate || isNotCohortMember) {
-        doc.memberIds.push(userId);
+      if (userIsNotAMember) {
+        memberIds.push(userId);
+      }
+
+      const userIsOwner = ownerIds.some(id => id.equals(userId));
```
alemikolo: You wrote separate helper for that checking. Couldn't be used here? There more place where it could be used...

server/controllers/list.js
```diff
@@ -605,70 +676,106 @@ const changeToMember = (req, resp) => {
...
+  const { userId } = req.body;
+  const {
+    user: { _id: ownerId }
+  } = req;
+
+  List.findOne({
+    _id: listId,
+    ownerIds: { $in: [ownerId] }
+  })
+    .populate('cohortId', 'memberIds ownerIds viewersIds')
```
alemikolo: viewers in cohort?

client/src/modules/list/model/reducer.js
```diff
@@ -62,7 +62,7 @@ const items = (state, action) => {
 
 const membersReducer = (state, action) => {
```
marekrozmus: no initial state?

server/common/utils/index.js
```diff
@@ -107,8 +101,19 @@ const responseWithCohort = (cohort, userId) => {
   };
 };
 
-const checkIfOwner = (ownerIds, userId) =>
-  ownerIds.indexOf(userId.toString()) > -1;
+const checkIfArrayContainsUserId = (idsArray, userId) => {
```
marekrozmus: do we need this method - can't we just use `includes` from `lodash`? Do we need to convert everything to strings?

server/common/utils/index.js
```diff
@@ -132,74 +137,59 @@ const checkIfCohortMember = (cohort, userId) => {
...
 
   return {
     _id: newMemberId,
     avatarUrl,
     displayName,
-    isGuest: checkIfGuest(cohort, newMemberId),
-    isOwner: checkIfOwner(ownerIds, newMemberId)
+    isGuest: checkIfGuest(cohortMembers, newMemberId),
+    isMember: false,
+    isOwner: false
```
marekrozmus: maybe create a helper function that will return some object like:
```
{
  isGuest: true,
  isMember: false,
  isOwner: true,
  isViewer: true
}
```

server/controllers/cohort.js
adamklepacz: yes, but I need as well the whole collection of users to respondWithCohortMembers with data like display_name for example, and I also need only ids.

So populating ownerIds array above it's necessary here.

server/controllers/cohort.js
adamklepacz: I need the whole collections of data here to respond with members with data like display_name etc...

server/controllers/cohort.js
alemikolo: but you populate memberIds - where you have all users data (members and owners), there is no need to duplicate users data on ownerIds, ownerIds is needed to check users roles...

server/controllers/cohort.js
alemikolo: but memberIds contains all ownerIds, if you populate memberIds there is no need to populate the same data for ownerIds. OwnerIds is needed to check roles only

server/controllers/list.js
```diff
@@ -569,29 +593,76 @@ const changeToOwner = (req, resp) => {
...
   const { id: listId } = req.params;
   const { userId } = req.body;
+
   const {
     user: { _id: ownerId }
   } = req;
 
-  List.findOne({ _id: listId, ownerIds: { $all: [ownerId, userId] } })
+  List.findOne({ _id: listId, ownerIds: ownerId })
     .populate('cohortId', 'memberIds ownerIds')
```
alemikolo: is this populate necessary here? I can't see an usage of cohortId? The same in addOwnerRoleController

server/controllers/list.js
```diff
@@ -605,70 +676,106 @@ const changeToMember = (req, resp) => {
...
-          .then(doc => {
-            const { _id, avatarUrl, displayName } = doc;
-            return { _id, avatarUrl, cohort, displayName, ownerIds };
-          })
-          .catch(() => {
-            throw new BadRequestException('User data not found.');
-          });
+      return User.findOne({ email }).exec();
+    })
+    .then(userData => {
```
alemikolo: Your code is working good, but consider to add this check in this then block:
```
      if (!userData) {
        throw new BadRequestException(`There is no user of email: ${email}`);
      }
```
I think we should return information that there no user was found, no that some error occurred. This if statement will be a good place to send an invitation email in some future improvements.

client/src/common/components/Members/components/MemberDetails.jsx
```diff
@@ -302,16 +352,20 @@ MemberDetails.propTypes = {
   displayName: PropTypes.string.isRequired,
   isCurrentUserAnOwner: PropTypes.bool.isRequired,
   isGuest: PropTypes.bool,
```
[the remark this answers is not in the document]
adamklepacz: of course, it is

client/src/common/components/Switcher.jsx
```diff
@@ -0,0 +1,30 @@
...
+  <div className="ss-switch">
+    <label className="ss-switch__label" htmlFor={htmlFor}>
+      {label}
+      <input
+        checked={checked}
+        className="ss-switch__input"
+        id={htmlFor}
+        onChange={onChange}
+        type="checkbox"
+        value={value}
```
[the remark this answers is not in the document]
adamklepacz: it's need here in MemberDetails.jsx 
```
  handleChangingRoles = event => {
    event.stopPropagation();
    const {
      target: { value: selectedRole }
    } = event;
    const { route } = this.props;

    switch (route) {
      case Routes.COHORT:
        this.changeCohortRole(selectedRole);
        break;
      case Routes.LIST:
        this.changeListRole(selectedRole);
        break;
      default:
        break;
    }
  };
```
where the value of checkbox sets the selected role

server/common/utils/index.js
adamklepacz: I've checked this solution with includes, earlier and it doesn't work cause it's doesn't recognize ObjectId format and cannot compare to userId 

Also checking on stackoveflow, some other guy had the same problem, and one of the answers was "convert objectId to string"

client/src/modules/list/index.jsx
```diff
@@ -101,14 +101,15 @@ class List extends Component {
       return null;
     }
 
-    const { cohortId, isArchived, isPrivate, name } = list;
+    const { cohortId, isArchived, isPrivate, name, isGuest } = list;
     const orderedItems = items ? items.filter(item => item.isOrdered) : [];
     const listItems = items ? items.filter(item => !item.isOrdered) : [];
+    const isCohortList = !!cohortId;
```
marekrozmus: do not use `!!` operator - use simple comparison.

server/controllers/list.js
```diff
@@ -191,10 +213,11 @@ const getListData = (req, resp) => {
 
   List.findOne({
     _id: listId,
-    $or: [{ ownerIds: userId }, { memberIds: userId }]
+    $or: [{ viewersIds: userId }]
   })
-    .populate('memberIds', 'avatarUrl displayName _id')
-    .populate('ownerIds', 'avatarUrl displayName _id')
+    // .populate('memberIds', 'avatarUrl displayName _id')
+    // .populate('ownerIds', 'avatarUrl displayName _id')
```
marekrozmus: commented out code

server/controllers/list.js
```diff
@@ -205,22 +228,17 @@ const getListData = (req, resp) => {
...
-              );
-            }
-
-            const { memberIds, ownerIds } = cohort;
-            return [...memberIds, ...ownerIds];
-          });
+        return Cohort.findOne({ _id: cohortId }).exec();
       }
     })
+    .then(cohort => {
```
marekrozmus: could it be just: `cohort ? cohort.memberIds : []`?

server/controllers/list.js
```diff
@@ -239,27 +258,30 @@ const getListData = (req, resp) => {
           .json({ cohortId, _id, isArchived, isPrivate, name });
       }
 
-      const allMembers = isPrivate
-        ? [...memberIds, ...ownerIds]
-        : uniqueMembers(cohortMembers, [...memberIds, ...ownerIds]);
+      /** TODO: przenieść mapowanie do wnętrza funkcji */
```
marekrozmus: do not use polish language - even in comments

server/controllers/list.js
```diff
@@ -605,70 +668,114 @@ const changeToMember = (req, resp) => {
...
-          throw new BadRequestException('List data not found.');
-        });
+    .then(() => {
+      let userToSend;
+
+      if (cohortMembers) {
+        userToSend = responseWithListMember(user, cohortMembers);
+      } else {
+        userToSend = responseWithListMember(user, []);
+      }
```
marekrozmus: could be just:
`const userToSend = responseWithListMember(user, cohortMembers ? cohortMembers : []);`

adamklepacz: can't use this because of es lint rule 
Unnecessary use of conditional expression for default assignment

but I have found another solution

server/common/utils/index.js
```diff
@@ -85,7 +79,7 @@ const responseWithItem = (item, userId) => {
 const responseWithCohorts = (cohorts, userId) =>
   _map(cohorts, ({ _doc }) => {
     const { favIds, memberIds, ownerIds, ...rest } = _doc;
-    const membersCount = [...memberIds, ...ownerIds].length;
+    const membersCount = [...memberIds].length;
```
alemikolo: Is destructuring needed here? Variable "memberIds" is an array, isn't it?

server/common/utils/index.js
```diff
@@ -95,8 +89,8 @@ const responseWithCohorts = (cohorts, userId) =>
   });
 
 const responseWithCohort = (cohort, userId) => {
-  const { _id, description, favIds, memberIds, name, ownerIds } = cohort;
-  const membersCount = [...memberIds, ...ownerIds].length;
+  const { _id, description, favIds, memberIds, name } = cohort;
+  const membersCount = [...memberIds].length;
```
alemikolo: Is destructuring needed here?

server/controllers/cohort.js
```diff
@@ -21,7 +21,8 @@ const createCohort = (req, resp) => {
   const cohort = new Cohort({
     name,
     description,
-    ownerIds: userId
+    ownerIds: userId,
+    memberIds: userId
```
alemikolo: alphabetic order

server/controllers/cohort.js
```diff
@@ -152,19 +153,18 @@ const getCohortDetails = (req, resp) => {
...
       if (isArchived) {
         return resp.status(200).json({ _id, isArchived, name });
       }
 
-      const owners = ownerIds.map(owner => owner.id);
+      const isOwner = checkIfArrayContainsUserId(ownerIds, req.user._id);
       const members = responseWithCohortMembers(
-        [...membersData, ...ownerIds],
-        owners
+        [...membersCollection],
```
alemikolo: There is no need for destructuring here

server/controllers/cohort.js
```diff
@@ -286,52 +288,79 @@ const removeOwner = (req, resp) => {
...
 
-      resp.status(404).send({ message: 'Cohort data not found.' });
-    }
-  );
+      return List.updateMany(
+        {
+          cohortId,
+          isPrivate: false,
+          viewersIds: { $in: [userId] },
+          memberIds: { $nin: [userId] },
```
alemikolo: alphabetic order

server/controllers/list.js
```diff
@@ -22,28 +21,51 @@ const {
...
-    if (err) {
-      return resp
-        .status(400)
-        .send({ message: 'List not saved. Please try again.' });
-    }
+  if (cohortId && !isListPrivate) {
+    Cohort.findOne({ _id: cohortId })
+      .then(cohort => {
+        const { memberIds } = cohort;
+        list.viewersIds = memberIds;
```
alemikolo: what if user isn't a cohort's member? Should we check here something like this:
```
if(memberIds.includes(userId)){
throw new BadRequestException();
}
```

server/controllers/list.js
```diff
@@ -191,10 +213,9 @@ const getListData = (req, resp) => {
 
   List.findOne({
     _id: listId,
-    $or: [{ ownerIds: userId }, { memberIds: userId }]
+    $or: [{ viewersIds: userId }]
```
alemikolo: $or is no need here because we have only one array to check

server/controllers/list.js
```diff
@@ -605,70 +654,108 @@ const changeToMember = (req, resp) => {
...
   const { email } = req.body;
+  let list;
+  let user;
+  let cohortMembers = [];
 
-  List.findOne({ _id: listId, $or: [{ ownerIds: currentUserId }] })
-    .populate('cohortId', 'memberIds ownerIds')
+  List.findOne({
+    _id: listId,
+    $or: [{ ownerIds: currentUserId }]
```
alemikolo: $or is unnecessary here

server/controllers/list.js
adamklepacz: when creating a list, he always gonna be viewer member and owner by default.
When creating a list within cohort he is already a cohort member. 
Otherwise, how he can create list within cohort??

alemikolo: Hmm, probably you're right but if he will have the session sid (cookie) and cohortId he could do this via postman

server/controllers/list.js
```diff
@@ -40,9 +40,16 @@ const createList = (req, resp) => {
...
-        list.viewersIds = memberIds;
 
-        return list.save();
+        if (checkIfArrayContainsUserId(memberIds, userId)) {
+          list.viewersIds = memberIds;
+
+          return list.save();
+        }
+
+        throw new BadRequestException(
```
alemikolo: You don't catch this specific error message, you prevent this edge case, the code is stopped here, but in catch statement, you should write something like this:
```
.catch(err => {
      if (err instanceof BadRequestException) {
        const { status, message } = err;

        return resp.status(status).send({ message });
      }

      resp.status(400).send({
        message: 'List not saved. Please try again.'
      });
    });
```

PR #121, opened 2019-04-29
[EOC-161, EOC-162] Implementation of setting roles in cohorts in db

alemikolo:
Setting member role during creating cohort and adding new member was already done. I made only small adjustments (to setting member role) and decided to include task EOC-162 to the current pull request.
This branch started from Adam's "EOC-152" so it includes all unmerged changes from EOC-152, but I compared it to EOC-152 not to master, so there are "fewer" files to check.

screencapture-localhost-3000-cohort-5cc69f7386377e054f9d1d42-2019-04-29-13_26_21


client/src/common/components/Members/components/MemberDetails.jsx
```diff
@@ -174,8 +178,9 @@ class MemberDetails extends PureComponent {
           </button>
           <SwitchButton
             checked={checked}
+            disabled={disabled}
             label={label}
-            onChange={this.handleChangingRoles}
+            onChange={disabled ? null : this.handleChangingRoles}
```
adamklepacz: try `!disabled && this.handleChangingRoles`

client/src/common/components/Members/components/MemberDetails.jsx
```diff
@@ -369,13 +374,14 @@ MemberDetails.propTypes = {
...
+  addOwnerRoleInCohort: PropTypes.func.isRequired,
   addOwnerRoleInList: PropTypes.func.isRequired,
-  changeRoleInCohort: PropTypes.func.isRequired,
   onClose: PropTypes.func.isRequired,
   removeCohortMember: PropTypes.func.isRequired,
   removeListMember: PropTypes.func.isRequired,
   removeMemberRoleInList: PropTypes.func.isRequired,
-  removeOwnerRoleInList: PropTypes.func.isRequired
+  removeOwnerRoleInList: PropTypes.func.isRequired,
+  removeOwnerRoleInCohort: PropTypes.func.isRequired
```
adamklepacz: sort props

client/src/modules/cohort/model/reducer.js
```diff
@@ -18,15 +18,29 @@ const membersReducer = (state, action) => {
       } = action;
       return state.filter(member => member._id !== userId);
     }
-    case CohortActionTypes.CHANGE_ROLE_SUCCESS: {
+    case CohortActionTypes.ADD_OWNER_ROLE_SUCCESS: {
       const {
-        payload: { isOwner, userId }
+        payload: { userId }
+      } = action;
```
adamklepacz: add line break

server/controllers/cohort.js
```diff
@@ -351,27 +317,37 @@ const removeMember = (req, resp) => {
         { $pull: { viewersIds: userId } }
       ).exec();
     })
-    .then(() =>
+    .then(() => {
```
adamklepacz: uneccessary braces

client/src/common/components/Members/components/MemberDetails.jsx
alemikolo: I really tried, I started from your solution, there is a react error that onChange should receive function or undefined/null if this adding of handler depends on other props...

client/src/common/components/Members/components/MemberDetails.jsx
```diff
@@ -86,20 +87,22 @@ class MemberDetails extends PureComponent {
...
+      removeOwnerRoleInCohort
     } = this.props;
 
     this.setState({ pending: true });
 
-    changeRoleInCohort(id, userId, selectedRole).finally(() =>
-      this.setState({ pending: false })
-    );
+    const action = isOwner ? removeOwnerRoleInCohort : addOwnerRoleInCohort;
+
```
adamklepacz: I think we sholud check in those methods whetever current user isOwner and then allow to fire the action
something like this 
```
If(isOwner) { 
//...rest of the code

this.setState(...)
action().....
}
```

here in this method, and down below in `changeListRole` method also.

server/controllers/cohort.js
```diff
@@ -353,22 +322,32 @@ const removeMember = (req, resp) => {
...
+      });
+    });
 };
 
-const changeToOwner = (req, resp) => {
+const addOwnerRole = (req, resp) => {
   const { id: cohortId } = req.params;
   const { userId } = req.body;
   const {
     user: { _id: ownerId }
```
adamklepacz: this should be named, currentUserId

server/controllers/cohort.js
```diff
@@ -353,22 +322,32 @@ const removeMember = (req, resp) => {
...
   } = req;
 
   Cohort.findOneAndUpdate(
-    { _id: cohortId, ownerIds: ownerId, memberIds: userId },
+    {
+      _id: cohortId,
+      ownerIds: { $in: [ownerId], $nin: [userId] },
```
adamklepacz: because here we've got a confusing query.
We're checking if the owner is in the ownerIds array. Logically the owner is always in ownerIds array.

We should check here if currentUser is an owner, and the new user, coming as userId is not an owner - what makes sense while reading.

server/controllers/cohort.js
```diff
@@ -401,31 +380,31 @@ const changeToMember = (req, resp) => {
     (err, doc) => {
```
adamklepacz: same about here, please update ownerId from request to currentUserid

server/controllers/cohort.js
```diff
@@ -401,31 +380,31 @@ const changeToMember = (req, resp) => {
...
 
       if (doc) {
         return resp.status(200).send({
-          message: "User has been successfully set as a cohort's member."
+          message: 'User has no owner role.'
         });
       }
 
-      resp.status(404).send({ message: 'Cohort data not found.' });
+      resp.status(400).send({ message: 'Cohort data not found.' });
```
adamklepacz: please convert those statements to match our overall pattern where we check: 
1. if(err)
2. if( !doc )
3. return status(200)

client/src/common/components/Members/components/MemberDetails.jsx
alemikolo: In render method we check if currentUser is an Owner and then disable render options that allow user to change roles and remove other members if he isn't an owner.
In these methods we would use the same props for this checking as we used in render method so I think It's not necessary.

PR #122, opened 2019-04-30
[EOC-166, EOC-167] Implementation of roles on frontend for COHORT

alemikolo:
Only cohort owner can load and see archived cohort.

Member view:
Firefox_Screenshot_2019-04-30T10-46-32 451Z

Owner view:
download (1)

### DIFF
--- a/client/src/modules/cohort/index.jsx
+++ b/client/src/modules/cohort/index.jsx
@@ -219,7 +219,7 @@ class Cohort extends PureComponent {
                     items={lists}
                     name="Lists"
                     onAddNew={this.handleDialogContext(DialogContext.CREATE)}
-                    pending={false}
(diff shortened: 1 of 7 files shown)


client/src/modules/cohort/index.jsx
```diff
@@ -213,7 +219,7 @@ class Cohort extends PureComponent {
                     items={lists}
                     name="Lists"
                     onAddNew={this.handleDialogContext(DialogContext.CREATE)}
-                    pending={pendingForDetails}
+                    pending={false}
```
marekrozmus: this is required?

alemikolo: to mój błąd, już naprawiłem

PR #123, opened 2019-04-30
[EOC-163, EOC-160,  EOC-164, EOC-165] Implementation of viewer role in front end for list

adamklepacz:
Viewer on the list: 
- [x] can not edit the title of the list,
- [x] can not edit the description of the list,
- [x] can not add item
- [x] can not vote for items,
- [x] can not mark as completed,
- [x] can not change the role of members,
- [x] can not add a link or description,
- [x] can not clone an item
- [x] can not comment (UI only),
- [x] can not archive the list
- [x] can expand the item,

Member on the list:
- [x] can not edit the title of the list,
- [x] can not edit the list description
- [x] can not archive the list,
- [x] can not change roles for members,
- [x] can add viewers,
- [x] can vote for items,
- [x] can be marked as purchased
- [x] can add description and link to the item,
- [x] can add comments (UI Only)
- [x] can clone item.

Owner on the list: 
- [x] can edit the description of the list,
- [x] can edit the list description,
- [x] can archive the list,
- [x] can change roles to other members,
- [x] has permissions from MEMBER and VIEWER

### DIFF
--- a/client/src/common/components/Forms/TextInput.jsx
+++ b/client/src/common/components/Forms/TextInput.jsx
@@ -40,7 +40,7 @@ class TextInput extends PureComponent {
   };
 
   render() {
-    const { placeholder } = this.props;
+    const { disabled, placeholder } = this.props;
     const { isEnlarged, value } = this.state;
     return (
       <div className="ss-text-input">
@@ -58,6 +58,7 @@ class TextInput extends PureComponent {
         )}
         <input
           className="ss-text-input__input"
+          disabled={!disabled}
           name={placeholder}
           onBlur={this.handleBlur}
           onChange={this.handleOnChange}
@@ -73,6 +74,7 @@ class TextInput extends PureComponent {
 }
 
 TextInput.propTypes = {
+  disabled: PropTypes.bool,
   initialValue: PropTypes.string,
   placeholder: PropTypes.string.isRequired,
 
--- a/client/src/common/components/Forms/TextInput.jsx
+++ b/client/src/common/components/Forms/TextInput.jsx
@@ -58,7 +58,7 @@ class TextInput extends PureComponent {
         )}
         <input
           className="ss-text-input__input"
-          disabled={!disabled}
+          disabled={disabled}
           name={placeholder}
           onBlur={this.handleBlur}
           onChange={this.handleOnChange}
--- a/client/src/common/components/Forms/Textarea.jsx
+++ b/client/src/common/components/Forms/Textarea.jsx
@@ -40,7 +40,7 @@ class Textarea extends PureComponent {
   };
 
   render() {
-    const { placeholder } = this.props;
+    const { disabled, placeholder } = this.props;
     const { isEnlarged, value } = this.state;
     return (
       <div className="ss-textarea">
@@ -58,6 +58,7 @@ class Textarea extends PureComponent {
         )}
         <textarea
           className="ss-textarea__textarea"
+          disabled={!disabled}
           name={placeholder}
           onBlur={this.handleBlur}
           onChange={this.handleOnChange}
@@ -72,6 +73,7 @@ class Textarea extends PureComponent {
 }
 
 Textarea.propTypes = {
+  disabled: PropTypes.bool,
   initialValue: PropTypes.string,
   placeholder: PropTypes.string.isRequired,
 
--- a/client/src/common/components/Forms/Textarea.jsx
+++ b/client/src/common/components/Forms/Textarea.jsx
@@ -58,7 +58,7 @@ class Textarea extends PureComponent {
         )}
         <textarea
           className="ss-textarea__textarea"
-          disabled={!disabled}
+          disabled={disabled}
           name={placeholder}
           onBlur={this.handleBlur}
           onChange={this.handleOnChange}
--- a/client/src/modules/list/components/Items/ListItem/ListItem.jsx
+++ b/client/src/modules/list/components/Items/ListItem/ListItem.jsx
@@ -15,7 +15,11 @@ import Comment from 'common/components/Comments/Comment';
 import { RouterMatchPropType } from 'common/constants/propTypes';
 import { cloneItem, updateItemDetails } from '../model/actions';
 import SaveButton from 'common/components/SaveButton';
-import { isUrlValid, makeAbortablePromise } from 'common/utils/helpers';
+import {
+  handleEventPropagation,
+  isUrlValid,
+  makeAbortablePromise
+} from 'common/utils/helpers';
 import ErrorMessage from 'common/components/Forms/ErrorMessage';
 import Preloader, { PreloaderSize } from 'common/components/Preloader';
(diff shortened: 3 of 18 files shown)


client/src/common/components/Forms/TextInput.jsx
```diff
@@ -58,6 +58,7 @@ class TextInput extends PureComponent {
         )}
         <input
           className="ss-text-input__input"
+          disabled={!disabled}
```
marekrozmus: ?

client/src/common/components/Forms/Textarea.jsx
```diff
@@ -58,6 +58,7 @@ class Textarea extends PureComponent {
         )}
         <textarea
           className="ss-textarea__textarea"
+          disabled={!disabled}
```
marekrozmus: ?

client/src/common/components/Forms/TextInput.jsx
adamklepacz: Zrzut ekranu 2019-05-01 o 12 34 44

DIsabled field is controlled by isMember prop.
When user isMember so the prop is set as TRUE - I need to negate this value here to make thisi disabled=false, to keep this field active for him.

When user is not a member and the prop has value 'false' I need to negate this value to true, to keep this field disabled= true.  Cause not member is not able to edit thosee fields.
Same answer for the second comment

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -308,18 +326,20 @@ class ListItem extends PureComponent {
...
               voteForItem={voteForItem}
               votesCount={votesCount}
             />
           )}
-          <button
-            className="list-item__icon z-index-high"
-            onClick={this.handleItemToggling(authorId, _id, isOrdered)}
-            type="button"
-          />
+          {isMember && (
```
alemikolo: consider to leave this button visible, but disallow toggling. Otherwise This green "done" indicator won't be visible for viewers. This button isn't necessary but it helps to distinguish list's items.

server/controllers/cohort.js
```diff
@@ -159,17 +159,22 @@ const getCohortDetails = (req, resp) => {
       if (isArchived) {
         return resp.status(200).json({ _id, isArchived, name });
       }
-
+      const arrayOfMemberIds = membersCollection.map(member => member._id);
       const isOwner = checkIfArrayContainsUserId(ownerIds, req.user._id);
+      const isMember = checkIfArrayContainsUserId(
```
alemikolo: this check in unnecessary, if user is in cohort (the cohort was find via query), them is a member for sure. So isMember should be set to true for each cohort member(user).

server/controllers/list.js
```diff
@@ -188,23 +188,28 @@ const addItemToList = (req, resp) => {
...
         });
       }
 
-      doc
-        ? resp.status(200).send(responseWithItem(newItem, userId))
-        : resp.status(404).send({ message: 'List  not found.' });
+      if (!doc) {
+        return resp
+          .status(400)
+          .send({ message: "You don't have permissions to add new item" });
```
alemikolo: proposal: "Only member can add an item."

server/controllers/list.js
```diff
@@ -309,7 +316,7 @@ const updateListItem = (req, resp) => {
     {
       _id: listId,
       'items._id': itemId,
-      $or: [{ ownerIds: req.user._id }, { memberIds: req.user._id }]
+      $or: [{ ownerIds: userId }, { memberIds: userId }]
```
alemikolo: `memberIds: userId` will be enough, each owner is already a member.

client/src/modules/list/index.jsx
```diff
@@ -129,11 +140,15 @@ class List extends Component {
               ) : (
                 <Fragment>
                   <div className="list__items">
-                    <ItemsContainer items={listItems}>
-                      <InputBar />
+                    <ItemsContainer isMember={isMember} items={listItems}>
+                      {isMember && <InputBar isMember={isMember} />}
```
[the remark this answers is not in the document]
adamklepacz: Yes you right. I wanted to make it two-level secure 😄

client/src/common/components/Forms/Textarea.jsx
```diff
@@ -58,6 +58,7 @@ class Textarea extends PureComponent {
...
           name={placeholder}
           onBlur={this.handleBlur}
           onChange={this.handleOnChange}
```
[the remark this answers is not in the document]
adamklepacz: I agree with onChange....but there is no need to disable onBlur, while it's only causing style changes.

server/controllers/list.js
adamklepacz: this the case, where we been talking about on monday.
Without this $or, toggling doesnt work. I think I can fix in comming PR, where I'm changing all callbacks to promises.

server/controllers/cohort.js
```diff
@@ -159,17 +159,22 @@ const getCohortDetails = (req, resp) => {
       if (isArchived) {
         return resp.status(200).json({ _id, isArchived, name });
       }
-
+      const arrayOfMemberIds = membersCollection.map(member => member._id);
```
alemikolo: Is this check really necessary? In line 139 we look for a cohort where "memberIds" contains current user id (we should remove check for the "ownerIds" in the future).  If the cohort was found, we know that the user is a member for sure and we don't have to check this twice (extract members' ids from membersColleciton and iterate over the "arrayOfMemberIds". If you need to set isMember prop before sending cohort details to the user (because this prop is necessary in frontend), you can set it directly to true. Is there any additional reason to leave this check?

client/src/modules/list/components/Items/VotingBox/VotingBox.jsx
```diff
@@ -1,34 +1,48 @@
...
+  render() {
+    const { isMember, isVoted, votesCount } = this.props;
+
+    return (
+      <button
+        className={classNames('voting-box', {
+          'voting-box__voted': isVoted,
+          'voting-box__disabled': !isMember
+        })}
+        onClick={this.handleOnClick}
```
alemikolo: disable invoking handleOnClick and you wouldn't have to check isMember inside that handler

adamklepacz:
> Proposal: I think it would be good to make some helper for `this.handleEventPropagation`, because this is the second place we use the same functionality...

Added 👍

client/src/modules/list/components/Items/VotingBox/VotingBox.jsx
```diff
@@ -1,34 +1,47 @@
...
+  }
+}
 
 VotingBox.propTypes = {
+  isMember: PropTypes.bool.isRequired,
   isVoted: PropTypes.bool.isRequired,
   votesCount: PropTypes.number.isRequired,
 
-  voteForItem: PropTypes.func.isRequired
+  voteForItem: PropTypes.func
```
marekrozmus: this needs to be required or check if exist in line 13

PR #124, opened 2019-05-03
[EOC-154] security escaping data provided by user in the forms and db

adamklepacz:
Regarding XSS injection, React is safe by default as long as we don't use dangerouslySetInnerHTML prop anywhere. 
Everything that user can put inside forms, inputs, etc is always treated as a string and will be never executed.

Regarding to Express.js and MongoDb data sanitizing, we should sanitize data that we are using for creating mongo queries, but only data that comes from req.body. This is what the user can put in the inputs inside the app. For example, he can put $gt: { '' } query inside the input, and this kind of string can be executed on the backend as originally MongoDB query. 

There is a simple way, to prevent this kind of injection. All we have to do is to detect if passed data is an object, and if it contains $ sign, and remove that $ sign. 
There is a simple 1KB library, called mongo-sanitize which is very helpful to perform this task.

I have sanitized all the data, comes from req.body on the backend, both in list and cohort controllers.


server/controllers/cohort.js
```diff
@@ -48,12 +49,14 @@ const getCohortsMetaData = (req, resp) => {
...
     .exec()
-    .then(docs =>
-      docs
-        ? resp.status(200).send(responseWithCohorts(docs, currentUserId))
-        : resp.status(404).send({ message: 'No cohorts data found.' })
-    )
-    .catch(err =>
+    .then(docs => {
+      if (!docs) {
+        return resp.status(400).send({ message: 'No cohorts data found.' });
```
alemikolo: You should throw an error here to make this promise rejected (if you'll use our own error types, remember to catch it catch statement). The return doesn't do this and code will try to send a second response which will cause an error/warning.

server/controllers/cohort.js
```diff
@@ -65,37 +68,38 @@ const getArchivedCohortsMetaData = (req, resp) => {
...
-        });
-      }
-
+    { sort: { created_at: -1 } }
+  )
+    .then(docs => {
       if (!docs) {
         return resp
-          .status(404)
+          .status(400)
```
alemikolo: You should throw an error here - see above

server/controllers/cohort.js
```diff
@@ -65,37 +68,38 @@ const getArchivedCohortsMetaData = (req, resp) => {
...
-    (err, docs) => {
-      if (err) {
-        return resp.status(400).send({
-          message:
-            'An error occurred while fetching the archived cohorts data. Please try again.'
-        });
-      }
-
+    { sort: { created_at: -1 } }
+  )
```
alemikolo: .exec()

server/controllers/cohort.js
```diff
@@ -104,39 +108,45 @@ const updateCohortById = (req, resp) => {
...
-        return resp.status(400).send({
-          message:
-            'An error occurred while updating the cohort. Please try again.'
-        });
+    dataToUpdate
+  )
+    .exec()
+    .then(doc => {
+      if (!doc) {
+        return resp.status(400).send({ message: 'Cohort  not found.' });
```
alemikolo: Throw an error here - see above

server/controllers/cohort.js
```diff
@@ -48,12 +49,14 @@ const getCohortsMetaData = (req, resp) => {
     .select('_id name description favIds memberIds ownerIds')
     .sort({ createdAt: -1 })
     .exec()
-    .then(docs =>
-      docs
-        ? resp.status(200).send(responseWithCohorts(docs, currentUserId))
-        : resp.status(404).send({ message: 'No cohorts data found.' })
-    )
-    .catch(err =>
+    .then(docs => {
```
marekrozmus: why this change?

server/controllers/cohort.js
marekrozmus: extra space after `Cohort`

server/controllers/list.js
```diff
@@ -187,26 +197,27 @@ const addItemToList = (req, resp) => {
...
-      if (err) {
-        return resp.status(400).send({
-          message:
-            'An error occurred while adding a new item. Please try again.'
-        });
+    { new: true }
+  )
+    .exec()
+    .then(doc => {
+      if (!doc) {
```
alemikolo: throw an error...

server/controllers/list.js
```diff
@@ -154,20 +160,24 @@ const getArchivedListsMetaData = (req, resp) => {
...
-        return resp.status(400).send({
-          message:
-            'An error occurred while fetching the archived lists data. Please try again.'
-        });
+    { sort: { created_at: -1 } }
+  )
+    .exec()
+    .then(docs => {
+      if (!docs) {
+        return resp
```
alemikolo: throw an error

server/controllers/list.js
```diff
@@ -115,23 +121,23 @@ const getListsMetaData = (req, resp) => {
...
-          message:
-            'An error occurred while fetching the lists data. Please try again.'
-        });
+  List.find(query, '_id name description isPrivate items favIds cohortId', {
+    sort: { created_at: -1 }
+  })
+    .exec()
+    .then(docs => {
+      if (!docs) {
+        return resp.status(400).send({ message: 'No lists data found.' });
```
alemikolo: throw an error

server/controllers/list.js
```diff
@@ -81,23 +83,27 @@ const createList = (req, resp) => {
...
+  const {
+    user: { _id: userId },
+    params: { id: listId }
+  } = req;
+
+  List.findOneAndDelete({ _id: listId, ownerIds: userId })
+    .exec()
+    .then(doc => {
+      if (!doc) {
+        return resp.status(400).send({ message: 'List not found.' });
```
alemikolo: throw an error

server/controllers/cohort.js
```diff
@@ -332,73 +345,75 @@ const removeMember = (req, resp) => {
...
+    { _id: cohortId, ownerIds: { $all: [currentUserId, sanitizedUserId] } },
+    { $pull: { ownerIds: userId } }
+  )
+    .exec()
+    .then(doc => {
       if (!doc) {
-        return resp.status(400).send({
-          message: 'Cohort data not found.'
-        });
+        return resp.status(400).send({ message: 'Cohort data not found.' });
```
alemikolo: throw an error

server/controllers/cohort.js
```diff
@@ -332,73 +345,75 @@ const removeMember = (req, resp) => {
...
-
-      if (doc) {
-        return resp.status(200).send({
-          message: "User has been successfully set as a cohort's owner."
-        });
+    { $push: { ownerIds: userId } }
+  )
+    .exec()
+    .then(doc => {
+      if (!doc) {
```
alemikolo: throw an error

server/controllers/list.js
```diff
@@ -308,29 +319,31 @@ const updateListItem = (req, resp) => {
   List.findOneAndUpdate(
     {
       _id: listId,
-      'items._id': itemId,
-      $or: [{ ownerIds: req.user._id }, { memberIds: req.user._id }]
+      'items._id': sanitize(itemId),
+      $or: [{ ownerIds: userId }, { memberIds: userId }]
```
alemikolo: Is there a way to change this line to `memberIds: userId` or this is fixed in another task/branch?

server/controllers/list.js
```diff
@@ -308,29 +319,31 @@ const updateListItem = (req, resp) => {
...
-        return resp.status(400).send({
-          message:
-            'An error occurred while updating the list data. Please try again.'
-        });
+    { new: true }
+  )
+    .exec()
+    .then(doc => {
+      if (!doc) {
+        return resp.status(400).send({ message: 'List data not found.' });
```
alemikolo: throw an error

server/controllers/list.js
```diff
@@ -343,26 +356,28 @@ const voteForItem = (req, resp) => {
...
-        return resp.status(400).send({
-          message: 'An error occurred while voting. Please try again.'
-        });
+    { $push: { 'items.$.voterIds': userId } },
+    { new: true }
+  )
+    .exec()
+    .then(doc => {
+      if (!doc) {
+        return resp.status(400).send({ message: 'List data not found.' });
```
alemikolo: throw an error

server/controllers/cohort.js
alemikolo: You should throw an error here to make this promise rejected. The return doesn't do this and controller will try to send a second response which will cause an error/warning.

server/controllers/cohort.js
alemikolo: exec()

server/controllers/cohort.js
```diff
@@ -104,39 +108,45 @@ const updateCohortById = (req, resp) => {
...
-      if (err) {
-        return resp.status(400).send({
-          message:
-            'An error occurred while updating the cohort. Please try again.'
-        });
+    dataToUpdate
+  )
+    .exec()
+    .then(doc => {
+      if (!doc) {
```
alemikolo: throw an error -see above

server/controllers/cohort.js
```diff
@@ -228,25 +238,27 @@ const addToFavourites = (req, resp) => {
...
-    (err, doc) => {
-      if (err) {
-        return resp.status(400).send({
-          message: "Can't mark cohort as favourite. Please try again."
-        });
+    }
+  )
+    .exec()
+    .then(doc => {
+      if (!doc) {
```
alemikolo: throw an error

server/controllers/cohort.js
```diff
@@ -258,38 +270,39 @@ const removeFromFavourites = (req, resp) => {
...
-      if (err) {
-        return resp.status(400).send({
-          message: "Can't remove cohort from favourites. Please try again."
-        });
+    }
+  )
+    .exec()
+    .then(doc => {
+      if (!doc) {
+        resp.status(400).send({ message: 'Cohort data not found.' });
```
alemikolo: throw an error

server/controllers/cohort.js
```diff
@@ -332,73 +345,75 @@ const removeMember = (req, resp) => {
...
-      if (doc) {
-        return resp.status(200).send({
-          message: "User has been successfully set as a cohort's owner."
-        });
+    { $push: { ownerIds: userId } }
+  )
+    .exec()
+    .then(doc => {
+      if (!doc) {
+        return resp.status(400).send({ message: 'Cohort data not found.' });
```
alemikolo: throw an error

server/controllers/cohort.js
alemikolo: throw an error

server/controllers/list.js
```diff
@@ -375,30 +390,35 @@ const clearVote = (req, resp) => {
...
-        return resp.status(400).send({
-          message: 'An error occurred while voting. Please try again.'
-        });
+    { $pull: { 'items.$.voterIds': userId } },
+    { new: true }
+  )
+    .exec()
+    .then(doc => {
+      if (!doc) {
+        return resp.status(400).send({ message: 'List data not found.' });
```
alemikolo: throw an error

server/controllers/list.js
```diff
@@ -409,145 +429,163 @@ const updateListById = (req, resp) => {
...
-        return resp.status(400).send({
-          message:
-            'An error occurred while updating the list data. Please try again.'
-        });
+    dataToUpdate
+  )
+    .exec()
+    .then(doc => {
+      if (!doc) {
+        return resp.status(400).send({ message: 'List data not found.' });
```
alemikolo: throw an error

server/controllers/list.js
```diff
@@ -409,145 +429,163 @@ const updateListById = (req, resp) => {
...
-        return resp.status(400).send({
-          message: "Can't mark list as favourite. Please try again."
-        });
+      $push: { favIds: userId }
+    }
+  )
+    .exec()
+    .then(doc => {
+      if (!doc) {
+        return resp.status(400).send({ message: 'List data not found.' });
```
alemikolo: throw an error

server/controllers/list.js
```diff
@@ -409,145 +429,163 @@ const updateListById = (req, resp) => {
...
-        return resp.status(400).send({
-          message: "Can't remove list from favourites. Please try again."
-        });
+      $pull: { favIds: userId }
+    }
+  )
+    .exec()
+    .then(doc => {
+      if (!doc) {
+        return resp.status(400).send({ message: 'List data not found.' });
```
alemikolo: throw an error

server/controllers/list.js
```diff
@@ -409,145 +429,163 @@ const updateListById = (req, resp) => {
...
-        return resp.status(200).send({
-          message: 'Owner successfully removed from list.'
-        });
+    { _id: listId, ownerIds: { $all: [currentUserId, sanitize(userId)] } },
+    { $pull: { ownerIds: userId, memberIds: userId, viewersIds: userId } }
+  )
+    .exec()
+    .then(doc => {
+      if (!doc) {
+        return resp.status(400).send({ message: 'List data not found.' });
```
alemikolo: throw an error

server/controllers/list.js
```diff
@@ -409,145 +429,163 @@ const updateListById = (req, resp) => {
...
-      if (doc) {
-        return resp.status(200).send({
-          message: 'Member successfully removed from list.'
-        });
+    }
+  )
+    .exec()
+    .then(doc => {
+      if (!doc) {
+        return resp.status(400).send({ message: 'List data not found.' });
```
alemikolo: throw an error

alemikolo:
This is first and foremost one remark that is repeated in each changed controller... I could mark it once and wrote to fix it in all places...

server/controllers/cohort.js
adamklepacz: I wanted to make it consistent across every controller.

PR #125, opened 2019-05-06
[EOC-145] Edit list limited shared

alemikolo:
download (2)
I edited list schema and changed "isPrivate" field to "type". In all places where this prop was used to check list's privacy there is comparison now e.g. `type === ListType.LIMITED` both on front-end and back-end.


alemikolo:
I edited list schema and changed "isPrivate" field to "type". In all places where this prop was used to check list's privacy there is comparison now e.g. `type === ListType.LIMITED` both on front-end and back-end.

client/src/common/components/Members/components/MemberDetails.jsx
```diff
@@ -369,9 +367,9 @@ MemberDetails.propTypes = {
   isGuest: PropTypes.bool,
   isMember: PropTypes.bool,
   isOwner: PropTypes.bool,
-  isPrivate: PropTypes.bool,
   match: RouterMatchPropType.isRequired,
   route: PropTypes.string,
+  type: PropTypes.string,
```
marekrozmus: shouldn't this be required?

alemikolo: type shouldn't be required because MemberDetails component is used in cohort also, and cohorts don't have type property

server/controllers/list.js
```diff
@@ -860,12 +860,87 @@ const cloneItem = (req, resp) => {
...
+    .then(list => {
+      const {
+        memberIds,
+        name,
+        ownerIds,
+        type,
+        viewersIds: viewersCollection
+      } = list;
+      const message = `"${name}" list's type change to ${type}.`;
+
```
adamklepacz: remove empty line here

server/controllers/list.js
```diff
@@ -860,12 +860,87 @@ const cloneItem = (req, resp) => {
...
+    })
+    .then(list => {
+      const {
+        memberIds,
+        name,
+        ownerIds,
+        type,
+        viewersIds: viewersCollection
+      } = list;
+      const message = `"${name}" list's type change to ${type}.`;
```
adamklepacz: move that message, directly to resp.status(......

server/common/variables.js
```diff
@@ -1,5 +1,9 @@
 const DB_URL = process.env.DB_URL || 'mongodb://localhost:27017/eoc';
 const ENDPOINT_URL = 'http://localhost:8080';
 const FRONTEND_URL = 'http://localhost:3000';
+const ListType = Object.freeze({
+  LIMITED: 'limited',
+  SHARED: 'shared'
+});
```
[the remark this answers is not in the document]
alemikolo: ListType is defined in list module in index.jsx. But this types are needed on back-end also, so I put it in variables.js file because on back-end we don't have analogous place...

client/src/common/components/CardItem.jsx
```diff
@@ -57,7 +58,7 @@ class CardItem extends PureComponent {
         role="figure"
       >
         <header className="card-item__header">
-          {isPrivate && <LockIcon />}
+          {type === ListType.LIMITED && <LockIcon />}
```
adamklepacz: I think it would be better to store this in variable above return
```
const isLimitedList = type === ListType.LIMITED;

return (........
{isListLimited && <LockIcon />}
```

server/controllers/list.js
```diff
@@ -21,25 +21,27 @@ const {
...
-    isPrivate: isListPrivate,
     memberIds: userId,
     name,
     ownerIds: userId,
+    type,
     viewersIds: userId
   });
 
-  if (cohortId && !isListPrivate) {
+  if (cohortId && type === ListType.SHARED) {
```
adamklepacz: I think it would be better to store this in a `const isListShared = type === ListType.SHARED`

server/common/variables.js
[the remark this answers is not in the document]
adamklepacz: Please don't close conversation where you have responded to my notice

PR #127, opened 2019-05-08
[EOC-169] Add tests for backend and frontend utils 

adamklepacz:
Zrzut ekranu 2019-05-08 o 07 18 21
I did some tests, to check if utility functions work properly. During tests, some of the functions needed refactoring. 

I find out, that we have a little mess on backend - where we had to directions named 'UTILS' - they're merged right now into one.


server/common/utils/index.test.js
```diff
@@ -0,0 +1,391 @@
...
+  const userId = ObjectId();
+
+  list.cohortId = ObjectId();
+
+  const result = responseWithList(list, userId);
+
+  it('should return object with desired properties', () => {
+    expectedListProperties.map(property =>
+      expect(result).toHaveProperty(property)
+    );
```
marekrozmus: what if there are more props then expected?

server/common/utils/index.test.js
```diff
@@ -0,0 +1,391 @@
...
+  });
+
+  it('should return object without favIds property', () => {
```
marekrozmus: why such test?

server/common/utils/index.test.js
```diff
@@ -0,0 +1,391 @@
...
+      expect(result).toHaveProperty(property)
+    );
+  });
+});
+
+describe('function responseWithLists', () => {
+  const userId = ObjectId();
+  const lists = responseWithLists(listMock, userId);
+
+  it('expect to have only desired meta-data properties without sensitive data', () => {
```
marekrozmus: if so then maybe `responseWithLists` is not the right name? Maybe it should be `reponseWithListMetaData`?

server/common/utils/index.test.js
```diff
@@ -0,0 +1,391 @@
...
+
+  it('returned object should not have items and favIds properties', () => {
+    lists.map(list =>
+      notExpected.map(property => expect(list).not.toHaveProperty(property))
+    );
+  });
+});
+
+describe('function checkIfCurrentUserVoted', () => {
+  it('should be true', () => {
```
marekrozmus: names of tests for this function does not say anything

server/common/utils/index.test.js
```diff
@@ -0,0 +1,391 @@
...
+  });
+});
+
+describe('function responseWithItem', () => {
+  const item = singleItemMock[0];
+  const userId = ObjectId();
+  const result = responseWithItem(item, userId);
+
+  it('should return single item with desired properties', () => {
+    expectedItemProperties.map(property =>
```
marekrozmus: how does this check 'should return single item`?

server/common/utils/index.test.js
```diff
@@ -0,0 +1,391 @@
...
+
+  it('should return cohort data without favIds, memberIds, ownerIds', () => {
+    notExpected.map(property => expect(result).not.toHaveProperty(property));
+  });
+});
+
+describe('function checkIfArrayContainsUserId', () => {
+  const idsArray = ['123', '456', '789'];
+
+  it('should return true', () => {
```
marekrozmus: more decriptive test names are needed

server/common/utils/index.test.js
```diff
@@ -0,0 +1,391 @@
...
+  cohortsMock,
+  expectedCohortMetaDataProperties
+} = require('../../tests/__mocks__/cohortsMock');
+const {
+  expectedUsersProperties,
+  usersMock
+} = require('../../tests/__mocks__/usersMock');
+
+describe('function isValidMongoId', () => {
+  it('should return true', () => {
```
alemikolo: in my opinion this message isn't clear: when results should be true?

server/common/utils/index.test.js
```diff
@@ -0,0 +1,391 @@
...
+    const id = ObjectId();
+    const result = isValidMongoId(id);
+
+    expect(id).toBeInstanceOf(ObjectId);
+    expect(result).toBe(true);
+  });
+
+  it('should return false', () => {
```
alemikolo: unclear message

server/tests/__mocks__/cohortsMock.js
```diff
@@ -0,0 +1,50 @@
+const cohortsMock = [
+  {
+    _id: { $oid: '5ccfe9631ad8a35fee6fbc75' },
+    favIds: [],
+    isArchived: false,
+    memberIds: ['5c9dc5a619ae7924074940f4'],
```
alemikolo: shouldn't memberIds be an array of ObjectIds, not strings?

server/common/utils/index.js
```diff
@@ -56,45 +61,87 @@ const responseWithItems = (userId, list) => {
   const { items } = list;
 
   return _map(items, item => {
-    const { voterIds, ...rest } = item.toObject();
```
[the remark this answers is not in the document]
adamklepacz: becasue rest is having a lot of rubbish data like internal cache etc.
So at the end of this function return should be like this :
```
  return {
      ...rest._doc,
      isVoted: checkIfCurrentUserVoted(item, userId),
      votesCount: voterIds.length
    };
```
and the problem is with `   ...rest._doc,`. When I'm testing this function, data comming from MOCK, doesn't have any _doc nested document and tests always fails,

client/src/modules/authorization/model/reducer.test.js
```diff
@@ -1,34 +1,35 @@
-import { AuthorizationActionTypes } from './actions';
-import currentUser from 'modules/authorization/model/reducer';
+// import { AuthorizationActionTypes } from './actions';
```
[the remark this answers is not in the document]
adamklepacz: because those tests doesn't work properly

server/tests/__mocks__/cohortsMock.js
adamklepacz: doesn't matter. 
Helpers deals both with ObjectIds and string ids.
Tests pass in every case

server/common/utils/index.test.js
adamklepacz: I have another test to check unnecessary props below this one.

server/common/utils/index.test.js
adamklepacz: It should return an object without favIds.
The purpose of this test is, that someone might modify helper in the future and by mistake may put favIds to returning object.

server/common/utils/index.test.js
adamklepacz: it's named badly.
Should be `should return the item with desired properties`

server/common/utils/helpers.test.js
```diff
@@ -0,0 +1,391 @@
...
+});
+
+describe('function checkIfCohortMember', () => {
+  const cohort = cohortsMock[0];
+
+  it('should return true if passed user id is cohort member', () => {
+    const userId = ObjectId();
+    const result = checkIfCohortMember(cohort, userId);
+
+    expect(result).toBe(false);
```
alemikolo: if method should return true when user is a cohort member why this test result should be false?

server/common/utils/helpers.test.js
```diff
@@ -0,0 +1,391 @@
...
+  });
+
+  it('should return false if passed user id is not cohort member', () => {
+    const userId = 'dsad134sa';
+    const result = checkIfCohortMember(cohort, userId);
+
+    expect(result).toBe(false);
+  });
+
+  it('should return true if passed user id is cohort member', () => {
```
alemikolo: why this test is here? It has the same message like the first test in this suite, but expects different result? Maybe one of those tests is unnecessary or description should be more specific?

server/common/utils/helpers.test.js
```diff
@@ -0,0 +1,391 @@
...
+  cohortsMock,
+  expectedCohortMetaDataProperties
+} = require('../../tests/__mocks__/cohortsMock');
+const {
+  expectedUsersProperties,
+  usersMock
+} = require('../../tests/__mocks__/usersMock');
+
+describe('function isValidMongoId', () => {
+  it('checks if passed ID is a valid mongo id', () => {
```
marekrozmus: should be named `returns true if passed ID is a valid mongo id`

server/common/utils/helpers.test.js
```diff
@@ -0,0 +1,391 @@
...
+    const id = ObjectId();
+    const result = isValidMongoId(id);
+
+    expect(id).toBeInstanceOf(ObjectId);
+    expect(result).toBe(true);
+  });
+
+  it('check if passed ID is an invalid mongo id', () => {
```
marekrozmus: should be named `returns false if passed ID is a an invalid mongo id`. Do we even need this function? Can't we just use the ObjectId.isValid(id)?

server/common/utils/helpers.test.js
adamklepacz: all these test are necessary, they expect different results

server/common/utils/helpers.test.js
```diff
@@ -0,0 +1,391 @@
...
+    lists.map(list =>
+      notExpected.map(property => expect(list).not.toHaveProperty(property))
+    );
+  });
+});
+
+describe('function checkIfCurrentUserVoted', () => {
+  it('checks if current user id is inside voterIds array', () => {
+    const userId = '123';
+    const item = {
```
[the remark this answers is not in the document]
adamklepacz: beforeEach is not needed

server/common/utils/helpers.test.js
```diff
@@ -0,0 +1,371 @@
...
+
+  it('returns false if user is not inside favIds array', () => {
+    const userId = '999';
+    const favIds = ['123', '456', '789'];
+
+    const result = isUserFavourite(favIds, userId);
+    expect(result).toBe(false);
+  });
+
+  it('returns false if user is not inside favIds array', () => {
```
marekrozmus: add `empty`

server/common/utils/helpers.test.js
```diff
@@ -0,0 +1,371 @@
...
+    lists.map(list =>
+      expectedListMetaDataProperties.map(property =>
+        expect(list).toHaveProperty(property)
+      )
+    );
+  });
+
+  const notExpected = ['favIds', 'items'];
+
+  it('returned list objects should not have items and favIds properties', () => {
```
marekrozmus: `returned` => `returns`

server/common/utils/helpers.test.js
```diff
@@ -0,0 +1,371 @@
...
+  };
+
+  it('returns true if current user id is inside voterIds array', () => {
+    const userId = '123';
+    const result = checkIfCurrentUserVoted(item, userId);
+
+    expect(result).toBe(true);
+  });
+
+  it('returns true if current user id is inside voterIds array', () => {
```
marekrozmus: test with same name => change name, merge tests or remove one

server/common/utils/helpers.test.js
```diff
@@ -0,0 +1,371 @@
...
+  });
+
+  it('returns false if the passed array does not contain the user id', () => {
+    const userId = '999';
+    const result = checkIfArrayContainsUserId(idsArray, userId);
+
+    expect(result).toBe(false);
+  });
+
+  it('returns false if the passed array does not contain the user id', () => {
```
marekrozmus: test with same name => change name, merge tests or remove one

server/common/utils/helpers.test.js
```diff
@@ -0,0 +1,371 @@
...
+  const ownerIds = ['123', '456', '789'];
+  const members = responseWithCohortMembers(usersMock, ownerIds);
+
+  it('returns objects with desired properties', () => {
+    expectedUsersProperties.map(property =>
+      members.map(member => expect(member).toHaveProperty(property))
+    );
+  });
+
+  it('returns user objects without sensitive data like email', () => {
```
marekrozmus: maybe set as in any other places: `returns object without sensitive data`

server/common/utils/helpers.test.js
```diff
@@ -0,0 +1,371 @@
...
+  const cohort = cohortsMock[0];
+
+  it('returns false if passed user id is not a cohort member', () => {
+    const userId = ObjectId();
+    const result = checkIfCohortMember(cohort, userId);
+
+    expect(result).toBe(false);
+  });
+
+  it('returns false if passed user id is not a cohort member', () => {
```
marekrozmus: test with same name => change name, merge tests or remove one

server/common/utils/helpers.test.js
```diff
@@ -0,0 +1,371 @@
...
+
+  const result = responseWithCohortMember(user, ownerIds);
+
+  it('returns user object with desired properties', () => {
+    expectedUsersProperties.map(property =>
+      expect(result).toHaveProperty(property)
+    );
+  });
+
+  it('returns user object without sensitive data like email', () => {
```
marekrozmus: maybe set as in any other places: `returns object without sensitive data`

PR #128, opened 2019-05-08
Add super hot fix

adamklepacz:
Sorting didn't work for votes.


client/src/modules/list/components/ItemsContainer.jsx
```diff
@@ -63,7 +63,7 @@ class ItemsContainer extends Component {
         });
         break;
       case SortOptionType.VOTES:
-        result.sort((a, b) => a.voterIds.length - b.voterIds.length);
+        result.sort((a, b) => a.votesCount.length - b.votesCount.length);
```
alemikolo: length is unneeded, votesCount is a number

PR #131, opened 2019-05-09
[EOC-153] Add comments functionality

alemikolo:
I refactored items reducer to make it an object with items' ids as keys not an array to easily access specific item, especially after adding comments feature.

chrome-capture
chrome-capture (1)
chrome-capture (2)


client/src/modules/list/components/Items/model/actionTypes.js
```diff
@@ -1,3 +1,12 @@
+export const CommentActionTypes = Object.freeze({
+  ADD_FAILURE: 'comment/ADD_FAILURE',
+  ADD_REQUEST: 'comment/ADD_REQUEST',
```
marekrozmus: remove `REQUEST` action types as they are not used anywhere

server/common/utils/index.js
```diff
@@ -56,22 +56,56 @@ const responseWithItems = (userId, list) => {
...
+      _id,
+      authorId,
+      authorName,
+      createdAt,
+      description,
+      isOrdered,
+      link,
+      name,
+      voterIds
+    } = item.toObject();
```
marekrozmus: why all these changes (replacing ... operator)?

client/src/common/components/Comments/Comment.scss
```diff
@@ -8,6 +8,9 @@
 
   &__body {
     text-align: left;
+    width: calc(100% - 56px);
```
adamklepacz: you don't need this.
Just remove from .comment class `Justify-content `property

client/src/common/components/Comments/CommentsList.jsx
```diff
@@ -0,0 +1,132 @@
...
+      match: {
+        params: { id: listId }
+      }
+    } = this.props;
+    const action = addComment(listId, itemId, comment);
+
+    action.then(() => this.hideAddComment());
+
+    return action;
+  };
```
adamklepacz: those 3 lines could be:
```
return addComment(listId, itemId,comment).then(() => this.hideAddComment());
```

client/src/common/components/Comments/CommentsList.jsx
```diff
@@ -0,0 +1,132 @@
...
+
+  render() {
+    const { comments, isFormAccessible } = this.props;
+    const { isNewCommentVisible, pending } = this.state;
+
+    return (
+      <div className="comments">
+        <div className="comments__heading">
```
adamklepacz: you can use <header> tag here

client/src/common/components/Comments/NewComment.jsx
```diff
@@ -28,38 +40,60 @@ class NewComment extends PureComponent {
...
   };
 
-  handleAddNewComment = () => {
-    const { onAddNewComment } = this.props;
+  handleAddComment = () => {
+    const { onAddComment } = this.props;
     const { comment } = this.state;
 
-    onAddNewComment(comment);
+    if (!_isEmpty(_trim(comment))) {
```
adamklepacz: save this as a constant
const commentNotEmpty = !_isEmpty(_trim(comment));
and use inside this if

client/src/common/components/Comments/NewComment.jsx
```diff
@@ -28,38 +40,60 @@ class NewComment extends PureComponent {
...
+            this.setState({ pending: false });
+          }
+        });
+    }
   };
 
   handleCommentChange = value => this.setState({ comment: value });
 
   render() {
+    const { comment, pending } = this.state;
```
adamklepacz: add break line

client/src/modules/list/components/Items/model/actions.js
```diff
@@ -186,3 +215,41 @@ export const cloneItem = (listId, itemId) => dispatch => {
...
+
+export const addComment = (listId, itemId, text) => dispatch => {
+  dispatch(addCommentRequest());
+  return postData(`${ENDPOINT_URL}/comments/add-comment`, {
+    itemId,
+    listId,
+    text
+  })
+    .then(resp => resp.json())
+    .then(json => {
```
adamklepacz: unnecessary braces

client/src/modules/list/components/Items/model/actions.js
```diff
@@ -186,3 +215,41 @@ export const cloneItem = (listId, itemId) => dispatch => {
...
+        err.message
+      );
+    });
+};
+
+export const fetchComments = (listId, itemId) => dispatch => {
+  dispatch(fetchCommentsRequest());
+  return getData(`${ENDPOINT_URL}/comments/${listId}/${itemId}/data`)
+    .then(resp => resp.json())
+    .then(json => {
```
adamklepacz: unnecessary  braces

server/common/utils/index.js
alemikolo: I changed these methods back. I used lean() method in controllers to receive clean data object without any mongo's wrapper and the spread operator can be used safely.

client/src/modules/list/components/Items/model/reducer.js
```diff
@@ -1,52 +1,90 @@
...
+      } = action;
+
+      return { [comment._id]: comment, ...state };
+    }
+    case CommentActionTypes.FETCH_SUCCESS: {
+      const {
+        payload: { comments }
+      } = action;
+
+      return { ...comments };
```
adamklepacz: do you need spread?
try`  return comments`

server/common/utils/index.js
```diff
@@ -56,22 +56,56 @@ const responseWithItems = (userId, list) => {
...
+    _id,
+    authorId,
+    authorName,
+    createdAt,
+    description,
+    isOrdered,
+    link,
+    name,
+    voterIds
+  } = item.toObject();
```
adamklepacz: go back to ..rest and use .lean() on query function

server/common/utils/index.js
```diff
@@ -176,6 +210,38 @@ const responseWithListMembers = (
     isMember: checkIfArrayContainsUserId(memberIds, user._doc._id)
   }));
 
+const responseWithComment = (comment, avatarUrl, displayName) => {
+  const { _id, authorId, createdAt, text } = comment.toObject();
```
adamklepacz: you don't have to do toObject here to destructure given properties

server/common/utils/index.js
```diff
@@ -176,6 +210,38 @@ const responseWithListMembers = (
...
+
+  return {
+    _id,
+    authorAvatarUrl: avatarUrl,
```
adamklepacz: why are you maping to another name? is this necessary?

server/controllers/cohort.js
```diff
@@ -210,8 +211,17 @@ const deleteCohortById = (req, resp) => {
       }
 
       documentName = doc.name;
```
adamklepacz: add break line

server/controllers/comment.js
```diff
@@ -0,0 +1,107 @@
...
+  } = req;
+
+  List.findOne(
+    {
+      _id: sanitize(listId),
+      memberIds: userId,
+      'items._id': sanitize(itemId),
+      'items.isOrdered': true
+    },
+    { items: { $elemMatch: { isOrdered: true } } }
```
adamklepacz: why are U checking this in two places?? I mean here and line 22.

server/controllers/comment.js
```diff
@@ -0,0 +1,107 @@
...
+      }
+
+      resp.status(200).json(responseWithComments(comments));
+    })
+    .catch(err => {
+      if (err instanceof BadRequestException) {
+        const { status, message } = err;
+
+        return resp.status(status).send({ message });
+      }
```
adamklepacz: add line break

client/src/common/components/Comments/CommentsList.jsx
alemikolo: In the end, it probably would work the same. But here is the difference: I think external "then" should depend on adding a comment to the DB, not on hiding form and it would be the other way around if I implement this proposal.

client/src/common/components/Comments/NewComment.jsx
alemikolo: I save only `_trim(comment)` because it is used twice

server/common/utils/index.js
alemikolo: we store user's Id as authorId in comment and item model, we use authorName to store user's display name in comments and item on frontend, so adhering to the convention I map avatarUrl to authorAvatarUrl...

PR #142, opened 2019-05-20
[EOC-175] Bug handle missing avatar

adamklepacz:
Zrzut ekranu 2019-05-20 o 11 33 16

### DIFF
--- a/client/src/common/components/Comments/Comment.jsx
+++ b/client/src/common/components/Comments/Comment.jsx
@@ -5,7 +5,7 @@ import classNames from 'classnames';
 
 import { UserIcon } from 'assets/images/icons';
 import { dateFromString } from 'common/utils/helpers';
-import UserIconPlaceholder from '../../../assets/images/user.svg';
+import UserIconPlaceholder from 'assets/images/user.svg';
 
 class Comment extends PureComponent {
   state = {
--- a/client/src/common/components/Comments/Comment.jsx
+++ b/client/src/common/components/Comments/Comment.jsx
@@ -1,37 +1,58 @@
-import React from 'react';
+import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import Linkify from 'react-linkify';
+import classNames from 'classnames';
 
 import { UserIcon } from 'assets/images/icons';
 import { dateFromString } from 'common/utils/helpers';
+import UserIconPlaceholder from '../../../assets/images/user.svg';
 
-const Comment = ({ comment }) => {
-  const { authorName, authorAvatarUrl, createdAt, text } = comment;
-  const date = dateFromString(createdAt);
+class Comment extends PureComponent {
+  state = {
+    isAvatarError: false
+  };
 
-  return (
-    <div className="comment">
-      <div className="comment__avatar">
-        {authorAvatarUrl ? (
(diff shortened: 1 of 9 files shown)


server/controllers/list.js
```diff
@@ -777,14 +777,15 @@ const addViewer = (req, resp) => {
...
+      const userExists = checkIfArrayContainsUserId(viewersIds, newMemberId);
+
+      if (userExists) {
+        throw new BadRequestException('User is already a member.');
+      }
 
       user = userData;
 
-      if (userNotExists) {
+      if (!userExists) {
```
marekrozmus: this check is not needed - this code won't be run as exception will be thrown in line 783

client/src/common/components/Members/components/MemberButton.jsx
```diff
@@ -1,22 +1,41 @@
...
-    <img
-      alt={`${member.displayName} avatar`}
-      className="member-button__avatar"
-      onClick={onDisplayDetails}
-      src={member.avatarUrl}
-      title={member.displayName}
-    />
-  </button>
-);
+import UserIconPlaceholder from 'assets/images/user.svg';
```
alemikolo: I'm not sure, but if we have to import images we should use '../../assets/images/user.svg' now.

adamklepacz: @marekrozmus   Should it be a full relative path?

alemikolo:
Comment is one more place, that this solution should be implemented.

client/src/common/components/Comments/Comment.jsx
```diff
@@ -1,37 +1,58 @@
-import React from 'react';
+import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import Linkify from 'react-linkify';
+import classNames from 'classnames';
 
 import { UserIcon } from 'assets/images/icons';
 import { dateFromString } from 'common/utils/helpers';
+import UserIconPlaceholder from '../../../assets/images/user.svg';
```
marekrozmus: why not just 'assets/images/user.svg'?

client/src/common/components/Members/components/MemberButton.jsx
```diff
@@ -1,22 +1,41 @@
...
-    <img
-      alt={`${member.displayName} avatar`}
-      className="member-button__avatar"
-      onClick={onDisplayDetails}
-      src={member.avatarUrl}
-      title={member.displayName}
-    />
-  </button>
-);
+import UserIconPlaceholder from '../../../../assets/images/user.svg';
```
marekrozmus: same here

client/src/common/components/Toolbar/components/UserBar/UserBar.jsx
```diff
@@ -9,9 +9,11 @@ import { getCurrentUser } from 'modules/authorization/model/selectors';
 import { UserPropType } from 'common/constants/propTypes';
 import { LogoutIcon, UserIcon, CohortIcon } from 'assets/images/icons';
 import Overlay, { OverlayStyleType } from 'common/components/Overlay';
+import UserIconPlaceholder from '../../../../../assets/images/user.svg';
```
marekrozmus: and here

PR #151, opened 2019-05-24
[EOC-134] Toggle list tile view on dashboard

alemikolo:
chrome-capture


client/src/common/components/CardPlus.scss
```diff
@@ -7,14 +7,15 @@
   color: $text-normal;
   display: flex;
   flex-direction: column;
-  height: 120px;
+  height: 100%;
+  // height: 120px;
```
marekrozmus: remove commented out code

client/src/common/components/CardPlus.scss
```diff
@@ -7,14 +7,15 @@
...
   justify-content: center;
   padding: 8px;
   width: 100%;
 
-  @include breakpoint($medium) {
-    width: 210px;
-  }
+  // @include breakpoint($medium) {
+  //   width: 210px;
+  // }
```
marekrozmus: remove commented out code

client/src/common/components/CollectionView.scss
```diff
@@ -30,46 +31,49 @@
...
     }
   }
 
   &__button {
     background: transparent;
     border: 0;
     cursor: pointer;
-    display: block;
     height: 100%;
+    width: 100%;
```
marekrozmus: why moved? previously was sorted alphabetically

client/src/modules/cohort/index.jsx
```diff
@@ -22,10 +20,10 @@ import FormDialog from 'common/components/FormDialog';
 import { archiveCohort, fetchCohortDetails } from './model/actions';
 import Dialog, { DialogContext } from 'common/components/Dialog';
 import ArchivedCohort from 'modules/cohort/components/ArchivedCohort';
-import GridList from 'common/components/GridList';
+import CollectionView from '../../common/components/CollectionView';
```
marekrozmus: why '../../'?

client/src/app/components/Layout/Layout.jsx
```diff
@@ -16,8 +16,15 @@ import Page404 from 'common/components/Page404';
 import About from 'modules/about';
 import PrivacyPolicy from 'modules/privacy-policy';
 import Cohorts from 'modules/cohort/components/Cohorts';
+import Toolbar, { ToolbarItem } from './Toolbar';
+import { ListModeIcon, TilesModeIcon } from 'assets/images/icons';
+import { Routes } from 'common/constants/enums';
+
+export class Layout extends PureComponent {
+  state = {
+    listMode: true
```
adamklepacz: create ENUM for this.
listMode true is not clear 
Would be better to have
```
 viewType: ViewType.LIST 
 viewType: ViewType.TILE
```

or diplayType: DisplayType.LIST   itd
you even named one of the methods as so.
more readable for future I think

client/src/app/components/Layout/Layout.jsx
```diff
@@ -41,8 +48,23 @@ export class Layout extends Component {
     loginUser();
   };
 
+  displayModeSwitch = () => {
```
adamklepacz: I was thinking that this method was for switching between list and tile view.
It should be named like  isListView = () => 
beacuse this function returns bollean I think isListView is the best name for this.

then in the render method, you will have 
this.isListView && <Render Something>   and it is very clear that when it is a list view display something...otherwise not :)

client/src/app/components/Layout/Layout.jsx
```diff
@@ -41,8 +48,23 @@ export class Layout extends Component {
...
+      location: { pathname }
+    } = this.props;
+
+    return (
+      pathname.includes(`${Routes.COHORT}/`) ||
+      pathname.includes(Routes.DASHBOARD)
+    );
+  };
+
+  handleListModeChange = () =>
```
adamklepacz: when having an ENUM for this you will have two clear method 

switchToListViewType = () =>

switchToTileViewType = () =>

client/src/app/components/Layout/Layout.jsx
```diff
@@ -41,8 +48,23 @@ export class Layout extends Component {
     loginUser();
   };
 
+  displayModeSwitch = () => {
+    const {
+      location: { pathname }
+    } = this.props;
+
+    return (
+      pathname.includes(`${Routes.COHORT}/`) ||
```
adamklepacz: and when the method will be named `isListView`  you can only check for one condition 

pathname.include(`Routes.LIST/`)   and whenever it returns true elements would be displayed, otherwise not

client/src/common/components/CollectionView.jsx
```diff
@@ -51,48 +52,71 @@ class GridList extends PureComponent {
...
+    return (
+      <li className="collection__item">
+        <button className="collection__button" onClick={onAddNew} type="button">
+          <CardPlus />
+        </button>
+      </li>
+    );
+  };
+
+  renderListMode = () => {
```
adamklepacz: renderAsList

client/src/common/components/CollectionView.jsx
```diff
@@ -51,48 +52,71 @@ class GridList extends PureComponent {
...
+          item={item}
+          onCardClick={this.handleCardClick(route, item._id)}
+          onFavClick={this.handleFavClick(item._id, item.isFavourite)}
+          route={route}
+        />
+      </li>
+    ));
+  };
+
+  renderTilesMode = () => {
```
adamklepacz: renderAsTiles 

- renderTilesMode means that you would like to render some "Mode" as element, component or switch button or something else. The mode is like an option for me that I can choose by interaction.

client/src/common/components/CollectionView.jsx
```diff
@@ -51,48 +52,71 @@ class GridList extends PureComponent {
...
+      </li>
+    );
+  };
+
+  renderListMode = () => {
+    const { color, items, route } = this.props;
+
+    return _map(items, item => (
+      <li className="collection__item" key={item._id}>
+        <ListModeItem
```
adamklepacz: consider renaming to ListViewItem

client/src/common/components/CollectionView.jsx
```diff
@@ -51,48 +52,71 @@ class GridList extends PureComponent {
...
+      </li>
+    ));
+  };
+
+  renderTilesMode = () => {
+    const { color, items, route } = this.props;
+
+    return _map(items, item => (
+      <li className="collection__item" key={item._id}>
+        <TilesModeItem
```
adamklepacz: consider renaming to TilesViewItem

client/src/common/components/CollectionView.jsx
```diff
@@ -51,48 +52,71 @@ class GridList extends PureComponent {
...
+        </button>
+      </li>
+    );
+  };
+
+  renderListMode = () => {
+    const { color, items, route } = this.props;
+
+    return _map(items, item => (
+      <li className="collection__item" key={item._id}>
```
adamklepacz: destructure _id, isFavourite here and in other places.

client/src/common/components/TilesModeItem.jsx
```diff
@@ -50,9 +44,9 @@ class CardItem extends PureComponent {
     return (
       <div
         className={classNames('card-item', {
-          'card-item--orange': color === CardColorType.ORANGE,
-          'card-item--archived': color === CardColorType.ARCHIVED,
-          'card-item--brown': color === CardColorType.BROWN
+          'card-item--orange': color === ColorType.ORANGE,
+          'card-item--archived': color === ColorType.GRAY,
```
adamklepacz: so change the class name also from archived to `'card-item--gray' `

client/src/modules/cohort/index.jsx
```diff
@@ -22,10 +20,10 @@ import FormDialog from 'common/components/FormDialog';
...
 import { ListType } from 'modules/list/consts';
 import MembersBox from 'common/components/Members';
-import { Routes } from 'common/constants/enums';
+import { ColorType, Routes } from 'common/constants/enums';
 import CohortHeader from './components/CohortHeader';
 import Preloader from '../../common/components/Preloader';
```
adamklepacz: change the path also here and check if there are any others `../../` routes like this across the project

client/src/app/components/Layout/Layout.jsx
alemikolo: I agree that this method name is unclear but this method doesn't switch between list and tiles views. Its duty is to check if the "view-switching" button should be displayed (this button should be displayed on the dashboard and on cohort view only because only on those two views lists are displayed). I moved the toolbar to the layout folder (just like you did in your task - on my demand of course). Maybe it was the wrong idea. Now, in this method I check the current route and base on this checking decide if this button should be displayed - handlers stay in layout component. I can move toolbar back (to main module's components) and pass variables and all handlers via props to dashboard/cohort. I changed this method name to `isViewSwitchDisplayed`

adamklepacz: No leave the toolbar where it is.

I know that this method is for checking if the switch button should be displayed but he name is to wrong and now even to complicated.

take a look, what this method is doing??  It's is checking if current path is LIST ? Am I right?

And later in render function 
`If(isListView) --> display switch button `  it's clear and easy to read for everybody from outside the project 

isViewTypeSwitchDisplayed would be a method to check if the button is displayed, rendered or something else. 

You are checking inside this method if path is /list and the method name is isViewTypeSwitchDisplated - does it make sense to you?

PR #152, opened 2019-05-24
[EOC-135] Invite user via email

adamklepacz:
When adding a new user to list or cohort there is a check if the user exists in the database or not. Later there is a response sent from backend if the user is CREATED or NO_USER.

If NO_USER came as a response it's mean that the user has no account in our app. 
Invite box is displayed and the user can decide if he wants to invite new member or cancel.

Invitation link directs to EOC dev app **(it's a temporary solution, later on, we need to direct to production version and we need to implement feature _"JOIN LIST/ JOIN COHORT"_ to send him to cohort or list after successful login).**

I've used [www.sendgrid.com](url) as mail sender service. The same tool they have in SandTime app. RIght now - API key is generated from my account. For development, we need to generate a new API key from our company account.

Zrzut ekranu 2019-05-24 o 20 08 28
Zrzut ekranu 2019-05-24 o 20 08 38
Zrzut ekranu 2019-05-24 o 20 08 58


client/src/common/components/Members/components/InviteNewUser.jsx
```diff
@@ -0,0 +1,44 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+
+import PendingButton from 'common/components/PendingButton';
+import { PreloaderTheme } from 'common/components/Preloader';
+
+class InviteNewUser extends PureComponent {
```
alemikolo: It could be a functional component...

client/src/common/components/Members/index.jsx
```diff
@@ -85,6 +101,21 @@ class MembersBox extends PureComponent {
...
+    const { email } = this.state;
+    const { inviteUser } = this.props;
+
+    return inviteUser(email).then(() => {
+      this.hideForm();
+      this.hideInvitationBox();
+    });
+  };
+
+  handleCancel = () => {
```
alemikolo: I think email should be removed from state on cancel;

client/src/modules/list/model/actions.js
```diff
@@ -373,14 +374,24 @@ export const addListViewer = (listId, email) => dispatch =>
...
-      );
+      if (json) {
+        dispatch(addViewerSuccess(json, listId));
+        createNotificationWithTimeout(
+          dispatch,
+          NotificationType.SUCCESS,
+          json.message || 'Viewer added successfully.'
+        );
+
+        return UserCreationStatus.CREATED;
```
alemikolo: You don't create the user here - You add/invite the user to the list:
`UserAddingStatus.ADDED` or `UserInvitingStatus.INVITED`

server/mailer/index.js
```diff
@@ -0,0 +1,35 @@
+const SendGridMail = require('@sendgrid/mail');
+
+const mailTemplate = require('./mail-template');
+
+const { SENDGRID_API_KEY } = process.env;
+
+SendGridMail.setApiKey(SENDGRID_API_KEY);
+
+const sendInvitation = (req, resp) => {
+  const { email: receiver } = req.body;
```
alemikolo: only wonder: shouldn't we sanitize an email address here?

server/mailer/mail-template.js
```diff
@@ -0,0 +1,85 @@
...
+              </tr>
+              <tr style="width:100%;">
+                <td style="width:100%;">${sender}</td>
+              </tr>
+              <tr style="width:100%;">
+                <td style="padding-top:40px;
+                           width:100%;">
+                  <a href="${TEMP_DEV}">
+                    <input
+                      value="JOIN EOC"
```
alemikolo: alphabetic order

server/mailer/index.js
```diff
@@ -0,0 +1,35 @@
...
+  const {
+    user: { email: sender }
+  } = req;
+
+  const message = {
+    to: receiver,
+    from: sender,
+    subject: 'Join EOC!',
+    html: mailTemplate(receiver, sender)
```
marekrozmus: this should also take as param host. Can be taken from 'req', and remove TEMP_DEV

server/mailer/index.js
adamklepacz: I think we don't need to sanitize data as long as we're not making any query to db. 
But I think we should create task for email validation.

client/src/common/components/Members/components/InviteNewUser.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+import PendingButton from 'common/components/PendingButton';
+import { PreloaderTheme } from 'common/components/Preloader';
+
+const InviteNewUser = ({ email, onCancel, onInvite }) => (
+  <div className="invite-user">
+    <p className="invite-user__message">
+      User
+      <span className="invite-user__email">{` ${email} `}</span>
+      has no account in EOC app. Would you like to invite via email?
```
marekrozmus: `to send invite`
`EOC` should be some constant so that in future we won't have to change the used name in many places. Use some constant like APP_NAME

server/mailer/index.js
```diff
@@ -0,0 +1,38 @@
...
+    user: { email: sender }
+  } = req;
+  const { protocol } = req;
+  const host = req.get('host');
+  const url = `${protocol}://${host}`;
+
+  const message = {
+    to: receiver,
+    from: sender,
+    subject: 'Join EOC!',
```
marekrozmus: `APP_NAME` constant should be used

server/mailer/mail-template.js
```diff
@@ -0,0 +1,83 @@
...
+                     border-radius:4px;
+                     margin: 20px 20px 20px 20px;
+                     padding: 40px;
+                     box-shadow:0 1px 3px #B7C0CC, 0 1px 2px #B7C0CC;"
+            >
+              <tr style="width:100%;">
+                <td style="width:100%;"
+                    align="left"
+                >
+                  EOC App
```
marekrozmus: `APP_NAME' should be used

server/mailer/mail-template.js
```diff
@@ -0,0 +1,83 @@
...
+                </td>
+              <tr style="width:100%;">
```
marekrozmus: is `</tr>` missing here

server/mailer/mail-template.js
```diff
@@ -0,0 +1,83 @@
...
+                  EOC App
+                </td>
+              <tr style="width:100%;"> 
+                <td
+                  style="font-size: 22px;
+                         padding-top:20px;
+                         padding-bottom:0px;
+                         width:100%;"
+                >
+                 Join EOC today!
```
marekrozmus: `APP_NAME` should be used

PR #153, opened 2019-05-25
[EOC-187] Edit list item name

adamklepacz:
Edit is available on  Enter press, Escape press, and click outside of the name component.
updateitemu

### DIFF
--- a/client/src/common/components/GridList.jsx
+++ b/client/src/common/components/GridList.jsx
@@ -18,11 +18,10 @@ import {
   removeCohortFromFavourites
 } from 'modules/cohort/model/actions';
 import Preloader from 'common/components/Preloader';
-import { stopPropagation } from 'common/utils/helpers';
 
 class GridList extends PureComponent {
   handleFavClick = (itemId, isFavourite) => event => {
-    stopPropagation(event);
+    event.stopPropagation();
 
     const {
       addCohortToFavourites,
--- a/client/src/common/components/GridList.jsx
+++ b/client/src/common/components/GridList.jsx
@@ -18,10 +18,12 @@ import {
   removeCohortFromFavourites
 } from 'modules/cohort/model/actions';
 import Preloader from 'common/components/Preloader';
+import { stopPropagation } from 'common/utils/helpers';
 
 class GridList extends PureComponent {
   handleFavClick = (itemId, isFavourite) => event => {
-    event.stopPropagation();
+    stopPropagation(event);
+
     const {
       addCohortToFavourites,
       addListToFavourites,
--- a/client/src/common/components/GridList.jsx
+++ b/client/src/common/components/GridList.jsx
@@ -22,7 +22,6 @@ import Preloader from 'common/components/Preloader';
 class GridList extends PureComponent {
   handleFavClick = (itemId, isFavourite) => event => {
     event.stopPropagation();
-
     const {
       addCohortToFavourites,
       addListToFavourites,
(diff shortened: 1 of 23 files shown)


client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -185,6 +188,12 @@ class ListItem extends PureComponent {
     );
   };
 
+  preventDefault = event => preventDefault(event);
```
alemikolo: is this additional wrapper for preventDefault necessary? Where this method is used in this component?

client/src/modules/list/components/Items/ListItemName/ListItemName.jsx
```diff
@@ -0,0 +1,168 @@
...
+      updateListItem
+    } = this.props;
+    const isNameUpdated = updatedName !== name;
+    const canBeUpdated = updatedName.trim().length > 1;
+
+    if (canBeUpdated && isNameUpdated) {
+      this.setState({ pending: true });
+
+      updateListItem(listId, itemId, { name: updatedName }).then(() => {
+        this.setState({
```
alemikolo: if updateListItem will fail pending won't stop.

client/src/modules/list/components/Items/ListItemName/ListItemName.jsx
```diff
@@ -0,0 +1,168 @@
...
+          pending: false,
```
alemikolo: As we discussed: because setting pending depends on async action when the component is unmounted setting state causes an error...

client/src/common/components/GridList.jsx
```diff
@@ -22,6 +22,7 @@ import Preloader from 'common/components/Preloader';
 class GridList extends PureComponent {
   handleFavClick = (itemId, isFavourite) => event => {
     event.stopPropagation();
+
```
marekrozmus: why all these new lines? Are they specific to this PR?

client/src/modules/list/components/Items/ListItemName/ListItemName.jsx
```diff
@@ -0,0 +1,167 @@
...
+
+    if (isNameInputFocused) {
+      document.addEventListener('keydown', this.handleKeyPress);
+      document.addEventListener('mousedown', this.handleClickOutside);
+    }
+
+    if (!isNameInputFocused) {
+      document.removeEventListener('keydown', this.handleKeyPress);
+      document.removeEventListener('mousedown', this.handleClickOutside);
+    }
```
marekrozmus: these 'if' blocks above should go to: 'handleNameInputFocus' and 'handleNameInputBlur'

client/src/common/components/GridList.jsx
adamklepacz: Not specific, I just thought it looks cleaner to have a line of a break after a function call

marekrozmus: Please roll back as it mess up git history. If you want to make some code cleanup for not relevant files then create separate branch.

client/src/modules/list/components/Items/ListItemName/ListItemName.jsx
```diff
@@ -0,0 +1,164 @@
...
+    this.nameInput = React.createRef();
+    this.listItemName = React.createRef();
+  }
+
+  componentDidUpdate() {
+    const { isNameInputFocused, name } = this.state;
+
+    if (isNameInputFocused) {
+      document.addEventListener('keydown', this.handleKeyPress);
+      document.addEventListener('mousedown', this.handleClickOutside);
```
marekrozmus: move these listeners to `handleNameInputFocus '

PR #155, opened 2019-05-27
[EOC-133] remove favs from cohorts

adamklepacz:


server/models/cohort.model.js
```diff
@@ -6,7 +6,6 @@ const { ObjectId } = Schema.Types;
 const CohortSchema = new Schema(
   {
     description: { type: String },
-    favIds: [ObjectId],
```
marekrozmus: what about db structure update iwth migration? please check if needed

PR #157, opened 2019-05-28
[EOC-188] Archive list items

alemikolo:
chrome-capture
chrome-capture (1)
chrome-capture (2)


client/src/modules/list/components/ArchivedItemsContainer.jsx
```diff
@@ -0,0 +1,114 @@
...
+import { RouterMatchPropType } from 'common/constants/propTypes';
+import Preloader from 'common/components/Preloader';
+
+class ArchivedItemsContainer extends PureComponent {
+  state = {
+    areArchivedItemsVisible: false,
+    pendingForArchivedItems: false
+  };
+
+  componentDidMount() {}
```
marekrozmus: ??

client/src/modules/list/components/ArchivedItemsContainer.jsx
```diff
@@ -0,0 +1,114 @@
...
+
+  handleArchivedItemsVisibility = () =>
+    this.setState(
+      ({ areArchivedItemsVisible }) => ({
+        areArchivedItemsVisible: !areArchivedItemsVisible
+      }),
+      () => this.handleArchivedItemsData()
```
marekrozmus: please check if `this.handleArchivedItemsData` will be enough

client/src/modules/list/components/ArchivedItemsContainer.jsx
```diff
@@ -0,0 +1,114 @@
...
+    const { archivedItems, isMember } = this.props;
+
+    return (
+      <div className="archived-items">
+        <button
+          className="link-button"
+          onClick={this.handleArchivedItemsVisibility}
+          type="button"
+        >
+          {` ${areArchivedItemsVisible ? 'hide' : 'show'} archived items`}
```
marekrozmus: why space before text?

client/src/modules/list/index.jsx
```diff
@@ -153,37 +156,41 @@ class List extends Component {
...
                 )}
+                <div className="list__members">
+                  <button
+                    className="link-button"
+                    onClick={this.handleMembersBoxVisibility}
+                    type="button"
+                  >
+                    {` ${isMembersBoxVisible ? 'hide' : 'show'} sack's members`}
+                  </button>
+
```
marekrozmus: remove new line

server/controllers/list.js
```diff
@@ -875,19 +888,23 @@ const updateItemDetails = (req, resp) => {
         itemToUpdate.link = link;
       }
 
-      if (isOrdered !== null) {
+      if (isOrdered !== undefined) {
```
adamklepacz: check if this will be enough `if(!isOrdered) `

server/controllers/list.js
```diff
@@ -875,19 +888,23 @@ const updateItemDetails = (req, resp) => {
...
         itemToUpdate.isOrdered = isOrdered;
+        updateMessage = `Item "${itemName}" successfully ordered.`;
       }
 
       if (authorId) {
         itemToUpdate.authorId = authorId;
       }
 
+      if (isArchived !== undefined) {
```
adamklepacz: check if this will be enough `if(!isArchived) `

client/src/app/styles/_buttons.scss
```diff
@@ -47,4 +53,12 @@
   &:hover {
     color: $text-orange;
   }
+
+  &:disabled {
+    cursor: not-allowed;
```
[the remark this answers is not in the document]
alemikolo: But this is link-button it has no background, I removed underline and turn off hover change?

client/src/modules/list/index.jsx
```diff
@@ -153,37 +156,40 @@ class List extends Component {
                     {isMember && <InputBar />}
                   </ItemsContainer>
                   <ItemsContainer
-                    archived
                     isMember={isMember}
                     items={doneItems}
+                    ordered
```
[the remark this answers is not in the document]
alemikolo: It was here since the beginning but as an archived prop I changed it to ordered because now the archived word is limited to real archive functionality. It is passed to the second itemsConteiner and it is always ordered in this case.

server/controllers/list.js
alemikolo: It won't be enough, because sometimes this prop isn't passed to this controller, in solution: `if(!isOrdered)` if isOrdered or isArchived prop would be false item wouldn't be updated...

server/controllers/list.js
alemikolo: It won't be enough, because sometimes this prop isn't passed to this controller, in solution: `if(!isArchived)` if isArchived or isOrdered prop would be false item wouldn't be updated...

PR #165, opened 2019-05-31
[EOC-182] Add link to description and remove link field

alemikolo:
Added [react-textarea-autosize](https://www.npmjs.com/package/react-textarea-autosize) library to easily manage textarea height. 
Blocked editing description for done items.
chrome-capture


client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,295 @@
...
+  componentDidUpdate(prevProps, prevState) {
+    this.checkIfDescriptionUpdated();
+    const { isDescriptionEdited: previousIsDescriptionEdited } = prevState;
+    const { isDescriptionEdited } = this.state;
+
+    if (!previousIsDescriptionEdited && isDescriptionEdited) {
+      this.descriptionTextarea.current.focus();
+      this.handleFocus();
+      document.addEventListener('keydown', this.handleKeyPress);
+      document.addEventListener('mousedown', this.handleClickOutside);
```
marekrozmus: move these event listeners to `handleFocus`

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    if (previousIsDescriptionEdited && !isDescriptionEdited) {
+      document.removeEventListener('keydown', this.handleKeyPress);
```
marekrozmus: move these event listeners to `handleBlur`

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,295 @@
...
+    if (isFocused && code === KeyCodes.ESCAPE) {
+      this.handleDescriptionUpdate();
+    }
+  };
+
+  handleFocus = () => this.setState({ isFocused: true });
+
+  handleBlur = () => this.setState({ isFocused: false });
+
+  handleClickOutside = event => {
```
marekrozmus: remove this and handle that in `handleBlur`

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,285 @@
...
+    this.checkIfDescriptionUpdated();
+    const { isDescriptionEdited: previousIsDescriptionEdited } = prevState;
+    const { isDescriptionEdited } = this.state;
+
+    if (!previousIsDescriptionEdited && isDescriptionEdited) {
+      this.descriptionTextarea.current.focus();
+      this.handleFocus();
+    }
+
+    if (previousIsDescriptionEdited && !isDescriptionEdited) {
```
adamklepacz: isPreviousDescriptionEdited

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,285 @@
...
+      },
+      name,
+      updateListItem
+    } = this.props;
+
+    if (disabled) {
+      return;
+    }
+
+    const isDescriptionUpdated = !_isEqual(
```
adamklepacz: why are you checking here if the description is updated if you have created a method for this?

this.checkIfDescriptionUpdated ?
You can set state in this method, and return a boolean value at the end and use it here

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,285 @@
...
+          if (!(err instanceof AbortPromiseException)) {
+            this.setState({ pending: false });
+          }
+        });
+    } else {
+      this.handleStopEditing();
+    }
+  };
+
+  checkIfDescriptionUpdated = () => {
```
adamklepacz: rename to `isDescriptionUpdated`

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,285 @@
...
+    const { description: previousDescription } = this.props;
+
```
adamklepacz: remove break line

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,285 @@
...
+  handleBlur = () => {
+    this.setState({ isFocused: false });
+    document.removeEventListener('keydown', this.handleKeyPress);
+    this.handleDescriptionUpdate();
+  };
+
+  handleStartEditing = () => {
+    this.setState({ isDescriptionEdited: true });
+  };
```
adamklepacz: do you need this method and property?
can't you assume that if an input has the focus it is also edited
and if it has blur it is not edited

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,285 @@
...
+    const { pending } = this.state;
+
+    return (
+      <Linkify
+        properties={{
+          target: '_blank',
+          className: 'list-item-description__link'
+        }}
+      >
+        <div
```
adamklepacz: use p tag for text to be more semantic

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,285 @@
...
+
+    return (
+      <Textarea
+        className="list-item-description__edit-field"
+        disabled={disabled || pending}
+        maxRows={5}
+        minRows={1}
+        name="description"
+        onChange={disabled ? null : this.handleDescriptionChange}
+        onBlur={this.handleBlur}
```
adamklepacz: sort props

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,285 @@
...
+  description: PropTypes.string.isRequired,
+  disabled: PropTypes.bool,
+  itemId: PropTypes.string.isRequired,
+  match: RouterMatchPropType.isRequired,
+  name: PropTypes.string.isRequired,
+
+  updateListItem: PropTypes.func.isRequired
+};
+
+const mapStateToProps = state => ({});
```
adamklepacz: remove this if you're not using this

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,285 @@
...
+
+export default withRouter(
+  connect(
+    mapStateToProps,
```
adamklepacz: pass null here

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
alemikolo: Yes, it's necessary. But textarea isn't always displayed. When there is no description add-button is displayed or if there is description we display a paragraph with a description to replace entered link by an anchor tag (we can't to that inside input/textarea). So when user clicks add-button or description div, this handler is invoked set `isDescriptionEdited: true` and render knows that textarea has to be displayed.

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -283,51 +212,38 @@ class ListItem extends PureComponent {
...
   };
 
+  renderDescription = () => {
+    const {
+      data: { _id: itemId, description, isOrdered, name },
+      isMember
+    } = this.props;
+    const isFieldDisabled = !isMember || isOrdered;
+
+    if (!description && (isOrdered || !isMember)) {
```
marekrozmus: `(isOrdered || !isMember)` is same as `isFieldDisabled`, use `isFieldDisabled`

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,279 @@
...
+    document.addEventListener('keydown', this.handleKeyPress);
+  };
+
+  handleBlur = () => {
+    this.setState({ isFocused: false });
+    document.removeEventListener('keydown', this.handleKeyPress);
+    this.handleDescriptionUpdate();
+  };
+
+  handleStartEditing = () => {
```
marekrozmus: remove braces

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,279 @@
...
+    this.setState({ isDescriptionEdited: true });
+  };
+
+  handleStopEditing = () => {
```
marekrozmus: remove braces

client/src/modules/list/components/Items/model/actions.js
```diff
@@ -197,26 +197,25 @@ export const clearVote = (itemId, listId) => dispatch =>
...
       );
     })
-    .catch(err => {
+    .catch(() => {
       dispatch(updateListItemFailure());
       createNotificationWithTimeout(
         dispatch,
         NotificationType.ERROR,
-        err.message
+        `Updating "${name}" item failed. Please try again.`
```
marekrozmus: Change to `Failed to update...` - we need to have all infos in one form.

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -283,51 +212,38 @@ class ListItem extends PureComponent {
...
+      data: { _id: itemId, description, isOrdered, name },
+      isMember
+    } = this.props;
+    const isFieldDisabled = !isMember || isOrdered;
+
+    if (!description && isFieldDisabled) {
+      return;
+    }
+
+    return (
```
marekrozmus: make one `return`

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,275 @@
...
+    return (
+      <Fragment>
+        <div onClick={disabled ? null : this.handleStartEditing}>
+          {isTitleVisible && (
+            <h2 className="list-item-description__title">Description</h2>
+          )}
+          {isDescriptionEdited ? (
+            this.renderTextarea()
+          ) : (
+            <Fragment>
```
marekrozmus: is this needed?

alemikolo: Unfortunately yes. Without <Fragment> the Eslint shows error: "Do not nest ternary expressions."

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,275 @@
...
+    const { isDescriptionEdited } = this.state;
+
+    if (!isPreviousIsDescriptionEdited && isDescriptionEdited) {
+      this.descriptionTextarea.current.focus();
+      this.handleFocus();
+    }
+
+    if (isPreviousIsDescriptionEdited && !isDescriptionEdited) {
+      this.handleBlur();
+    }
```
adamklepacz: please extract textarea field to separate component to avoid such things in component Did update

You will have 'focus() in componentDidMount()

alemikolo: In my opinion, both solutions have their advantages and disadvantages. I tried to implement your solution, but to make it work exactly the same, I would need more time. I think it's better now to concentrate on the last task in the current sprint than to spend time getting the same effect in a different way (the current solution works correctly).

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,275 @@
...
+      isFocused: false,
+      pending: false
+    };
+
+    this.descriptionTextarea = React.createRef();
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    this.isDescriptionUpdated();
+    const { isTextareaVisible: previousIsTextareaVisible } = prevState;
```
adamklepacz: what does it mean ?
previous Is Textarea Visible? poprzednia jest textarea widoczna?

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,275 @@
...
+      this.pendingPromise = makeAbortablePromise(
+        updateListItem(listId, itemId, { description }, name)
+      );
+      this.pendingPromise.promise
+        .then(() => {
+          this.setState({ pending: false });
+          this.handleHideTextarea();
+        })
+        .catch(err => {
+          if (!(err instanceof AbortPromiseException)) {
```
adamklepacz: will it ever go inside this if??

If the error of instance AbortPromiseException occurs, so the condition will be true, and there is a negation before - it will be evaluated finally to false 
the code inside will not be executed.

but if the will not be the instance of Abort Promise Exception it will go inside and set pending to false,

can't we just this.setState({pending: false}) always, independently if the error is AbortPromiseException or default error??

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,275 @@
...
+            this.setState({ pending: false });
+          }
+        });
+    } else {
```
adamklepacz: else not needed
just add return inside then

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,275 @@
...
+        onClick={this.handleShowTextarea}
+        type="button"
+      >
+        Add Description
+      </button>
+    );
+  };
+
+  renderTextarea = () => {
+    const { descriptionInputValue, pending } = this.state;
```
adamklepacz: descriptionTextareaValue

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,275 @@
...
+    const { description, disabled } = this.props;
+
+    const isSaveButtonVisible = isDescriptionUpdated && !disabled;
+    const isTitleVisible = description || isTextareaVisible;
+
+    return (
+      <Fragment>
+        <div onClick={disabled ? null : this.handleShowTextarea}>
+          {isTitleVisible && (
+            <h2 className="list-item-description__title">Description</h2>
```
adamklepacz: I think the title should be always visible.
After adding the description it's not clear what it is

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,275 @@
...
+      </Linkify>
+    );
+  };
+
+  renderAddButton = () => {
+    const { disabled } = this.props;
+
+    return (
+      <button
+        className="list-item-description__add-button"
```
adamklepacz: add class link-button to match other buttons around this area
and remove left padding

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
alemikolo: Yes, it will be executed. When the component is unmounted the AbortPromiseException is thrown, then this code inside this if statement won't be executed, and it shouldn't be executed. That's why this check was created for. To prevent calling setState when the component is unmounted. In all other cases (errors) this setState will be executed.
So If a component is unmounted, AbortPromiseException is thrown, and this check prevents calling setState on unmounting component

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
alemikolo: Hmm... The title is displayed always when there is a description or the textarea is visible. It is hidden only when there is no description and add-button is visible. So I think it's clear enough...

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,274 @@
...
+
+    const { description } = this.props;
+    const trimmedDescription = description.trim();
+
+    this.state = {
+      isDescriptionUpdated: false,
+      descriptionTextareaValue: trimmedDescription,
+      isTextareaVisible: false,
+      isFocused: false,
+      pending: false
```
adamklepacz: sort

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,274 @@
...
+    }
+  }
+
+  componentWillUnmount() {
+    if (this.pendingPromise) {
+      this.pendingPromise.abort();
+    }
+  }
+
+  handleKeyPress = event => {
```
adamklepacz: handleEscapePress

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -0,0 +1,274 @@
...
+
+  renderDescription = () => {
+    const { description, disabled } = this.props;
+    const { pending } = this.state;
+
+    return (
+      <Linkify
+        properties={{
+          target: '_blank',
+          className: 'list-item-description__link'
```
adamklepacz: sort props

PR #171, opened 2019-06-05
[BUG] Fix favourites

adamklepacz:


adamklepacz:
@marekrozmus there are already.
this function is moved from down the file.

Zrzut ekranu 2019-06-05 o 12 08 43

PR #172, opened 2019-06-05
[EOC-197] Add tooltip to fav button

adamklepacz:


client/src/common/components/ListViewItem.jsx
```diff
@@ -71,6 +71,7 @@ class ListViewItem extends PureComponent {
             })}
             disabled={pending}
             onClick={this.handleFavClick}
+            title="Add to favourites"
```
alemikolo: maybe change to sth like this title={`$ {isFavourite ? 'Add to' : 'Remove from`} favourites`}

PR #175, opened 2019-06-10
[EOC-209] activity adding item to list proof of concept

alemikolo:
Users can see activities only from list/cohorts they belong to.

Activity Schema:

- actorId - id of the user who performed an action
- activityType - string to recognize which message's template should be displayed
- cohortId, listId, itemId - ids for action's subjects, they can be null, frontend will use actionType to choose the specific message and fill it with needed and specific variables.
- editedUserId - id of the user who was the target of the action (e.g. invited/removed from list/cohort)
- editedValue - field to store e.g. the previous name of the edited element or name of the deleted element (can't be populated any more)

screencapture-localhost-3000-dashboard-2019-06-10-09_17_08
screencapture-localhost-3000-dashboard-2019-06-10-09_24_28
download


client/src/common/components/Activities/components/activity.jsx
```diff
@@ -0,0 +1,110 @@
...
+        createdAt,
+        item: { itemName }
+      }
+    } = this.props;
+    const { isAvatarError } = this.state;
+    const date = dateFromString(createdAt);
+
+    return (
+      <div className="activity">
+        <div
```
marekrozmus: could this div be extracted to separate `Avatar` component with all error handling functionality?

client/src/common/components/Activities/model/actionTypes.js
```diff
@@ -0,0 +1,4 @@
+export const ActivityActionTypes = Object.freeze({
+  FETCH_SUCCESS: 'activity/FETCH_SUCCESS',
+  FETCH_FAILURE: 'activity/FETCH_FAILURE'
```
marekrozmus: shouldn't we have plural here - `activities`? This fetches all activities, not only one?

client/src/common/components/Activities/model/reducer.js
```diff
@@ -0,0 +1,15 @@
+import { ActivityActionTypes } from './actionTypes';
+
+const activities = (state = [], action) => {
+  switch (action.type) {
+    case ActivityActionTypes.FETCH_SUCCESS: {
+      const { payload } = action;
+
+      return [...payload];
```
marekrozmus: you don't need new object here so `payload` is enough

client/src/common/components/Dropdown.jsx
```diff
@@ -0,0 +1,102 @@
...
+        isVisible: false
+      });
+    }
+  };
+
+  handleClickOutside = event => {
+    const isClickedOutside = !this.dropdown.current.contains(event.target);
+
+    if (isClickedOutside) {
+      this.setState({ isVisible: false });
```
marekrozmus: Please be consist about the new lines - here it is one liner and in line 39 you've got 3 lines.
Please use one line until prettier says otherwise.

client/src/common/components/Members/components/MemberDetails.jsx
```diff
@@ -310,13 +301,7 @@ class MemberDetails extends PureComponent {
...
     return (
       <header className="member-details__header">
-        <div
-          className={classNames('member-details__avatar', {
-            'member-details__avatar--error': isAvatarError
-          })}
-        >
-          {this.renderAvatar()}
-        </div>
+        <div className="member-details__avatar">{this.renderAvatar()}</div>
```
marekrozmus: remove the `renderAvatar` method and use here `<Avatar />` component

client/src/common/components/Activities/components/ActivitiesList.jsx
```diff
@@ -0,0 +1,84 @@
...
+
+  render() {
+    const { activities } = this.props;
+    const { pending } = this.state;
+
+    return (
+      <div className="activities-list">
+        <h2 className="activities-list__heading">Activities</h2>
+        <div className="activities-list__content">
+          {activities.length ? (
```
adamklepacz: wouldn't be `activities ?` enough?

client/src/common/components/Activities/components/Activity.jsx
```diff
@@ -0,0 +1,92 @@
...
+import Avatar from 'common/components/Avatar';
+
+class Activity extends PureComponent {
+  renderCohortInfo = () => {
+    const {
+      activity: { cohort }
+    } = this.props;
+
+    if (cohort) {
+      const { cohortId, cohortName } = cohort;
```
adamklepacz: break line before return

client/src/common/components/Activities/components/Activity.jsx
```diff
@@ -0,0 +1,92 @@
...
+    }
+  };
+
+  renderListInfo = () => {
+    const {
+      activity: { list }
+    } = this.props;
+
+    if (list) {
+      const { listId, listName } = list;
```
adamklepacz: new line before return

client/src/common/components/Activities/components/Activity.jsx
```diff
@@ -0,0 +1,92 @@
...
+          </Link>
+        </Fragment>
+      );
+    }
+  };
+
+  render() {
+    const {
+      activity: {
+        actor: { actorAvatarUrl, actorName },
```
adamklepacz: why actor?

client/src/common/components/Comments/Comment.jsx
```diff
@@ -1,42 +1,24 @@
...
-              onError={this.handleError}
-              src={isAvatarError ? UserIconPlaceholder : authorAvatarUrl}
-            />
-          ) : (
-            <UserIcon />
-          )}
+        <div className="comment__avatar">
+          <Avatar
+            avatarUrl={authorAvatarUrl}
+            className="activity__image"
```
adamklepacz: wrong class name.
According to BEM children can be nested only within the parent.
Where is `.activity` parent?

This should be raather independent Block element, so `.activity-avatar` or something

server/controllers/activity.js
```diff
@@ -0,0 +1,124 @@
...
+          editedValue,
+          editedUser,
+          item,
+          list
+        };
+      });
+
+      resp.send(activities);
+    })
+    .catch(err => resp.sendStatus(400));
```
adamklepacz: unused err here

client/src/common/components/Activities/components/ActivitiesList.jsx
alemikolo: Unfortunately no. This is an array of activities, so we have to check this array's length because if there are no activities, an empty array is returned. We check items array in the same way (ItemsList.jsx, line:45) but we can also check _empty(activities).

client/src/common/components/Activities/components/Activity.jsx
alemikolo: I found this naming convention in one solution regarding activities in bank application. An actor is a user who performed an action. Of course, it could be a simple name: user. I used "actor" to distinguish it from the second user we can have in activity data: eg. a user who was edited. If you think it's strange and unclear I can't change it

adamklepacz: why not just 
`performer` ? 
`performerAvatrUrl `,  `performerName`

client/src/common/components/Activities/model/actions.js
```diff
@@ -0,0 +1,26 @@
...
+
+const fetchActivitiesFailure = () => ({
+  type: ActivitiesActionTypes.FETCH_FAILURE
+});
+
+export const fetchActivities = () => dispatch =>
+  getData('/api/activities/data')
+    .then(resp => resp.json())
+    .then(json => dispatch(fetchActivitiesSuccess(json)))
+    .catch(err => {
```
adamklepacz: err not used

client/src/common/components/Activities/model/selectors.js
```diff
@@ -0,0 +1,10 @@
+import _orderBy from 'lodash/orderBy';
+
+export const getActivities = state => {
+  const { activities } = state;
```
adamklepacz: add break line

PR #177, opened 2019-06-11
[EOC-215] BUG User invitation panel appears after sending empty input

alemikolo:


client/src/common/utils/fetchMethods.js
```diff
@@ -19,7 +19,21 @@ const handleFetchErrors = resp => {
...
 
-  if (resp.status >= ResponseStatusCode.BAD_REQUEST && resp.status < 600) {
+  if (resp.status === ResponseStatusCode.BAD_REQUEST) {
+    const contentType = resp.headers.get('content-type');
+    if (contentType.includes('application/json')) {
+      return resp.json().then(json => {
+        if (json.message) {
+          throw new Error(json.message);
+        }
+        throw new Error();
```
marekrozmus: make one line with ternary operator

client/src/common/components/Members/components/MembersForm.jsx
```diff
@@ -81,7 +86,7 @@ class MembersForm extends PureComponent {
           className={classNames('primary-button', {
             'primary-button--disabled': disabled
           })}
-          disabled={disabled}
+          disabled={disabled || pending || isEmpty}
```
adamklepacz: create const above

`const isButtonDisabled = disabled || pending || isEmpty; `

and use it here

PR #178, opened 2019-06-11
[EOC-218] BUG Limited text flashes when entering private list

alemikolo:


client/src/modules/list/index.jsx
```diff
@@ -133,7 +133,7 @@ class List extends Component {
     }
 
     const { cohortId, isArchived, isMember, isOwner, name, type } = list;
-    const isCohortList = cohortId !== null;
+    const isCohortList = cohortId !== null && cohortId !== undefined;
```
adamklepacz: you can do it just 
`const isCohortList = !cohortId;`

If cohort id will be undefined or null you'll get true. If cohort id will be some data, you will get false.

adamklepacz: Short example:
Zrzut ekranu 2019-06-11 o 15 32 42

alemikolo: I should do `const isCohortList = cohortId`, but then I receive an  error that cohortID should be a boolean

PR #185, opened 2019-06-18
[EOC-227] Socket adding items to sack

adamklepacz:


client/src/modules/list/components/Items/InputBar/InputBar.jsx
```diff
@@ -36,6 +39,10 @@ class InputBar extends Component {
     if (isTipVisible) {
       this.hideTipAfterTimeout();
     }
+
+    if (!this.socket) {
+      this.socket = io(FULL_URL);
```
alemikolo: why you delete namespace '/sack'? Can we use rooms without namespaces?

server/sockets/list.js
```diff
@@ -0,0 +1,24 @@
+const { ItemActionTypes } = require('../common/variables');
+
+/* WS postfix stands for Web Socket, to differentiate
+ * this from controllers naming convention
+ */
+const addItemToListWS = socket => {
+  socket.on('listRoom', room => {
```
alemikolo: 1. Could we move joining and leaving rooms listeners to socket index, and not repeat this in every specific event (addItemToListWS)?
```
 ioInstance.on('connection', socket => {
    const {
      request: { user }
    } = socket;
    const userExist = user || false;

     if (!userExist) {
      return;
    }

socket.on('listRoom', room => {
ocket.join(room);
  });
socket.on('leavingRoom', listId => {
    socket.leave(`list-${listId}`);
  });

    addItemToListWS(socket);
  });
```
2. If we have leaveRoom event it'd be good to have joinRoom event not listRoom or joinListRoom, leaveListRoom

client/src/modules/list/components/Items/InputBar/InputBar.jsx
adamklepacz: yes we can.
You said yesterday that they're attached to global namespace

PR #186, opened 2019-06-20
[EOC-238] Reset password ui form

adamklepacz:
Zrzut ekranu 2019-06-20 o 15 16 50
Zrzut ekranu 2019-06-20 o 15 17 01


client/src/locales/en.json
```diff
@@ -68,6 +68,10 @@
   "authorization.auth-box.sign-in": "Sign in with:",
   "authorization.auth-box.cta": "Try it out:",
   "authorization.auth-box.demo-button": "Live demo",
+  "authorization.reset-password.heading": "Reset your password",
+  "authorization.reset-password.email-label": "Enter email",
+  "authorization.reset-password.tip-content": "Please enter correct email",
```
marekrozmus: Please read this article:
https://uxplanet.org/error-message-guidelines-6ce257d3d0bd

Maybe text: "Please enter a valid emila address" would be nicer :)

client/src/modules/authorization/components/ResetPassword.jsx
```diff
@@ -0,0 +1,79 @@
...
+      // handle submit here
+      return;
+    }
+
+    this.handleTipVisibility();
+  };
+
+  handleTipVisibility = () => {
+    this.setState({ tipVisible: true });
+    setTimeout(() => this.setState({ tipVisible: false }), 6000);
```
marekrozmus: do not hide it with setTimeout - just leave it and recheck when form is resubmitted

client/src/modules/authorization/components/ResetPassword.jsx
```diff
@@ -0,0 +1,79 @@
...
+              type="email"
+              value={email}
+            />
+          </label>
+          {tipVisible && (
+            <span className="reset-password__message-error">
+              <FormattedMessage id="authorization.reset-password.tip-content" />
+            </span>
+          )}
+          <button
```
alemikolo: Consider: disable button if the field is empty or its value is invalid.

adamklepacz: Only for an empty field.
When it filled even with a wrong value user should be able to press the button to get the TIP with the information about the mistake he made. Otherwise, the user will have no feedback, what he did wrong.

client/src/modules/authorization/components/ResetPassword.jsx
```diff
@@ -0,0 +1,82 @@
...
+    const isEmailCorrect = this.handleEmailValidation();
+
+    if (isEmailCorrect) {
+      // handle submit here
+      this.hideTip();
+
+      return;
+    }
+
+    this.showTip();
```
marekrozmus: Please reverse the if. Check if not ok then just show tip and return. Handling submittion would have probably more than one line.
```
if (!isEmailCorrect) {
  this.showTip();
  return;
}
...
```

client/src/modules/authorization/components/ResetPassword.jsx
```diff
@@ -0,0 +1,82 @@
...
+
+  handleEmailValidation = () => {
+    const { email } = this.state;
+
+    return validator.isEmail(email);
+  };
+
+  handleSubmit = event => {
+    event.preventDefault();
+    const isEmailCorrect = this.handleEmailValidation();
```
marekrozmus: do you need this `handleEmailValidation` method? It is used only once. You can just use validator here.

PR #187, opened 2019-06-20
[EOC-242] Leaving cohort

adamklepacz:
Zrzut ekranu 2019-06-20 o 18 19 09
Zrzut ekranu 2019-06-20 o 18 19 16
Zrzut ekranu 2019-06-20 o 18 19 48


client/src/common/components/Members/components/MemberDetails.jsx
```diff
@@ -358,6 +383,59 @@ class MemberDetails extends PureComponent {
...
+  renderLeaveOption = () => {
+    const {
+      _id: userId,
+      currentUser: { id: currentUserId },
+      onClose,
+      route
+    } = this.props;
+    const { isLeaveConfirmationVisible, pendingForLeaving } = this.state;
+
+    if (userId === currentUserId) {
```
marekrozmus: make an early return instead of putting all this code in if block.
```
if (userId !== currentUserId) {
    return null;
}

return (
   <div...
```

server/controllers/cohort.js
```diff
@@ -488,15 +488,66 @@ const addMember = (req, resp) => {
...
+        ownerIds.splice(doc.ownerIds.indexOf(userId), 1);
+      }
+
+      if (isMember) {
+        memberIds.splice(doc.memberIds.indexOf(userId), 1);
+      }
+
+      return doc.save();
+    })
+    .then(() => {
```
marekrozmus: remove braces

client/src/modules/cohort/model/reducer.js
```diff
@@ -47,6 +47,14 @@ const membersReducer = (state = [], action) => {
           : member
       );
     }
+    case CohortActionTypes.LEAVE_SUCCESS: {
+      const {
+        payload: { userId }
+      } = action;
+      const userIndex = state.indexOf(userId);
+
+      return [...state.slice(0, userIndex), ...state.slice(userIndex + 1)];
```
alemikolo: only proposal: maybe it's time to change members to an object?

client/src/common/components/Members/components/MemberDetails.jsx
```diff
@@ -358,6 +383,59 @@ class MemberDetails extends PureComponent {
...
+      return (
+        <div className="member-details__leave-box">
+          {isLeaveConfirmationVisible ? (
+            <div className="member-details__leave-box-confirmation">
+              <FormattedMessage id="common.member-details.leave-box-heading" />
+              <footer className="member-details__leave-box-confirmation-footer">
+                <PendingButton
+                  className="primary-button"
+                  onClick={this.handleCohortLeave}
+                  pending={pendingForLeaving}
```
alemikolo: is pendingForLeaving necessary? It is used only here but the pending button doesn't wait for "pending" prop. It has own pending which is initialized inside it.

client/src/modules/cohort/model/actionTypes.js
```diff
@@ -2,12 +2,12 @@ export const CohortActionTypes = Object.freeze({
...
   ADD_MEMBER_REQUEST: 'cohort/ADD_MEMBER_REQUEST',
   ADD_MEMBER_SUCCESS: 'cohort/ADD_MEMBER_SUCCESS',
-  ARCHIVE_FAILURE: 'cohort/ARCHIVE_FAILURE',
-  ARCHIVE_REQUEST: 'cohort/ARCHIVE_REQUEST',
-  ARCHIVE_SUCCESS: 'cohort/ARCHIVE_SUCCESS',
   ADD_OWNER_ROLE_FAILURE: 'cohort/ADD_OWNER_ROLE_FAILURE',
   ADD_OWNER_ROLE_REQUEST: 'cohort/ADD_OWNER_ROLE_REQUEST',
   ADD_OWNER_ROLE_SUCCESS: 'cohort/ADD_OWNER_ROLE_SUCCESS',
+  ARCHIVE_FAILURE: 'cohort/ARCHIVE_FAILURE',
+  ARCHIVE_REQUEST: 'cohort/ARCHIVE_REQUEST',
```
alemikolo: Consider: You only change the order here, but if you edited those lines maybe it's good to removed unused request actions because they are not needed now?

adamklepacz:
> "Please try again" in error message in this case does not help. Error should help to solve the situation.
> The success info should rather be like: "You have left [cohortName] cohort".|

@marekrozmus 
but how if this is always added for NotificationType.Error inside Notification component??

client/src/common/components/Members/components/MemberDetails.jsx
```diff
@@ -192,6 +194,27 @@ class MemberDetails extends PureComponent {
...
+      currentUser: { id: currentUserId, name },
+      leaveCohort,
+      match: {
+        params: { id }
+      }
+    } = this.props;
+
+    this.setState({ pending: true });
+
+    return leaveCohort(id, currentUserId, name).finally(() => {
```
marekrozmus: why braces after => ?

client/src/locales/en.json
```diff
@@ -84,6 +86,9 @@
   "cohort.index.hide-archived": "Hide archived sacks",
   "cohort.index.show-archived": "Show archived sacks",
   "cohort.index.no-archived-sacks": "There are no archived sacks in the {name} cohort!",
+  "cohort.actions.leave-cohort": "'{data}' successfully leaved cohort.",
```
marekrozmus: there is no such word like `leaved` - should be `left'

client/src/locales/en.json
```diff
@@ -84,6 +86,9 @@
...
+  "cohort.actions.leave-cohort-fail": "'Failed to leave cohort for user: '{data}'.",
```
marekrozmus: why ` at the beginning of string?

server/controllers/cohort.js
```diff
@@ -488,15 +488,64 @@ const addMember = (req, resp) => {
...
+    .then(doc => {
+      if (!doc) {
+        throw new BadRequestException();
+      }
+
+      const { memberIds, ownerIds } = doc;
+      const isOwner = checkIfArrayContainsUserId(ownerIds, userId);
+      const isMember = checkIfArrayContainsUserId(memberIds, userId);
+
+      if (ownerIds.length === 1 && isOwner) {
```
marekrozmus: 1. first check `isOwner` - if not then the length is not important for us.
2. merge two if into one
```
if (isOwner) {
    if (ownerIds.length === 1) {
        throw new ...
    }

    ownerIds.splice...
}

File: server/controllers/cohort.js [hunk 10]
```diff
@@ -488,15 +488,64 @@ const addMember = (req, resp) => {
...
+        throw new BadRequestException(
+          'cohort.actions.leave-cohort-only-one-owner'
+        );
+      }
+
+      if (isOwner) {
+        ownerIds.splice(doc.ownerIds.indexOf(userId), 1);
+      }
+
+      if (isMember) {
```
marekrozmus: If you are owner then you still can be member? If not then you could use just `else if` block.

marekrozmus:
> > "Please try again" in error message in this case does not help. Error should help to solve the situation.
> > The success info should rather be like: "You have left [cohortName] cohort".|
> 
> @marekrozmus
> but how if this is always added for NotificationType.Error inside Notification component??



> > "Please try again" in error message in this case does not help. Error should help to solve the situation.
> > The success info should rather be like: "You have left [cohortName] cohort".|
> 
> @marekrozmus
> but how if this is always added for NotificationType.Error inside Notification component??

Just first thing off top of my head - create other type like NotificationType.ERROR_TRY_AGAIN and use it accordingly.

server/controllers/cohort.js
adamklepacz: Every higher role inherits from the previous one, so even when I'm an owner I'm still a member.

server/controllers/cohort.js
```diff
@@ -488,15 +488,63 @@ const addMember = (req, resp) => {
...
+      const isOwner = checkIfArrayContainsUserId(ownerIds, userId);
+      const isMember = checkIfArrayContainsUserId(memberIds, userId);
+
+      if (isOwner) {
+        if (ownerIds.length === 1) {
+          throw new BadRequestException(
+            'cohort.actions.leave-cohort-only-one-owner'
+          );
+        }
+        ownerIds.splice(doc.ownerIds.indexOf(userId), 1);
```
marekrozmus: new line after block end

PR #189, opened 2019-06-24
[EOC-228] Sockets deleting items from sacks

adamklepacz:
When the item is archived, cloned, deleted or restored on that particular moment and the other user wants to perform other actions, the item is blocked by a transparent overlay. When he clicks this transparent overlay - he gets a darker overlay with a message that someone is editing this item at the moment so he can't perform anything.

Whenever the item is archived, cloned, deleted or restored it gets busyBySomeone state property to true :) 

Zrzut ekranu 2019-06-24 o 19 43 52

**_ATENCION:_** 
**This item has a transparent overlay. Here on the screen its gray only for the purpose of this screenshot.**
Zrzut ekranu 2019-06-24 o 19 44 04
Zrzut ekranu 2019-06-24 o 19 44 11


server/sockets/index.js
```diff
@@ -9,7 +9,13 @@ const cookieParser = require('cookie-parser');
 const sessionStore = new MongoStore({
   mongooseConnection: mongoose.connection
 });
-const { addItemToListWS } = require('./list');
+const {
+  addItemToListWS,
+  archiveItemWs,
```
marekrozmus: small `s` - everywhere else it is capital

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -146,9 +247,19 @@ class ListItem extends PureComponent {
...
+  handleNameFocus = () => this.setState({ isNameEdited: true, busy: true });
+
+  handleNameBlur = () => this.setState({ isNameEdited: false, busy: false });
 
-  handleNameBlur = () => this.setState({ isNameEdited: false });
+  handleDescriptionFocus = () => this.setState({ busy: true });
+
+  handleDescriptionBlur = () => this.setState({ busy: false });
+
+  handleCommentsBlur = () => this.setState({ busy: false });
```
alemikolo: I'm not sure if comments should be blocked, they are save in separate collection so I think it won't cause errors. Writing comment can take a long time so for all this time item will be locked. It will be very uncomfortable and annoying. Voting also, but voting is a very quick action so it won't cause big discomfort. But if you think about the list with several dozen users and only one of them can add a comment...

PR #190, opened 2019-06-24
[EOC-231] Create backend for signing up

alemikolo:
Screenshot 2019-06-24_23-13-21-986
Screenshot 2019-06-24_23-15-52-258


server/controllers/userAuth.js
```diff
@@ -24,8 +33,102 @@ const sendDemoUser = (req, resp) => {
...
+  if (isEmpty(sanitizedEmail) || !isEmail(sanitizedEmail)) {
+    errors.emailError = 'authorization.input.email.invalid';
+  }
+
+  if (!matches(sanitizedPassword, /^[^\s]{4,32}$/)) {
+    errors.passwordError = 'authorization.input.password.invalid';
+  }
+
+  if (sanitizedPassword !== sanitizedPasswordConfirm) {
+    errors.passwordConfirmError = 'authorization.input.password.not-match';
```
marekrozmus: are these localization ids? If yes than this shoud not be this way - in case we want to change the id on client we will need to update backend. This should be some string error which is mapped on client to specific localization id.

server/controllers/userAuth.js
```diff
@@ -24,8 +33,102 @@ const sendDemoUser = (req, resp) => {
...
+  }
+
+  if (_find(errors, error => error !== undefined)) {
```
marekrozmus: use `_some` so that return value is boolean

server/controllers/userAuth.js
```diff
@@ -24,8 +33,102 @@ const sendDemoUser = (req, resp) => {
...
+              }
+
+              return { displayName, email, signUpHash };
+            });
+        }
+
+        throw new BadRequestException('authorization.user-already-exist');
+      } else {
+        const salt = bcrypt.genSaltSync(12);
+        const hashedPassword = bcrypt.hashSync(password + email, salt);
```
marekrozmus: you don't need to generate salts with bcrypt (it is part of hash) just use:
`bcrypt.hash(password, 12)` and then `bcrypt.compare(logInData.password, user.password)`

server/controllers/userAuth.js
```diff
@@ -24,8 +33,102 @@ const sendDemoUser = (req, resp) => {
   resp.redirect('/');
 };
 
+const signUp = (req, resp, next) => {
+  const { email, password, passwordConfirm, username } = req.body;
+  const sanitizedEmail = sanitize(email);
+  const sanitizedPassword = sanitize(password);
+  const sanitizedPasswordConfirm = sanitize(passwordConfirm);
```
adamklepacz: sanitizing is removing $ sing from strings, what if the user wants to have 'qwerty12345$' password?
Are we informing user on the front that $ is not allowed?

server/controllers/userAuth.js
```diff
@@ -24,8 +33,102 @@ const sendDemoUser = (req, resp) => {
...
+    .exec()
+    .then(user => {
+      if (user) {
+        const { _id, displayName, email, idFromProvider, isActive } = user;
+
+        if (!idFromProvider && !isActive) {
+          const signUpHash = crypto.randomBytes(32).toString('hex');
+          const expirationDate = new Date().getTime() + 3600000;
+
+          return User.findOneAndUpdate(
```
adamklepacz: this User that you're looking for here is the same user that you allready get in the 69 line.

Instead of calling once again findOneAndUpdate, can't you just do user.save(); ??
**It just a question and proposal**

alemikolo: if we do something like this:
```
user.signUpHash = crypto.randomBytes(32).toString('hex');
return user.save();
```
we get eslint error: Assignment to property of function parameter 'user'.eslint(no-param-reassign)

server/controllers/userAuth.js
```diff
@@ -24,8 +33,98 @@ const sendDemoUser = (req, resp) => {
...
+              if (!user) {
+                throw new Error();
+              }
+
+              return { displayName, email, signUpHash };
+            });
+        }
+
+        throw new BadRequestException('authorization.user-already-exist');
+      } else {
```
adamklepacz: else not needed here

PR #192, opened 2019-06-25
Hotifx sockets

adamklepacz:
Web Sockets were not working because of a bad merge inside 
client/src/modules/list/index.jsx   file.

I have found "React can't call setState after a component is unmounted..." and fix is inside ListItem.jsx file.

### DIFF
--- a/client/src/modules/list/components/Items/ListItem/ListItem.jsx
+++ b/client/src/modules/list/components/Items/ListItem/ListItem.jsx
@@ -30,8 +30,6 @@ import ListItemDescription from '../ListItemDescription';
 import { ItemStatusType } from '../model/actionTypes';
 
 class ListItem extends PureComponent {
-  isMounted = false;
-
   constructor(props) {
     super(props);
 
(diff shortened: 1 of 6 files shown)


client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -30,6 +30,8 @@ import ListItemDescription from '../ListItemDescription';
 import { ItemStatusType } from '../model/actionTypes';
 
 class ListItem extends PureComponent {
+  isMounted = false;
```
marekrozmus: that is huge antipattern and we should avoid it: https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html

Try to move the sockets up - to the list itself, not to every item and create bool property for list item like `blocked` and set it from list.

This should be simple component and now it is infected with sockets ;)

adamklepacz:
@marekrozmus 
 I removed socket emitting and receiving logic from ListItem to ItemsList component.
And now, no error related with React can't call set state appears in the console.
Everything works as expected. 

Anyway, I can't totally remove socket connection from ListItem component, cause this item need to pass socket as an argument to all actions related to archiving, deleting, restoring, etc. 
Because inside the action I'm emitting action about "ARCHIVE_ITEM" "RESTORE_ITEM" etc

PR #194, opened 2019-06-26
[EOC-234] Resend activation link if already registered

alemikolo:
Almost all functionality has been done in EOC-231 task. I added updating displayName and password in signup controller for already registered users in the situation that the user tries signup with the same email but uses other name or password

### DIFF
--- a/server/controllers/userAuth.js
+++ b/server/controllers/userAuth.js
@@ -67,12 +67,18 @@ const signUp = (req, resp, next) => {
         const { _id, displayName, email, idFromProvider, isActive } = user;
 
         if (!idFromProvider && !isActive) {
+          const hashedPassword = bcrypt.hashSync(password + email, 12);
           const signUpHash = crypto.randomBytes(32).toString('hex');
           const expirationDate = new Date().getTime() + 3600000;
 
           return User.findOneAndUpdate(
             { _id },
-            { signUpHash, signUpHashExpirationDate: expirationDate }
+            {
+              displayName: sanitizedUsername,
+              password: hashedPassword,
+              signUpHash,
+              signUpHashExpirationDate: expirationDate
+            }
           )
             .exec()
             .then(user => {


server/controllers/userAuth.js
```diff
@@ -67,12 +67,18 @@ const signUp = (req, resp, next) => {
...
         if (!idFromProvider && !isActive) {
+          const hashedPassword = bcrypt.hashSync(password + email, 12);
           const signUpHash = crypto.randomBytes(32).toString('hex');
           const expirationDate = new Date().getTime() + 3600000;
 
           return User.findOneAndUpdate(
             { _id },
-            { signUpHash, signUpHashExpirationDate: expirationDate }
+            {
+              displayName: sanitizedUsername,
```
marekrozmus: shouldn't it find by email? Email is unique not the displayName - we cannot update email address we should use it to serach for existing account.

alemikolo: When a user tries to sign up I find if there is a user of the given email (line 63). If there is a user and if their account is inactive (and not provided by Google) I use  `_id` of this currently found user (line 75) to update their document - lines 77-80. I update only displayName, password, signUpHash and signUpHashExpirationDate, I have to update the signUpHash, and I decided to update displayName and password in this case that user would have the last data their entered, but email stays the same,

```
User.findOne({ email: sanitizedEmail })
    .exec()
    .then(user => {
      if (user) {
        const { _id, displayName, email, idFromProvider, isActive } = user;

        if (!idFromProvider && !isActive) {
          const hashedPassword = bcrypt.hashSync(password + email, 12);
          const signUpHash = crypto.randomBytes(32).toString('hex');
          const expirationDate = new Date().getTime() + 3600000;

          return User.findOneAndUpdate(
            { _id },
            {
              displayName: sanitizedUsername,
              password: hashedPassword,
              signUpHash,
              signUpHashExpirationDate: expirationDate
            }
          )
```

PR #199, opened 2019-06-27
[EOC-236] Update login form page to enter credentials

alemikolo:
Password is not validated, password input becomes green if the user enters something.

Screenshot 2019-06-27_11-51-58-670
Screenshot 2019-07-01_13-01-58-632
Screenshot 2019-06-27_12-06-51-173
Screenshot 2019-06-27_12-06-35-965


client/src/modules/authorization/components/SignInForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+
+  componentWillUnmount() {
+    if (this.pendingPromise) {
+      this.pendingPromise.abort();
+    }
+  }
+
+  onEmailChange = (email, isValid) => {
+    const { signUpErrorId } = this.state;
+    const error = isValid ? '' : signUpErrorId;
```
adamklepacz: add some air before const declaration

client/src/modules/authorization/components/SignInForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+    this.setState({
+      email,
+      isEmailValid: isValid,
+      signUpErrorId: error
+    });
+  };
+
+  onPasswordChange = (password, isValid) => {
+    const { signUpErrorId } = this.state;
+    const error = isValid ? '' : signUpErrorId;
```
adamklepacz: add some air before const declaration

client/src/modules/authorization/components/SignInForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+      isFormValid: isEmailValid && isPasswordValid && !signUpErrorId
+    });
+  };
+
+  handleSignIn = () => {
+    const { email, password } = this.state;
+    const { signIn } = this.props;
+
+    this.setState({ pending: true });
+
```
adamklepacz: remove this line

client/src/modules/authorization/components/SignInForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+    this.pendingPromise = makeAbortablePromise(signIn(email, password));
+
+    return this.pendingPromise.promise
+      .then(() => this.setState({ pending: true }))
```
adamklepacz: this line is not necessary if you set pending status to true in line 84

client/src/modules/authorization/components/SignInForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+      .catch(err => {
+        if (!(err instanceof AbortPromiseException)) {
+          const newState = { pending: false };
+
+          if (err instanceof UnauthorizedException) {
+            newState.signUpErrorId =
+              'authorization.actions.sign-in.invalid-credentials';
+          } else {
```
adamklepacz: else not needed here

client/src/modules/authorization/components/SignInForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+      id: 'common.try-again'
+    })}`;
+
+    return <p className="sign-in-form__error">{message}</p>;
+  };
+
+  render() {
+    const { isFormValid, pending, signUpErrorId } = this.state;
+    const { onCancel } = this.props;
+    const hasSignUpFailed = signUpErrorId !== '';
```
adamklepacz: you can also check the length 

const hasSignUpFailed = signUpErrorId.length > 0

client/src/modules/authorization/components/SignInForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+
+    return (
+      <div className="sign-in">
+        <h1 className="sign-in-form__heading">
```
adamklepacz: should be `sign-in__heading`

client/src/modules/authorization/components/SignInForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+          <FormattedMessage id="authorization.auth-box.sign-in" />
+        </h1>
+        {signUpErrorId && this.renderSignUpError()}
+        <form className="sign-in-form__form" noValidate>
```
adamklepacz: sign-in__form

client/src/modules/authorization/components/SignInForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+            <button
+              className="primary-button"
+              disabled={pending}
+              onClick={onCancel}
+              type="button"
+            >
+              <FormattedMessage id="common.button.cancel" />
+            </button>
+            <PendingButton
+              className="primary-button sign-in-form__confirm"
```
adamklepacz: sign-in__confirm etc...

client/src/locales/en.json
```diff
@@ -87,7 +87,9 @@
   "authorization.input.password.confirm": "Confirm password",
   "authorization.input.password.invalid": "Please use min 4 and max 32 signs and don't use space characters.",
   "authorization.input.password.not-match": "Passwords do not match.",
+  "authorization.actions.sign-up.confirmation-link-sent": "In order to complete sign up process please click the link that we have just sent to {data}.",
   "authorization.actions.sign-up.user-already-exist": "User of email {data} already exist.",
+  "authorization.actions.sign-in.invalid-credentials": "The email and the password you entered didn't match our records.",
```
marekrozmus: maybe some better text like the google have: 'Wrong password. Try again or click Forgot password to reset it.'

client/src/modules/authorization/components/SignInForm.jsx
alemikolo: `else` is needed because there is no return in `if` statement. If I remove `else` signInErrorId would always be 'common.something-went-wrong'

adamklepacz: so you can add 'return' at the end

alemikolo: 1. If we add return in if statement we will get the eslint error "Return statement should not contain assignment."
2. Return inside this if statement will exit catch statement

client/src/modules/authorization/components/SignInForm.jsx
```diff
@@ -0,0 +1,206 @@
...
+            >
+              <FormattedMessage id="authorization.sign-in" />
+            </PendingButton>
+          </div>
+        </form>
+        <p className="sign-in__forgot-password">
+          <Link className="sign-in__link" to="/reset-password">
+            {formatMessage({ id: 'authorization.forgot-password' })}
+          </Link>
+          ?
```
marekrozmus: do not create questions like this. It might work in this language but fail in others. Create separate strings for the question and forgot password link.

client/src/common/utils/fetchMethods.js
```diff
@@ -33,6 +35,18 @@ const handleFetchErrors = resp => {
...
+  if (resp.status === ResponseStatusCode.UNAUTHORIZED) {
+    const contentType = resp.headers.get('content-type');
+
+    if (contentType.includes('application/json')) {
+      return resp.json().then(json => {
+        throw new UnauthorizedException(json.message || '');
+      });
+    }
+
+    throw new Error();
```
adamklepacz: Should this throw UnauthorizedException(); ??

client/src/modules/authorization/components/SignInForm.jsx
```diff
@@ -0,0 +1,219 @@
...
+
+  emailValidator = value => {
+    const { isEmail } = validator;
+
+    if (!isEmail(value)) {
+      return 'authorization.input.email.invalid';
+    }
+
+    return '';
+  };
```
adamklepacz: would be more understandable to set this errors to erros object inside state.

and then in render method destructure those errors from state as pass them to SignUpInput / AuthInput?? 

I think the validator should rather return boolean value or set info about error validation to state. 
Now its kinda hard to understand that validators are returning some strings and these strings are passed down to some other components ETC ETC - it's getting complicated. I think setting this to state and passing from state to a component is more readable.

client/src/modules/authorization/components/SignInForm.jsx
```diff
@@ -0,0 +1,219 @@
...
+      <div className="sign-in">
+        <h1 className="sign-in__heading">
+          <FormattedMessage id="authorization.auth-box.sign-in" />
+        </h1>
+        {signInErrorId && this.renderSignInError()}
+        <form className="sign-in__form" noValidate>
+          <AuthInput
+            disabled={pending}
+            formError={hasSignUpFailed}
+            focus
```
adamklepacz: sort props

client/src/modules/authorization/components/SignInForm.jsx
```diff
@@ -0,0 +1,219 @@
...
+            name="email"
+            onChange={this.onEmailChange}
+            type="text"
+            validator={this.emailValidator}
+          />
+          <AuthInput
+            disabled={pending}
+            formError={hasSignUpFailed}
+            labelId="authorization.input.password.label"
+            name="password"
```
adamklepacz: are these name attributes necessary on all of these inputs? 
what are the benefits of using them?

They don't have any SEO value 
https://webmasters.stackexchange.com/questions/61609/using-the-html-input-name-attribute-and-seo
http://maileohye.com/title-name-attributes-html-anchors/

client/src/modules/authorization/components/SignInForm.jsx
```diff
@@ -0,0 +1,219 @@
...
+          <FormattedMessage id="authorization.auth-box.sign-in" />
+        </h1>
+        {signInErrorId && this.renderSignInError()}
+        <form className="sign-in__form" noValidate>
+          <AuthInput
+            disabled={pending}
+            formError={hasSignUpFailed}
+            focus
+            labelId="authorization.input.email.label"
+            name="email"
```
adamklepacz: ??? 

check next comment

client/src/modules/authorization/AuthBox.jsx
```diff
@@ -147,11 +180,11 @@ class AuthBox extends PureComponent {
             <div className="authbox__intro">
               <AppLogo />
             </div>
-            {isSignUpFormVisible ? (
-              <SignUpForm onCancel={this.handleSignUpFormVisibility} />
+            {isSignUpFormVisible || isSignInFormVisible ? (
```
adamklepacz: you are checking it here, and inside renderForms method.
Does it make sense to check it 2 times?

wouldn't be enough  to just
`this.renderForms()` ?

client/src/modules/authorization/components/SignInForm.jsx
alemikolo: It's a custom component, not input itself. This name is used not only for input's name attribute but mainly for label, id. Maybe it's not important for SEO (name), but it's important for accessibility, screen readers, etc. It's not a big cost to use it

client/src/modules/authorization/components/SignInForm.jsx
alemikolo: This functionality is used in sign up form already. These strings are not passing down. You pass whole validator function to AuthInput (child component). And AuthInput (child) component uses that validator function and set errors in its state (when of course this is needed). AuthInput is a generic component. Its parent has to pass what kind of input it will be and what validator it should use to check entered value.
Child component uses a validator passed from a parent, so this validator has to inform the child about the message which has to be displayed, otherwise, the child should have hardcode all possible errorIds or there would have to be few very similar input components.

client/src/modules/authorization/AuthBox.jsx
alemikolo: Not exactly, because there is a ternary operator if signIn- or signUp- form isn't displayed buttons are displayed. 
So for the first time, I check if any of these forms should be displayed and inside this method, I have to check which one should be displayed.
I can leave this like it is
or change it to 3 separate conditions:
```
if(isSignUpFormVisible && (...)}
if(isSignInFormVisible && (...)}
if(!isSignInFormVisible &&  !isSignUpFormVisible && (...)}
```
for now, I change it to (it it almost the same but maybe more clear:
```
const areFormsVisible = isSignUpFormVisible || isSignInFormVisible;
(...)
{areFormsVisible ? (
  this.renderForms()
) : (
  <Fragment>
    {this.renderSignInButtons()}
    {this.renderSignUpButton()}
     {this.renderDemoButton()}
   </Fragment>
 )}
(...)
}

adamklepacz: > const areFormsVisible = isSignUpFormVisible || isSignInFormVisible;
> (...)
> {areFormsVisible ? (
>   this.renderForms()
> ) : (
>   <Fragment>
>     {this.renderSignInButtons()}
>     {this.renderSignUpButton()}
>      {this.renderDemoButton()}
>    </Fragment>
>  )}
> (...)
> }

Leave like this.

PR #201, opened 2019-06-27
[EOC-225] Leaving lists

adamklepacz:


client/src/common/components/Members/components/MemberDetails.jsx
```diff
@@ -194,20 +203,32 @@ class MemberDetails extends PureComponent {
...
-    } = this.props;
+  handleLeave = () => {
+    const { onCohortLeave, onListLeave, route } = this.props;
 
     this.setState({ pending: true });
 
-    return leaveCohort(id, currentUserId, name).finally(() =>
-      this.setState({ pending: false })
-    );
+    if (route === Routes.LIST) {
```
alemikolo: proposal to not repeat creating abortable promise:
```
let action;
switch (route) {
  case Routes.LIST:
    action = onListLeave;
    break;
  case Routes.COHORT:
    action = onCohortLeave;
    break;
  default;
}

this.pendingPromise = makeAbortablePromise(action());
...
```

client/src/modules/list/model/actions.js
```diff
@@ -517,3 +526,21 @@ export const changeType = (listId, listName, type) => dispatch =>
...
+
+export const leaveList = (listId, userId, cohortId, userName) => dispatch =>
+  patchData(`/api/lists/${listId}/leave-list`)
+    .then(() => {
+      dispatch(leaveListSuccess(listId, userId));
+      createNotificationWithTimeout(dispatch, NotificationType.SUCCESS, {
+        notificationId: 'list.actions.leave-list',
+        data: userName
+      });
+      history.replace(`/${cohortId ? 'cohorts' : 'dashboard'}`);
```
alemikolo: hmm, why if you leave the cohort's list you should be redirected to all cohorts view? To the dashboard or this specific cohort view - what do you think?

client/src/common/components/Members/components/MemberDetails.jsx
```diff
@@ -194,20 +203,32 @@ class MemberDetails extends PureComponent {
...
 
-    return leaveCohort(id, currentUserId, name).finally(() =>
-      this.setState({ pending: false })
-    );
+    let action;
+    switch (route) {
+      case Routes.LIST:
+        action = onListLeave;
+        break;
+
```
marekrozmus: remove new line or add after each `break`

client/src/locales/en.json
```diff
@@ -244,6 +244,9 @@
   "list.actions.remove-member-role-fail": "Failed to remove owner role from: '{data}'.",
   "list.actions.change-type": "Sack '{data}' type changed.",
   "list.actions.change-type-fail": "Failed to change: '{data}' type.",
+  "list.actions.leave-list-one-owner": "You cant not leave the sack, as you are the only owner.",
```
marekrozmus: `cant not`

server/controllers/list.js
```diff
@@ -1233,6 +1233,57 @@ const deleteItem = (req, resp) => {
...
+    .exec()
+    .then(list => {
+      if (!list) {
+        throw new BadRequestException();
+      }
+
+      const { viewersIds, memberIds, ownerIds } = list;
+      const isOwner = checkIfArrayContainsUserId(ownerIds, userId);
+      const isMember = checkIfArrayContainsUserId(memberIds, userId);
+      const isViewer = checkIfArrayContainsUserId(viewersIds, userId);
```
marekrozmus: maybe it would be better to create methods like: 'isOwner(list, userId)`, `isMember(list, userId)` and use them in `if` statements?

adamklepacz: We had that before. 
But it was repeatable, every method has same code inside but was named differently.
So about 2 months ago we merged all these 3 methods into one.

marekrozmus: you can always create methods like this:
```
const isOwner = (list, userId) => checkIfArrayContainesUserId(list, userId)
```

client/src/modules/cohort/index.jsx
```diff
@@ -337,6 +359,7 @@ Cohort.propTypes = {
     isArchived: PropTypes.bool,
     name: PropTypes.string
   }),
+  currentUser: UserPropType.isRequired,
```
marekrozmus: alphabetical ordering

client/src/modules/list/model/actions.js
```diff
@@ -517,3 +526,21 @@ export const changeType = (listId, listName, type) => dispatch =>
         data: listName
       });
     });
+
+export const leaveList = (listId, userId, cohortId, userName) => dispatch =>
+  patchData(`/api/lists/${listId}/leave-list`)
```
marekrozmus: do we need this `-list` postfix? we know from context that it is about list with specific id

server/common/utils/helpers.js
```diff
@@ -19,6 +19,18 @@ const checkIfArrayContainsUserId = (idsArray, userId) => {
   return arrayOfStrings.indexOf(userIdAsString) !== -1;
 };
 
+const isOwner = (idsArray, userId) =>
```
marekrozmus: please change so that method gets cohort or list. Then we can use:
`isOwner(cohort, userId)`

server/common/utils/helpers.test.js
```diff
@@ -201,6 +204,71 @@ describe('function checkIfArrayContainsUserId', () => {
   });
 });
 
+describe('function isOwner', () => {
+  const list = listMock[0];
+  const { ownerIds } = list;
+
+  it('returns true if the passed array does contains the userID', () => {
```
marekrozmus: remove `does` and update casing in `userID`

server/common/utils/helpers.test.js
```diff
@@ -201,6 +204,71 @@ describe('function checkIfArrayContainsUserId', () => {
...
+
+    expect(result).toBe(false);
+  });
+});
+
+describe('function isMember', () => {
+  const list = listMock[0];
+  const { memberIds } = list;
+
+  it('returns true if the passed array does contains the userID', () => {
```
marekrozmus: remove does and update casing in userID

server/common/utils/helpers.test.js
```diff
@@ -201,6 +204,71 @@ describe('function checkIfArrayContainsUserId', () => {
...
+
+    expect(result).toBe(false);
+  });
+});
+
+describe('function isViewer', () => {
+  const list = listMock[0];
+  const { viewersIds } = list;
+
+  it('returns true if the passed array does contains the userID', () => {
```
marekrozmus: remove does and update casing in userID

server/common/utils/helpers.test.js
```diff
@@ -178,7 +181,7 @@ describe('function responseWithCohort', () => {
 describe('function checkIfArrayContainsUserId', () => {
   const idsArray = ['123', '456', '789'];
 
-  it('returns true if the passed array does contains the userID', () => {
+  it('returns true if the passed array does contains the userI', () => {
```
marekrozmus: `d` is missing

server/common/utils/helpers.test.js
```diff
@@ -201,6 +204,71 @@ describe('function checkIfArrayContainsUserId', () => {
...
+
+    expect(result).toBe(false);
+  });
+});
+
+describe('function isMember', () => {
+  const list = listMock[0];
+  const { memberIds } = list;
+
+  it('returns true if the passed array does contains the userId', () => {
```
marekrozmus: `does contains` should be just `contains`

server/common/utils/helpers.test.js
```diff
@@ -201,6 +204,71 @@ describe('function checkIfArrayContainsUserId', () => {
...
+
+    expect(result).toBe(false);
+  });
+});
+
+describe('function isViewer', () => {
+  const list = listMock[0];
+  const { viewersIds } = list;
+
+  it('returns true if the passed array does contains the userId', () => {
```
marekrozmus: `does contains` should be just `contains`

PR #202, opened 2019-07-01
[EOC-239] Add reset password backend with sending email

adamklepacz:
Zrzut ekranu 2019-07-01 o 10 11 28


server/controllers/userAuth.js
```diff
@@ -126,9 +126,74 @@ const signUp = (req, resp, next) => {
...
+    return resp
+      .status(400)
+      .send({ message: 'authorization.actions.wrong-email' });
+  }
+
+  User.findOne({ email: sanitizedEmail })
+    .exec()
+    .then(user => {
+      if (!user) {
+        throw new Error('authorization.actions.no-user-of-email');
```
alemikolo: proposal: what do you think about some route params, redirect to them and then using this param on frontend (match: { params: { your-param }}}) and decide about messageId:
frontend:
```
<Route component={ResetPassword} path="/reset-password/:result?" />
```
backend:
```
return resp.redirect('/reset-password/no-user');
return resp.redirect('/reset-password/provider-user');
return resp.redirect('/reset-password/inactive-user');
```

server/controllers/userAuth.js
```diff
@@ -126,9 +126,74 @@ const signUp = (req, resp, next) => {
...
+const resetPassword = (req, resp, next) => {
+  const resetToken = crypto.randomBytes(32).toString('hex');
+  const resetTokenExpirationDate = new Date().getTime() + 3600000;
+  const { email } = req.body;
+  const sanitizedEmail = sanitize(email);
+  const { isEmail, isEmpty } = validator;
+
+  if (isEmpty(sanitizedEmail) || !isEmail(sanitizedEmail)) {
+    return resp
+      .status(400)
```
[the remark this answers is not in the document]
adamklepacz: yeah it should be normal 400, I've put it by a mistake

server/controllers/userAuth.js
adamklepacz: I haven't seen such a solution for handling errors.
I don't think it is a proper way to do it. 

@marekrozmus what do you think?

client/src/locales/en.json
```diff
@@ -91,6 +91,10 @@
   "authorization.reset-password.heading": "Reset your password",
   "authorization.reset-password.email-label": "Enter email",
   "authorization.reset-password.button-content": "Reset password",
+  "authorization.actions.no-user-of-email": "There is no user of email {data}.",
```
marekrozmus: We shouldn't inform about such situation. We should just show information that email was sent and that is all. On backend we should only send if email exists in database. Nobody needs to know if that email is registered in our app or not.

client/src/locales/en.json
```diff
@@ -91,6 +91,10 @@
...
+  "authorization.actions.google-user-no-reset-password": "This user: {data} is signed up via Google Account. You can not reset password to Google account here.",
```
marekrozmus: This is also too much information. We should have some general information for both cases like: "An email with instructions to create a new password has been sent to {email} if it is associated with an EOC account. Your existing password has not been changed." This one is from NVIDIA - check some other services and pick the best. But as You can see they do not say if that email is registered in their DB or not.

server/controllers/userAuth.js
```diff
@@ -126,9 +126,72 @@ const signUp = (req, resp, next) => {
...
+          // eslint-disable-next-line no-param-reassign
+          resp.locales = {
+            displayName,
+            email,
+            resetToken
+          };
+
+          next();
+        })
+        .catch(() => {
```
marekrozmus: why do you catch and throw new error? What is wrong with the catched one ;)?

server/mailer/index.js
```diff
@@ -68,4 +68,38 @@ const sendSignUpConfirmationLink = (req, resp) => {
     .catch(() => resp.sendStatus(400));
 };
 
-module.exports = { sendInvitation, sendSignUpConfirmationLink };
+const sendResetPasswordLink = (req, resp) => {
+  const { email: receiver, displayName, resetToken } = resp.locales;
+  const { protocol } = req;
+  const host = req.get('host');
+  const url = `${protocol}://${host}`;
```
marekrozmus: do you need `protocol`? Isn't `host` is enough?

server/mailer/index.js
```diff
@@ -68,4 +68,38 @@ const sendSignUpConfirmationLink = (req, resp) => {
...
+  const resetUrl = `${url}/reset/${resetToken}`;
+  const title = `${PROJECT_NAME} - Reset your password`;
+  const info =
+    'Reset your password by clicking reset button. If you have not requested password reset to your account, just ignore this message.';
+  const value = 'Reset password';
+
+  const message = {
+    to: receiver,
+    from: 'no.replay@app.eoc.com',
```
marekrozmus: `reply`?

server/controllers/userAuth.js
adamklepacz: This is because when the user somehow avoids email validation on the front end, the validation rules from backend will send error and by catching that error here I can show the "ERROR MESSAGE" under the input once again here. 
Without this catch, the user will not get any response.

server/mailer/index.js
adamklepacz: It works without protocol, but I have added this because on production server we have https
and I thought this solution is more bulletproof. 
I can remove, what do you think?

server/controllers/userAuth.js
alemikolo: but validation is made before the query is sent DB and if something is wrong with email status 406 is send and nothing else is executed

alemikolo: hmm consider: 
```
const resetPassword = (req, resp, next) => {
  const { email } = req.body;
  const sanitizedEmail = sanitize(email);
  const { isEmail } = validator;

  if (!isEmail(sanitizedEmail)) {
    return resp.sendStatus(406);
  }

  User.findOne({ email: sanitizedEmail })
    .exec()
    .then(user => {
      if (!user) {
        throw new Error();
      }

      const { isActive, idFromProvider } = user;

      if (idFromProvider) {
        throw new Error('authorization.actions.reset-successful');
      }

      if (!isActive) {
        throw new Error(
          'authorization.actions.reset-password-not-active-account'
        );
      }

      const resetToken = crypto.randomBytes(32).toString('hex');
      const resetTokenExpirationDate = new Date().getTime() + 3600000;

      return User.findOneAndUpdate(
        {
          email
        },
        { resetToken, resetTokenExpirationDate },
        { new: true }
      ).exec();
    })
    .then(user => {
      const { displayName, email, resetToken } = user;
      // eslint-disable-next-line no-param-reassign
      resp.locales = {
        displayName,
        email,
        resetToken
      };

      next();
    })
    .catch(err => {
      const { message } = err;

      if (message) {
        return resp.status(400).send({ message });
      }

      resp.sendStatus(400);
    });
};
```

server/controllers/userAuth.js
```diff
@@ -141,13 +141,13 @@ const resetPassword = (req, resp, next) => {
...
       if (!user) {
-        throw new Error('authorization.actions.no-user-of-email');
+        throw new Error();
       }
 
       const { displayName, isActive, idFromProvider } = user;
 
       if (idFromProvider) {
-        throw new Error('authorization.actions.google-user-no-reset-password');
+        throw new Error('authorization.actions.reset-successful');
```
alemikolo: throw an error with message of id successful? It's unclear for me

server/controllers/userAuth.js
adamklepacz: > hmm consider:
> 
> ```
> const resetPassword = (req, resp, next) => {
>   const { email } = req.body;
>   const sanitizedEmail = sanitize(email);
>   const { isEmail } = validator;
> 
>   if (!isEmail(sanitizedEmail)) {
>     return resp.sendStatus(406);
>   }
> 
>   User.findOne({ email: sanitizedEmail })
>     .exec()
>     .then(user => {
>       if (!user) {
>         throw new Error();
>       }
> 
>       const { isActive, idFromProvider } = user;
> 
>       if (idFromProvider) {
>         throw new Error('authorization.actions.reset-successful');
>       }
> 
>       if (!isActive) {
>         throw new Error(
>           'authorization.actions.reset-password-not-active-account'
>         );
>       }
> 
>       const resetToken = crypto.randomBytes(32).toString('hex');
>       const resetTokenExpirationDate = new Date().getTime() + 3600000;
> 
>       return User.findOneAndUpdate(
>         {
>           email
>         },
>         { resetToken, resetTokenExpirationDate },
>         { new: true }
>       ).exec();
>     })
>     .then(user => {
>       const { displayName, email, resetToken } = user;
>       // eslint-disable-next-line no-param-reassign
>       resp.locales = {
>         displayName,
>         email,
>         resetToken
>       };
> 
>       next();
>     })
>     .catch(err => {
>       const { message } = err;
> 
>       if (message) {
>         return resp.status(400).send({ message });
>       }
> 
>       resp.sendStatus(400);
>     });
> };
> ```

this is 99% the same implementation that I have.

server/mailer/index.js
marekrozmus: it will be redirected automatically - you don't need to use `protocol`

server/controllers/userAuth.js
```diff
@@ -126,9 +126,70 @@ const signUp = (req, resp, next) => {
...
+      if (!isActive) {
+        throw new Error(
+          'authorization.actions.reset-password-not-active-account'
+        );
+      }
+
+      const resetToken = crypto.randomBytes(32).toString('hex');
+      const resetTokenExpirationDate = new Date().getTime() + 3600000;
+
+      return User.findOneAndUpdate(
```
marekrozmus: do you need to search for this document again? Are you using unsanitized email?

adamklepacz: yes, because I have to update User model properties.

Sanitized email fixed.

PR #203, opened 2019-07-01
[EOC-248] Closing add item form on the list view

adamklepacz:


client/src/modules/list/components/Items/InputBar/InputBar.jsx
```diff
@@ -35,6 +35,10 @@ class InputBar extends Component {
     this.socket = undefined;
   }
 
+  componentDidMount() {
+    document.addEventListener('keydown', this.handleEscapePress);
```
alemikolo: It can't be done like this, because this handler is added always when you launch the list view and it is never removed until you exit list view. Currently, it is invoked always you press escape key wheteher or not you use this inputBar. Maybe add this handler on focus and remove on blur and componentUnMount?

client/src/modules/list/components/Items/InputBar/InputBar.jsx
```diff
@@ -151,6 +143,12 @@ class InputBar extends Component {
     if (isInputEmpty) {
       this.hideForm();
     }
+
+    document.removeEventListener('keydown', this.handleEscapePress);
```
alemikolo: This handleEscapePress is added correctly now - only when the input gets focus, but removing is not working yet - when you close input via escape press - the listener isn't removed and it is still invoked, so I think we it should be removed onEscapePress also

client/src/modules/list/components/Items/InputBar/InputBar.jsx
```diff
@@ -143,8 +146,6 @@ class InputBar extends Component {
     if (isInputEmpty) {
       this.hideForm();
     }
```
alemikolo: when input loses focus listener is still invoking, I think that on blur we should always remove the listener, not only when the input isn't empty.

PR #205, opened 2019-07-03
[EOC-237] Sign in backend update with login password

alemikolo:
Sending user data via cookie is removed.


server/config/auth.js
```diff
@@ -26,54 +33,80 @@ passport.use(
...
+passport.use(
+  StrategyType.LOCAL,
+  new LocalStrategy(
+    {
+      usernameField: 'email',
+      passwordField: 'password'
+    },
+    (email, password, done) => {
+      findAndAuthenticateUser(email, password)
+        .then(user => {
```
marekrozmus: one line: `user => done(null, user ? user : false)

server/config/auth.js
```diff
@@ -26,54 +33,80 @@ passport.use(
...
-        return seedDemoData(newUser._id);
+      User.findOne({
+        idFromProvider: DEMO_MODE_ID
       })
-      .then(() => done(null, newUser))
-      .catch(err => done(null, false, { message: err.message }));
-  })
+        .lean()
+        .exec()
+        .then(user => {
```
marekrozmus: one line: `user => user ? user : new User({ ...createDemoUser() }).save()`

server/config/auth.js
```diff
@@ -89,20 +122,46 @@ passport.deserializeUser((id, done) => {
...
+    if (!user) {
+      return resp.sendStatus(401);
+    }
+
+    req.login(user, err => {
+      if (err) {
+        return next(err);
+      }
+
+      next();
```
marekrozmus: is `next()` same as `next(falsyValue)`? if yes then one line can be used: `err => next(err)`

alemikolo: This next() functionality is a little bit complicated for me. If an error is 

> ... returned from asynchronous functions invoked by route handlers and middleware, you must pass them to the next() function, where Express will catch and process them.
> If you pass anything to the next() function (except the string 'route'), Express regards the current request as being an error and will skip any remaining non-error handling routing and middleware functions.

So I think, that if I pass anything to the 'next()' function,  express.js will treat it as an error. So I if I would like to go to the next function in a chain of middlewares I should call next() function with no params.

client/src/modules/authorization/model/actions.js
```diff
@@ -75,17 +56,27 @@ export const resendConfirmationLink = hash =>
...
+    .then(resp => resp.json())
+    .then(json => {
+      dispatch(loginSuccess(json));
       createNotificationWithTimeout(dispatch, NotificationType.SUCCESS, {
         notificationId: 'authorization.actions.login'
       });
+    });
+
+export const getLoggedUser = () => dispatch =>
+  getData('auth/user')
```
marekrozmus: shouldn't be `/auth/user`?

server/config/auth.js
```diff
@@ -26,54 +33,70 @@ passport.use(
...
+passport.use(
+  StrategyType.LOCAL,
+  new LocalStrategy(
+    {
+      usernameField: 'email',
+      passwordField: 'password'
+    },
+    (email, password, done) => {
+      findAndAuthenticateUser(email, password)
+        .then(user => {
```
marekrozmus: do you need braces?

client/src/app/components/Layout/Layout.jsx
```diff
@@ -53,9 +56,12 @@ export class Layout extends PureComponent {
   }
 
   setAuthenticationState = () => {
-    const { loginUser } = this.props;
+    const { currentUser, getLoggedUser } = this.props;
```
adamklepacz: line of break

client/src/modules/authorization/components/AuthInput.jsx
```diff
@@ -59,11 +59,11 @@ class AuthInput extends PureComponent {
 
   setValidationTheme = errorMessageId => {
     const { value } = this.state;
-    const { externalErrorId, formError } = this.props;
+    const { externalErrorId, formError, noValidate } = this.props;
 
     const isValid = !externalErrorId && !errorMessageId;
     const invalidTheme = !isValid || formError;
-    const validTheme = isValid && value && !formError;
+    const validTheme = isValid && value && !formError && !noValidate;
```
adamklepacz: what does validTheme do?

client/src/modules/authorization/components/SignInForm.jsx
```diff
@@ -159,7 +160,11 @@ class SignInForm extends PureComponent {
           <FormattedMessage id="authorization.auth-box.sign-in" />
         </h1>
         {signInErrorId && this.renderSignInError()}
-        <form className="sign-in__form" noValidate>
+        <form
+          className="sign-in__form"
+          noValidate
```
adamklepacz: what does this noValidate prop do?

server/config/auth.js
```diff
@@ -26,54 +33,68 @@ passport.use(
...
 
+// Use LocalStrategy to authenticate user
+passport.use(
+  StrategyType.LOCAL,
+  new LocalStrategy(
+    {
+      usernameField: 'email',
+      passwordField: 'password'
+    },
+    (email, password, done) => {
```
adamklepacz: unnecessary braces

server/config/auth.js
```diff
@@ -26,54 +33,68 @@ passport.use(
...
-          provider,
-          surname
-        }).save();
-      })
-      .then(user => {
-        newUser = user;
+  StrategyType.DEMO,
+  new LocalStrategy(
+    { usernameField: 'email', passwordField: 'password' },
+    (email, password, done) => {
```
adamklepacz: unnecessary braces

server/config/auth.js
```diff
@@ -89,20 +110,46 @@ passport.deserializeUser((id, done) => {
     .catch(err => done(null, false, { message: err.message }));
 });
 
-const setDemoUser = passport.authenticate('local', {
-  failureRedirect: '/'
-});
+const authenticate = (req, resp, next, type) => {
```
adamklepacz: unnecessary braces

server/config/auth.js
```diff
@@ -89,20 +110,46 @@ passport.deserializeUser((id, done) => {
...
+      if (err) {
+        return next(err);
+      }
+
+      next();
+    });
+  })(req, resp, next);
+};
+
+const setUser = (req, resp, next) => {
```
adamklepacz: unnecessary braces

server/config/auth.js
```diff
@@ -89,20 +110,46 @@ passport.deserializeUser((id, done) => {
...
+  authenticate(req, resp, next, StrategyType.LOCAL);
+};
+
+const setDemoUser = (req, resp, next) => {
```
adamklepacz: unnecessary braces

server/controllers/userAuth.js
```diff
@@ -9,30 +9,18 @@ const User = require('../models/user.model');
 
 const sendUser = (req, resp) => {
   const { avatarUrl, _id: id, displayName: name } = req.user;
-
-  resp.cookie('user', JSON.stringify({ avatarUrl, id, name }));
-  resp.redirect('/');
+  resp.send({ avatarUrl, id, name });
```
adamklepacz: line of break above

client/src/modules/authorization/components/AuthInput.jsx
alemikolo: validTheme and invalidTheme decide if and which styles should be applied. These props were here before

client/src/modules/authorization/components/SignInForm.jsx
alemikolo: Maybe this prop has an unclear name. In some special situations, you can pass it to AuthInput to prevent displaying a green color even if the input is valid. This component is used in different forms so it can be configured a little bit different.

server/config/auth.js
alemikolo: The brackets are needed here

client/src/modules/authorization/components/SignInForm.jsx
adamklepacz: maybe 
`noSuccessTheme`  would be better?

`noValidate` sounds like there is a validate function somewhere, and it should not validate when this prop is passed

client/src/modules/authorization/components/SignInForm.jsx
```diff
@@ -175,6 +180,7 @@ class SignInForm extends PureComponent {
             formError={hasSignUpFailed}
             labelId="authorization.input.password.label"
             name="password"
+            noSuccessTheme
```
marekrozmus: what does it do?

alemikolo: If the input is valid AuthInput displays this input in green. If there is no validator this field is always valid, so if the user enters something its appearance change to green. This is default behavior but in some special situations, this prop helps to prevent displaying a green color even if the input is valid.

alemikolo:
I had to add a small fix to the layout component. On page refresh dashboard always was displayed. This error occurred because I changed the way of sending user data to the client and old logic didn't work.

client/src/app/components/Layout/Layout.jsx
```diff
@@ -25,39 +26,36 @@ import Cohorts from 'modules/cohort/components/Cohorts';
...
-    } = prevProps;
+    } = this.props;
 
-    if (!prevCurrentUser && currentUser) {
-      const newPath = pathname === '/' ? '/dashboard' : pathname;
-      history.push(newPath);
+    if (_isEmpty(currentUser)) {
+      this.setState({ pending: true });
+      getLoggedUser()
+        .then(() => {
```
marekrozmus: remove braces

PR #207, opened 2019-07-04
[EOC-246] Prevent item description edition when link is clicked

alemikolo:


client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -87,6 +87,12 @@ class ListItemDescription extends PureComponent {
   };
 
   handleShowTextarea = event => {
+    const { nodeName } = event.target;
+
+    if (nodeName === 'A') {
```
marekrozmus: what does it do? where is node with name `A`?

alemikolo: Item's description can contain links. This div which contains the item's description has the onClick handler attached. On click, it turns on description edit mode, but when the user clicks a link inside this div, edit mode shouldn't be turned on.
Event.target.nodeName is a read-only property and returns the name of the current Node as a string. When the user clicks a link inside this clickable div nodeName will return "A". Thanks to that I can detect that anchor tag was clicked and prevent turning edit mode on. But the link is still working normally.

adamklepacz: maybe create enum like
const NodeTypes = Object.freeze({
     LINK: 'A'
});

and compare to it

PR #208, opened 2019-07-04
[EOC-274] Block adding user to list or cohort when account is inactive

alemikolo:


server/controllers/cohort.js
```diff
@@ -422,7 +422,7 @@ const addMember = (req, resp) => {
       return User.findOne({ email: sanitize(email) }).exec();
     })
     .then(user => {
-      if (!user) {
+      if (!user || user.isActive === false) {
```
marekrozmus: `!user || !user.isActive` - do not compare boolean value to `true` or `false`

server/controllers/list.js
```diff
@@ -877,7 +877,7 @@ const addViewer = (req, resp) => {
       return User.findOne({ email: sanitize(email) }).exec();
     })
     .then(userData => {
-      if (!userData) {
+      if (!userData || userData.isActive === false) {
```
marekrozmus: same here

alemikolo: I think that in this case, I have to compare to false because of users authenticated via Google - they don't have property isActive. For them, it will be `undefined` so this comparison will be true then. Or I can check: ` if(!user || (!idFromProvider && !user.isActive)`

server/controllers/cohort.js
alemikolo: I think that in this case, I have to compare to false because of users authenticated via Google - they don't have property isActive. For them, it will be `undefined` so this comparison will be true then. Or I can check: ` if(!user || (!idFromProvider && !user.isActive)`

adamklepacz: undefined is also a falsy value so you will get true if isActive filed is empty, undefined or null
!null = true
!undefined = true
!'' = true

adamklepacz: if(!user || (!idFromProvider && !user.isActive)

this is good I think

alemikolo:
I change it to ` if (!user || (!user.idFromProvider && !user.isActive))`

PR #209, opened 2019-07-04
[EOC-270] Block access to lists in the archived cohort

alemikolo:


server/common/utils/helpers.js
```diff
@@ -60,25 +61,34 @@ const responseWithList = (list, userId) => {
...
-      listToSend.cohortId = cohortId;
-    }
-
-    return listToSend;
-  });
+  _filter(
+    _map(lists, list => {
+      const { cohortId: cohort, favIds, items, ...rest } = list;
+
+      if (cohort && cohort.isArchived === true) {
```
marekrozmus: do not compare boolean values to `true` or `false` - just use them

server/common/utils/helpers.js
```diff
@@ -60,25 +61,34 @@ const responseWithList = (list, userId) => {
...
+        unhandledItemsCount
+      };
+
+      if (cohortId) {
+        listToSend.cohortId = cohortId;
+      }
+
+      return listToSend;
+    }),
+    list => list
```
marekrozmus: why filter `list => list`?

alemikolo: Because of lines 68-70: in _map() I check that if this is cohort's list and if this cohort is Archived. If so I exit current iteration. In the final array for all lists in archived cohorts, I get `undefined` so I use this strange filer to return only existing lists without these undefines.

server/common/utils/helpers.js
```diff
@@ -60,25 +61,34 @@ const responseWithList = (list, userId) => {
...
+      const { cohortId: cohort, favIds, items, ...rest } = list;
+
+      if (cohort && !cohort.isArchived) {
+        return;
+      }
+
+      const activeItems = items.filter(item => !item.isArchived);
+      const doneItemsCount = activeItems.filter(item => item.isOrdered).length;
+      const unhandledItemsCount = activeItems.length - doneItemsCount;
+      const cohortId = cohort ? cohort._id : null;
```
adamklepacz: do you need to do this?
if cohortId will be undefined or null it will be falsy, if there will be a value it will be truthy so it will work the same way

server/common/utils/helpers.js
marekrozmus: maybe use in this case https://lodash.com/docs/4.17.11#compact

PR #211, opened 2019-07-05
[EOC-247] Refactor members reducer from array to object with

alemikolo:


client/src/modules/cohort/model/reducer.js
```diff
@@ -3,57 +3,44 @@ import _keyBy from 'lodash/keyBy';
...
-const membersReducer = (state = [], action) => {
+const membersReducer = (state = {}, action) => {
   switch (action.type) {
     case CohortActionTypes.ADD_MEMBER_SUCCESS: {
       const {
         payload: { data }
       } = action;
 
-      return [...state, data];
+      return { [data._id]: data, ...state };
```
adamklepacz: destructure _id from data

client/src/modules/list/model/reducer.js
```diff
@@ -9,79 +9,64 @@ import {
...
-const membersReducer = (state = [], action) => {
+const membersReducer = (state = {}, action) => {
   switch (action.type) {
     case ListActionTypes.ADD_VIEWER_SUCCESS: {
       const {
         payload: { data }
       } = action;
 
-      return [...state, data];
+      return { [data._id]: data, ...state };
```
adamklepacz: destructure _id

client/src/modules/cohort/model/selectors.js
```diff
@@ -36,5 +36,5 @@ export const getArchivedCohorts = createSelector(
 
 export const getMembers = createSelector(
   getCohortDetails,
-  cohort => cohort && _keyBy(cohort.members, '_id')
+  cohort => cohort && cohort.members
```
marekrozmus: this should be only: `cohort => cohort.members` if no members then empty object should be returned. Why checking if cohort exists?

client/src/modules/list/model/reducer.js
```diff
@@ -9,79 +9,64 @@ import {
...
-          ? {
-              ...member,
-              isMember: false,
-              isOwner: false
-            }
-          : member
-      );
+      return {
+        ...state,
+        [userId]: { ...state[userId], isOwner: false, isMember: false }
```
marekrozmus: if member role is removed then iwner role is also removed? What i n case if we are the only owner and we remove member role?

client/src/modules/list/model/selectors.js
```diff
@@ -106,5 +106,5 @@ export const getCohortArchivedLists = createSelector(
 
 export const getMembers = createSelector(
   getList,
-  list => list && _keyBy(list.members, '_id')
+  list => list && list.members
```
marekrozmus: same issue as with cohort

client/src/modules/list/model/reducer.js
alemikolo: We have to be a member if we want to be an owner. We can't have a higher level role without having a lower level one. We can't remove a member role when we are the only owner.

client/src/modules/cohort/model/selectors.js
alemikolo: without this check we get an error: Uncaught TypeError: Cannot read property 'members' of undefined

client/src/modules/list/model/selectors.js
alemikolo: without this check we get an error: Uncaught TypeError: Cannot read property 'members' of undefined

client/src/modules/list/model/reducer.js
```diff
@@ -13,10 +13,13 @@ const membersReducer = (state = {}, action) => {
   switch (action.type) {
     case ListActionTypes.ADD_VIEWER_SUCCESS: {
       const {
-        payload: { data }
+        payload: {
+          data,
+          data: { _id: id }
```
adamklepacz: just _id

client/src/modules/cohort/model/reducer.js
```diff
@@ -7,10 +7,13 @@ const membersReducer = (state = {}, action) => {
   switch (action.type) {
     case CohortActionTypes.ADD_MEMBER_SUCCESS: {
       const {
-        payload: { data }
+        payload: {
+          data,
+          data: { _id: id }
```
adamklepacz: just _id

client/src/modules/cohort/model/selectors.js
marekrozmus: ok, let's leave it as it is for now - we will need to review cohort component and it's model/selectors

PR #213, opened 2019-07-05
[EOC-281] App crashes on entering deleted list via url

adamklepacz:


client/src/modules/list/index.jsx
```diff
@@ -60,12 +61,14 @@ class List extends Component {
...
-      this.receiveWSEvents();
-    });
+    this.fetchData()
+      .then(() => {
+        this.handleBreadcrumbs();
+        this.handleSocketConnection();
+        this.receiveWSEvents();
+        this.setState({ pendingForDetails: false });
+      })
+      .catch(() => noOp());
```
alemikolo: what if an error occurs and this won't be NotFoundExceptio e.g with status 400? You won't be redirected to 404 so how pending will be turning off then? Will this situation cause some problems?

client/src/common/exceptions/NotFoundException.js
```diff
@@ -0,0 +1 @@
+export class NotFoundException extends Error {}
```
marekrozmus: What is 'not found`?

adamklepacz: Exception for 404 error

PR #214, opened 2019-07-05
[EOC-278] & [EOC-282] [BUG] leaving cohort but reamin on lits cause an error when back to cohort/ [BUG] App crashes on entering deleted cohort via URL

alemikolo:


server/controllers/list.js
```diff
@@ -304,7 +304,7 @@ const getListData = (req, resp) => {
         cohortName,
         description,
         isArchived,
-        isGuest: !isMember(list, userId),
+        isGuest: cohortId && !isMember(cohort, userId),
```
adamklepacz: what is this is a private list not within the cohort??
Does it recognize that the user is a guest??

PR #215, opened 2019-07-08
[EOC-279] Resend password reset link when reset token is expired

adamklepacz:
Reset password flow is now complete.

Zrzut ekranu 2019-07-08 o 10 18 57
Zrzut ekranu 2019-07-08 o 10 37 58


client/src/locales/en.json
```diff
@@ -79,7 +79,8 @@
   "authorization.sign-in": "Sign in",
   "authorization.sign-up": "Sign up",
   "authorization.sign-up.confirmation-link-sent": "In order to complete sign up process please click the link that we have just sent to {data}.",
-  "authorization.sign-up.button.resend-link": "Resend activation link",
+  "authorization.sign-up.button.resend-link": "Resend link",
```
alemikolo: maybe: Resend recovery link

client/src/locales/en.json
```diff
@@ -79,7 +79,8 @@
...
+  "authorization.link-expired.recovery-expired": "Password recovery link expired.",
```
alemikolo: has expired

client/src/modules/authorization/components/LinkExpired.jsx
```diff
@@ -16,11 +20,22 @@ class LinkExpired extends PureComponent {
...
-        params: { hash }
+        params: { token },
+        path
       }
     } = this.props;
 
-    return resendConfirmationLink(hash)
+    let action;
+
+    if (path.includes(Routes.CONFIRMATION_LINK_EXPIRED)) {
```
alemikolo: proposal: maybe use else or switch to prevent checking both if statements every time

client/src/modules/authorization/model/actions.js
```diff
@@ -113,12 +113,14 @@ export const resetPassword = email => dispatch =>
...
     passwordConfirmation
   })
     .then(() => history.replace('/password-recovery-success'))
     .catch(err => {
-      // TODO: IF err instanceof BadRequestException redirect to 'LinkExpired' component
       throw err;
     });
+
+export const resendRecoveryLink = token =>
+  postData(`/auth/resend-recovery-link/${token}`);
```
alemikolo: maybe if this is post send the token in request body instead of in the URL?

server/controllers/userAuth.js
```diff
@@ -301,9 +320,62 @@ const updatePassword = (req, resp) => {
...
+  )
+    .lean()
+    .exec()
+    .then(user => {
+      if (!user) {
+        throw new Error('authorization.actions.reset');
+      }
+      const { isActive } = user;
+
+      if (!isActive) {
```
alemikolo: You can disable resetting password for inactive accounts in query":
```
  User.findOneAndUpdate(
    {
      resetToken: sanitizedToken,
      isActive: true
    },
(...)
```
Of course, we won't be able to return a specific message for this case as you did in line 348, but we try to avoid setting messages Id on the backend and in I'm wondering if this is really necessary for this situation...

client/src/locales/en.json
adamklepacz: It's a common button now.
Displayed when the confrimation and recovery process is running

client/src/modules/authorization/model/actions.js
adamklepacz: why not in url ?

server/controllers/userAuth.js
adamklepacz: I did it on purpose. Otherwise, user will have no idea why he can not reset the password

client/src/modules/authorization/components/LinkExpired.jsx
adamklepacz: how can I use switch here?

server/controllers/userAuth.js
alemikolo: I know you did it on purpose. But user with inactive account can't reset password: in resetPassword controller, we check isActive field and if it is false we throw an error and don't send an email with the link. So I think such a user won't be able to access linkExpired page with a proper token in URL resending link will fail. So, in my opinion, sending this additional message is unnecessary here. But I maybe I didn't understand this logic correctly...

client/src/modules/authorization/model/actions.js
alemikolo: leave as it is. This more logical for me. When we send data via POST we usually send it in the request body, except listId and cohortId...

client/src/modules/authorization/components/LinkExpired.jsx
alemikolo: Switch: my mistake. For me, if else would be better than to ifs but leave as it is if you want

server/controllers/userAuth.js
adamklepacz: @marekrozmus  what do you think?
Should we send here specific message?
Or just add isActive to query and throw a basic error when user account is not activated?

marekrozmus: if account is not activated or account does not exist send some common error like: "account does not exist or not activated" - user should know if he/she activated/created account.

adamklepacz: Done

client/src/modules/authorization/model/actions.js
```diff
@@ -106,19 +106,23 @@ export const resetPassword = email => dispatch =>
...
     });
 
 export const updatePassword = (token, password, passwordConfirmation) =>
-  postData(`/auth/recovery-password/${token}`, {
+  postData(`/auth/update-password/${token}`, {
     password,
     passwordConfirmation
   })
     .then(() => history.replace('/password-recovery-success'))
     .catch(err => {
```
marekrozmus: this catch is not needed if you just rethrow the error

PR #217, opened 2019-07-08
[EOC-277] Archived cohort view disable restore and delete for member

alemikolo:
I merged branch EOC-278 to this branch to avoid some errors which were fixed there.


client/src/common/components/ArchivedMessage.jsx
```diff
@@ -59,25 +60,27 @@ class ArchivedMessage extends PureComponent {
...
-              className="archived-message__button primary-button"
-              disabled={pending}
-              type="button"
-              onClick={this.handleRestoring}
-            >
-              <FormattedMessage id="common.archived-message.restore-button" />
-            </button>
-            {pending && !isDialogVisible && <Preloader />}
-          </div>
+          {isOwner && (
```
adamklepacz: you can also early return above this return 
```

if(!isOwner) {
    return null;
}
```

client/src/common/exceptions/NotFoundException.js
```diff
@@ -0,0 +1 @@
+export class NotFoundException extends Error {}
```
adamklepacz: Please also rename this to SourceNotFoundException

client/src/common/exceptions/index.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable import/default */
+import { AbortPromiseException } from './AbortPromiseException';
+import { NotFoundException } from './NotFoundException';
+import { UnauthorizedException } from './UnauthorizedException';
+import { ValidationException } from './ValidationException';
+
+export {
+  AbortPromiseException,
+  NotFoundException,
```
adamklepacz: SoucreNotFoundException

client/src/modules/cohort/index.jsx
```diff
@@ -81,17 +90,32 @@ class Cohort extends PureComponent {
...
-      })
-      .finally(() => {
         this.handleBreadcrumbs();
         this.setState({ pendingForDetails: false });
+      })
+      .catch(err => {
+        if (!(err instanceof AbortPromiseException)) {
+          this.setState({ pendingForDetails: false });
+        }
+        noOp();
```
adamklepacz: @marekrozmus  is this noOp needed here? 
if an err occurs we just ignore them. Do we have to run noOp function here which takes time?

server/controllers/cohort.js
```diff
@@ -506,7 +512,7 @@ const leaveCohort = (req, resp) => {
 
       const { memberIds, ownerIds } = doc;
 
-      if (isOwner(doc, userId)) {
+      if (isCohortOwner(doc, userId)) {
```
adamklepacz: in line 186. you have this assigned to a variable and here you are calling this within if. 
Please select one approach to be consistent

client/src/modules/cohort/index.jsx
alemikolo: I removed noOp

PR #220, opened 2019-07-09
[EOC-250] Editing item name and description

adamklepacz:
I have also added the Higher order component on this branch which holds the socket.io connection. - all others sockets connection are removed from components.
Every component that needs connection is wrapped by withSocket() HOC and then it has access to the same instance of socket connection shared across the app.

### DIFF
--- a/client/src/modules/list/components/Items/InputBar/InputBar.jsx
+++ b/client/src/modules/list/components/Items/InputBar/InputBar.jsx
@@ -6,7 +6,6 @@ import classNames from 'classnames';
 import _trim from 'lodash/trim';
 import { injectIntl, FormattedMessage } from 'react-intl';
 import _flowRight from 'lodash/flowRight';
-import io from 'socket.io-client';
 
 import { getCurrentUser } from 'modules/authorization/model/selectors';
 import {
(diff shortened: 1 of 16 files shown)


client/src/modules/list/components/Items/InputBar/InputBar.jsx
```diff
@@ -227,6 +222,7 @@ InputBar.propTypes = {
   currentUser: UserPropType.isRequired,
   intl: IntlPropType.isRequired,
   match: RouterMatchPropType.isRequired,
+  socket: PropTypes.objectOf(PropTypes.any),
```
marekrozmus: why not just `PropTypes.object`?

client/src/modules/list/components/Items/ListItemName/ListItemName.jsx
```diff
@@ -155,15 +180,20 @@ ListItemName.propTypes = {
   itemId: PropTypes.string.isRequired,
   match: RouterMatchPropType.isRequired,
   name: PropTypes.string.isRequired,
+  onBusy: PropTypes.func.isRequired,
+  onFree: PropTypes.func.isRequired,
```
marekrozmus: move to functions block

client/src/modules/list/components/Items/InputBar/InputBar.jsx
adamklepacz: EsLint: PropType object is forbidden

PR #224, opened 2019-07-10
[EOC-261] Socket adding members to a cohort

alemikolo:
**Adding cohort member via sockets was refactored.** After this change: adding, deleting, restoring, list's items via socket stopped working (because of problems with two sockets instances - withSocket and socket instance from file `sockets/index.js`). So I had to change a little bit code not related to my task for the application to continue to work correctly.
I had to remove the use of withSocket hoc from List and Comments components and replaced it with the socket from file `sockets/index.js`. An additional change was related to rooms' names. In new methods: joinRoom and leaveRoom we create room name using Routes enum, where we don't have route list but route sack. So I had to change all the rooms' names from `list-${id}` to `sack-${id}`.
I didn't remove socket's event listeners and other socket s' functionalities related to items and comments. They'll be removed in other tasks related to items.

On adding a new member to cohort 'A':
1. The user who performed action gets data via current logic (fetch data),
2. The members of Cohort 'A' who are currently on this cohort view gets the new member data, 
3. The members of Cohort 'A' who are currently on cohorts view get updated cohort metadata (membersCount),
4. If the new cohort member is on cohorts view gets cohort metadata.
5. If the new cohort member is on dashboard view gets shared lists (metadata) of this cohort
6. Users who are currently on cohort's shared lists views gets new cohort member data.


server/sockets/cohort.js
```diff
@@ -0,0 +1,43 @@
...
+        .then(doc => {
+          if (doc) {
+            const { memberIds } = doc;
+            const cohort = responseWithCohort(doc);
+
+            memberIds.forEach(id => {
+              const memberId = id.toString();
+
+              if (cohortClients.has(memberId)) {
+                socket.broadcast
```
marekrozmus: please check if needed

server/sockets/index.js
```diff
@@ -42,12 +47,36 @@ const socketListenTo = server => {
       return;
     }
 
-    socket.on('joinRoom', room => {
+    socket.on('joinListRoom', room => {
```
marekrozmus: remove braces

server/sockets/cohort.js
alemikolo: Theoretically, it is not needed. All work fine without it. There is a small chance that if somebody adds a new user and will go quickly to the cohorts view or cohort's shared list's view and the backend will perform its duty slowly, they probably receive data via socket also. But even if it happens, it won't cause an error because the received data will overwrite the existing one. So I decided to remove it.

client/src/modules/cohort/components/Cohorts.jsx
```diff
@@ -174,8 +201,10 @@ Cohorts.propTypes = {
   cohorts: PropTypes.objectOf(PropTypes.object),
   currentUser: UserPropType.isRequired,
   intl: IntlPropType.isRequired,
+  socket: PropTypes.objectOf(PropTypes.any),
 
   createCohort: PropTypes.func.isRequired,
+  addCohortToStoreWS: PropTypes.func.isRequired,
```
marekrozmus: alphabetical ordering

alemikolo:
Socket logic (regarding adding cohort member) was refactored

client/src/common/utils/helpers.js
```diff
@@ -17,3 +19,5 @@ export const makeAbortablePromise = promise => {
 };
 
 export const dateFromString = string => new Date(string).toLocaleString();
+
+export const capitalizeString = string => _upperFirst(string);
```
marekrozmus: can't you just use `_upperFirst' where it is needed? Why this alias?

server/sockets/index.js
```diff
@@ -65,6 +98,9 @@ const socketListenTo = server => {
     restoreItemWS(socket);
     updateItemWS(socket);
     addCommentWS(socket);
+    sendListsOnAddCohortMemberWS(socket, dashboardClients);
+
```
adamklepacz: why this empty line?

alemikolo: to separate cohort and list listeners

PR #225, opened 2019-07-11
[EOC-260] [SOCKET] Adding members to a list

alemikolo:
**Adding list member via sockets was refactored.** After this change: adding, deleting, restoring, list's items via socket stopped working (because of problems with two sockets instances - withSocket and socket instance from file `sockets/index.js`). So I had to change a little bit code not related to my task for the application to continue to work correctly.
I had to remove the use of withSocket hoc from List and Comments components and replaced it with the socket from file `sockets/index.js`. An additional change was related to rooms' names. In new methods: joinRoom and leaveRoom we create room name using Routes enum, where we don't have route list but route sack. So I had to change all the rooms' names from `list-${id}` to `sack-${id}`.
I didn't remove socket 's event listeners and other socket s' functionalities related to items and comments. They'll be removed in other tasks related to items.

### DIFF
--- a/client/src/modules/cohort/model/actions.js
+++ b/client/src/modules/cohort/model/actions.js
@@ -94,9 +94,9 @@ const fetchArchivedCohortsMetaDataFailure = errMessage => ({
   payload: errMessage
 });
 
-const addMemberSuccess = (data, cohortId) => ({
+const addMemberSuccess = data => ({
   type: CohortActionTypes.ADD_MEMBER_SUCCESS,
-  payload: { cohortId, data }
+  payload: data
 });
 
 const addMemberFailure = () => ({
@@ -158,9 +158,6 @@ export const createCohort = data => dispatch =>
         data: data.name
       });
     });
-
-export const addCohortToStoreWS = data => dispatch =>
-  dispatch(createCohortSuccess(data));
 
 export const fetchCohortsMetaData = () => dispatch =>
   getData('/api/cohorts/meta-data')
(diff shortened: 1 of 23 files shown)


client/src/modules/cohort/model/actions.js
```diff
@@ -294,6 +294,7 @@ export const addCohortMember = (cohortId, email, socket) => dispatch =>
     .then(json => {
       if (json._id) {
         const data = { cohortId, json };
+
```
marekrozmus: is this needed for this PR?

alemikolo:
Socket logic (regarding adding cohort member) was refactored

server/sockets/index.js
```diff
@@ -60,28 +62,34 @@ const socketListenTo = server => {
...
       socket.join(room);
+      cohortViewClients.set(userId, socket.id);
     });
 
     socket.on('leaveCohortRoom', data => {
-      const { room } = data;
+      const { room, userId } = data;
 
       socket.leave(room);
+      cohortViewClients.set(userId, socket.id);
```
adamklepacz: should be delete on leave?

server/sockets/index.js
```diff
@@ -92,15 +100,16 @@ const socketListenTo = server => {
...
     updateItemState(socket);
     deleteItemWS(socket);
     restoreItemWS(socket);
     updateItemWS(socket);
     addCommentWS(socket);
-    sendListsOnAddCohortMember(socket, dashboardClients);
+    sendListsOnAddCohortMemberWS(socket, dashboardViewClients);
 
-    addCohortMemberWS(socket, cohortsClients);
+    addCohortMemberWS(socket, allCohortsViewClients);
```
adamklepacz: why this empty line in 111?

alemikolo: To separate list's and cohort's listeners

PR #229, opened 2019-07-15
[EOC-286] Refactor sockets architecture for items

adamklepacz:
withHOC is removed cause I don't need it anymore.

adding item, archiving, cloning, restoring, commenting works via new implementation of  WebSocket now.


client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -407,6 +388,15 @@ class ListItem extends PureComponent {
       done,
       isNameEdited
     } = this.state;
+    let isNameBusy;
+    let isEdited;
+
+    if (busy) {
+      const { nameBusy, descriptionBusy } = busy;
+
+      isNameBusy = nameBusy;
```
alemikolo: Can we use nameBusy directly instead of creating new alias?

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -332,21 +326,29 @@ class ListItem extends PureComponent {
...
+      data: { _id: itemId, description, isOrdered, name, busy },
       isMember
     } = this.props;
     const isFieldDisabled = !isMember || isOrdered;
+    let isDescriptionBusy;
+
+    if (busy) {
+      const { descriptionBusy } = busy;
+
+      isDescriptionBusy = descriptionBusy;
```
alemikolo: Can we use descriptionBusy directly instead of creating new alias?

client/src/sockets/enums.js
```diff
@@ -0,0 +1,30 @@
...
+  ARCHIVE_SUCCESS: ItemActionTypes.ARCHIVE_SUCCESS,
+  CLEAR_VOTE_SUCCESS: ItemActionTypes.CLEAR_VOTE_SUCCESS,
+  CLONE_SUCCESS: ItemActionTypes.CLONE_SUCCESS,
+  DELETE_SUCCESS: ItemActionTypes.DELETE_SUCCESS,
+  RESTORE_SUCCESS: ItemActionTypes.RESTORE_SUCCESS,
+  SET_VOTE_SUCCESS: ItemActionTypes.SET_VOTE_SUCCESS,
+  UPDATE_SUCCESS: ItemActionTypes.UPDATE_SUCCESS
+});
+
+export const ItemStatusType = Object.freeze({
```
alemikolo: Do we need separate enum for that? Could we add them to ItemsActionTypes. They are used in the same way as actionTypes?

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -16,10 +16,10 @@ import {
   archiveItem,
   clearVote,
   cloneItem,
-  cloneItemWS,
+  itemBusy,
```
alemikolo: itemBusy doesn't sound like an action/method - maybe setItemBusy, the same for itemFree

client/src/modules/list/model/actions.js
```diff
@@ -12,6 +12,7 @@ import { createNotificationWithTimeout } from 'modules/notification/model/action
 import history from 'common/utils/history';
 import { UserAddingStatus } from 'common/components/Members/const';
 import { ResourceNotFoundException } from 'common/exceptions/ResourceNotFoundException';
+import socketInstance from 'sockets';
```
alemikolo: maybe use name socket instead of socketInstance like in other files

client/src/modules/list/components/Items/ListItem/ListItem.jsx
adamklepacz: how can I get access to `nameBusy` variable outside if block without aliasing?

client/src/modules/list/components/Items/ListItem/ListItem.jsx
adamklepacz: how can I get access to `descriptionBusy` variable outside if block without aliasing?

client/src/sockets/enums.js
adamklepacz: ItemActionTypes describes actions related to the items, like deleting archiving, etc.

BUSY and FREE are not an action, they represent STATE of the item so I don't think that mixing them all together is a good idea.

client/src/modules/list/components/Items/ListItem/ListItem.jsx
marekrozmus: You can always do sth like this:
`const { nameBusy, descriptionBusy } = busy || {}`
Do you need also some deafult values? If yes then:
`const { nameBusy = false, descriptionBusy = false } = busy || {}`

client/src/modules/list/model/reducer.js
```diff
@@ -210,6 +211,8 @@ const lists = (state = {}, action) => {
     case ItemActionTypes.SET_VOTE_SUCCESS:
     case ItemActionTypes.TOGGLE_SUCCESS:
     case ItemActionTypes.UPDATE_SUCCESS:
+    case ItemStatusType.BUSY:
+    case ItemStatusType.FREE:
```
marekrozmus: change to `LOCK`, `UNLOCK`

client/src/locales/en.json
```diff
@@ -152,7 +152,6 @@
   "list.list-item.clone": "Clone item",
   "list.list-item.confirmation": "Do you really want to archive '{name}' item?",
   "list.list-item.author": "Added by: {authorName}",
-  "list.list-item.busy-info": "This item is currently edited by someone.",
```
marekrozmus: no more text about locked item?

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -217,35 +179,60 @@ class ListItem extends PureComponent {
 
   preventDefault = event => event.preventDefault();
 
-  handleNameFocus = () => {
-    this.setState({ isNameEdited: true });
-    this.itemBusy();
+  handleNameBusy = () => {
```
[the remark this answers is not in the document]
adamklepacz: yes, we need to know. 
After discussion with Aleksander we decided that we **shouldn't be blocking the whole item** when the name itself is edited. Another user should be still able to edit the description, add comments at the same time.

client/src/locales/en.json
adamklepacz: no, because currently we are only disabling currently edited field like name, description.

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -40,8 +40,8 @@ class ListItem extends PureComponent {
     this.state = {
       areDetailsVisible: false,
       busy: {
```
alemikolo: consider: lock?

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -179,37 +179,37 @@ class ListItem extends PureComponent {
 
   preventDefault = event => event.preventDefault();
 
-  handleNameBusy = () => {
+  handleNameLock = () => {
```
alemikolo: consider: handleLockName?

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -179,37 +179,37 @@ class ListItem extends PureComponent {
...
     this.setState(
-      { busy: { nameBusy: true, descriptionBusy: false } },
+      { busy: { nameLock: true, descriptionLock: false } },
       this.setItemBusy
     );
   };
 
-  handleNameFree = () => {
+  handleNameUnlock = () => {
```
alemikolo: consider: handleUnlockName

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -218,12 +218,12 @@ class ListItem extends PureComponent {
       }
     } = this.props;
 
-    setItemBusy(itemId, listId, { nameBusy, descriptionBusy });
+    setItemBusy(itemId, listId, { nameLock, descriptionLock });
```
alemikolo: consider: `UnlockItem = () => ...`

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -218,12 +218,12 @@ class ListItem extends PureComponent {
...
   };
 
   setItemFree = () => {
```
alemikolo: consider: `LockItem = () => ...`

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -106,13 +72,13 @@ class ListItem extends PureComponent {
       disableToggleButton: true
     }));
 
-    this.itemBusy();
+    this.setItemBusy();
```
marekrozmus: these also to `lockItem` and `unlockItem`

client/src/modules/list/components/Items/ListItem/ListItem.jsx
adamklepacz: obsłuż zablokowanie nazwy -> handle name lock

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -184,13 +147,12 @@ class ListItem extends PureComponent {
...
-      },
-      socket
+      }
     } = this.props;
 
-    this.itemBusy();
+    this.lockItem();
 
-    return archiveItem(listId, itemId, name, socket);
+    return archiveItem(listId, itemId, name);
```
marekrozmus: no need to unlock item here?

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -41,48 +39,16 @@ class ListItem extends PureComponent {
 
     this.state = {
       areDetailsVisible: false,
-      busyBySomeone: false,
-      busyInfoVisibility: false,
       done: isOrdered,
       isNameEdited: false,
-      isConfirmationVisible: false
+      isConfirmationVisible: false,
+      lock: {
```
marekrozmus: is this needed? `nameLock` and `descriptionLock` are pretty clear

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -39,13 +39,13 @@ class ListItem extends PureComponent {
...
     this.state = {
       areDetailsVisible: false,
-      busy: {
-        nameLock: false,
-        descriptionLock: false
-      },
       done: isOrdered,
       isNameEdited: false,
-      isConfirmationVisible: false
+      isConfirmationVisible: false,
```
alemikolo: alphabetic order

client/src/modules/list/components/Items/ListItem/ListItem.jsx
adamklepacz: nope, because it's unmounted after archive action so there is nothing to unlock.

PR #235, opened 2019-07-16
[EOC-252] voting and toggling to the item

adamklepacz:


client/src/sockets/receiveEvents.js
```diff
@@ -1,4 +1,9 @@
-import { ItemsEvents, ItemStatusType, CommentEvents } from './enums';
+import {
+  ItemsEvents,
+  ItemStatusType,
+  CommentEvents,
```
alemikolo: alphabetic order

server/sockets/index.js
```diff
@@ -67,6 +94,9 @@ const socketListenTo = server => {
     updateItemWS(socket);
     addCommentWS(socket);
     cloneItemWS(socket);
+    setVoteWS(socket);
+    clearVoteWS(socket);
+    markAsDoneWS(socket, dashboardClients);
```
alemikolo: 1. alphabetic order
2. Does this method change item to unhandled also? If so, its name shouldn't be `markAsDoneWS` (changeOrderStateWS, toggleOrderStatus etc... )

client/src/sockets/receiveEvents.js
```diff
@@ -14,5 +19,9 @@ export const receiveEvents = (dispatch, socket) => {
     socket.on(event, data => dispatch({ type: event, payload: data }))
   );
 
+  Object.values(ListEvents).forEach(event =>
+    socket.on(event, data => dispatch({ type: event, payload: data }))
+  );
+
```
marekrozmus: Please create separate methods instead of comments:
`receiveItemEvents` and `receiveListEvents`

PR #236, opened 2019-07-17
[EOC-259] [SOCKET] Leaving cohort

alemikolo:


client/src/modules/list/model/reducer.js
```diff
@@ -199,6 +200,24 @@ const lists = (state = {}, action) => {
     }
     case CohortActionTypes.ARCHIVE_SUCCESS:
       return {};
+    case CohortActionTypes.LEAVE_SUCCESS: {
+      const { payload: leavedCohortId } = action;
+
+      return _keyBy(
+        _filter(state, list => {
+          const { cohortId, isMember, type } = list;
```
adamklepacz: add a line of break

client/src/modules/list/model/reducer.js
```diff
@@ -199,6 +200,24 @@ const lists = (state = {}, action) => {
...
+            cohortId !== leavedCohortId ||
+            type === ListType.LIMITED ||
+            (type === ListType.SHARED && isMember)
+          ) {
+            return list;
+          }
+        }),
+        '_id'
+      );
+    }
```
adamklepacz: can you explain what this reducer is doing?
I can't see any leaving process from members, owners array for the user here.

server/sockets/cohort.js
```diff
@@ -2,6 +2,29 @@ const { CohortActionTypes } = require('../common/variables');
...
+          if (clients.has(memberId)) {
+            socket.broadcast
+              .to(clients.get(memberId))
+              .emit(CohortActionTypes.CREATE_SUCCESS, cohort);
+          }
+        });
+      }
+    });
+};
+
```
adamklepacz: create /helperes directory within sockets directory, and put it in index.js file and import it here 

Zrzut ekranu 2019-07-18 o 07 55 40

server/sockets/list.js
```diff
@@ -168,16 +168,43 @@ const leaveList = socket =>
...
+      },
+      '_id created_at cohortId name description items favIds type'
+    )
+      .lean()
+      .exec()
+      .then(docs => {
+        if (docs) {
+          const sharedListIds = docs.map(list => list._id.toString());
+
+          sharedListIds.forEach(listId => {
```
adamklepacz: unnecessary braces

client/src/modules/list/model/reducer.js
alemikolo: If the user leaves the cohort we have to remove some lists from the store, lists to which the user lost access to by leaving this cohort. (cohort's shared lists where they were a viewer only). And this is what this reducer does - filters the lists to remove those ones which are forbidden for this user now.

PR #240, opened 2019-07-18
[EOC-296] Fix notifications while reseting password

adamklepacz:


server/controllers/userAuth.js
```diff
@@ -201,19 +201,13 @@ const resetPassword = (req, resp, next) => {
...
-      if (idFromProvider) {
-        throw new Error('authorization.actions.reset');
-      }
-
-      if (!isActive) {
-        throw new Error(
-          'authorization.actions.reset-password-not-active-account'
-        );
+      if (idFromProvider || !isActive) {
+        return resp.sendStatus(200);
```
alemikolo: `resp.send()` will be enough

server/controllers/userAuth.js
```diff
@@ -237,15 +231,7 @@ const resetPassword = (req, resp, next) => {
...
-    .catch(err => {
-      const { message } = err;
-
-      if (message) {
-        return resp.status(400).send({ message });
-      }
-
-      resp.sendStatus(400);
-    });
+    .catch(() => resp.sendStatus(400));
```
alemikolo: Shouldn't we always send status OK? (E.g if there is no user also)

adamklepacz: if there is no user we send OK status

PR #241, opened 2019-07-19
[EOC-297] clear redux data on log out reload page

adamklepacz:


client/src/modules/authorization/model/actions.js
```diff
@@ -26,7 +26,10 @@ const loginSuccess = data => ({
 
 export const logoutCurrentUser = () => dispatch =>
   postRequest('/auth/logout')
-    .then(() => dispatch(logoutSuccess()))
+    .then(() => {
+      dispatch(logoutSuccess());
```
marekrozmus: this is not needed anymore if we reload the page

client/src/modules/authorization/model/actions.js
```diff
@@ -26,7 +26,10 @@ const loginSuccess = data => ({
...
+      window.location.reload();
```
marekrozmus: on what page we will be after logging out?

adamklepacz: on login page

PR #242, opened 2019-07-22
[EOC-354] unify margins and paddings

adamklepacz:


client/src/modules/about/About.scss
```diff
@@ -1,15 +1,15 @@
 .about {
   &__intro {
-    margin-top: 5px;
+    margin-top: $s-margin;
     width: 100%;
     background: url('../../assets/images/hero-bg.png') bottom;
     background-size: auto 260px;
     background-repeat: no-repeat;
-    padding: 0 0 100px;
+    padding: 0 0 calc($xxl-padding + $xl-padding + $s-padding);
```
alemikolo: could we use `$xxxl-padding` instead of this padding calculation to have our paddings always the same as padding-constants values? Or if we really need padding different from our constants - some scss function which returns proper padding which will be a multiple of four?

client/src/app/styles/_variables.scss
```diff
@@ -16,3 +16,19 @@ $l-font: 18px;
 $xl-font: 20px;
 $xxl-font: 24px;
 $xxxl-font: 32px;
+
+// margins
+$s-margin: 4px;
+$m-margin: 8px;
+$l-margin: 16px;
+$xl-margin: 32px;
```
marekrozmus: why so huge values, we might rather need some smaller one like 2px

client/src/app/styles/_variables.scss
```diff
@@ -16,3 +16,19 @@ $l-font: 18px;
...
+$xxl-margin: 64px;
+$xxxl-margin: 128px;
+
+// paddings
+$s-padding: 4px;
+$m-padding: 8px;
+$l-padding: 16px;
+$xl-padding: 32px;
```
marekrozmus: same here

client/src/app/styles/_buttons.scss
```diff
@@ -48,7 +48,7 @@
   font-size: $m-font;
   color: $text-normal;
   cursor: pointer;
-  padding: 12px 0 12px 0;
+  padding: $m-padding 0 $m-padding 0;
```
marekrozmus: isn't just `12px 0` enough?

client/src/app/components/Layout/Toolbar/components/UserBar/UserBar.scss
```diff
@@ -118,8 +118,7 @@
     font-size: 12px;
     line-height: 1.2;
     color: $text-normal;
-    margin-right: 8px;
-    margin-bottom: 3px;
+    margin: 0 $m-margin $s-margin 0;
```
marekrozmus: just use `margin-right` and `margin-bottom` - it is more clear

client/src/common/components/Dialog.scss
```diff
@@ -62,10 +62,10 @@
 
   &__button {
     &:first-of-type {
-      margin: 0 0 15px;
+      margin: 0 0 $l-margin;
 
       @include breakpoint($xxsmall) {
-        margin: 0 20px 0 0;
+        margin: 0 $l-margin 0 0;
```
marekrozmus: just use `margin-right`

client/src/modules/list/components/Items/InputBar/InputBar.scss
```diff
@@ -11,6 +11,7 @@
     transform: translateY(-50%);
     width: auto;
     text-transform: uppercase;
+    min-height: 36px;
```
marekrozmus: why not using constant?

client/src/app/styles/_variables.scss
adamklepacz: on a meeting, in the last week, we said we gonna have margins multiplied by 4px so that's why I did like so :)
 so, what margins we should have?

client/src/modules/list/components/Items/InputBar/InputBar.scss
adamklepacz: how would I name this constant and where to place them?
If I should use constant here, should I create constant for every place where I have 37px, 38px, 39px??

client/src/app/styles/_variables.scss
marekrozmus: it is ok but do we need constants for margins like 64 and 128px? How often will we use them?

client/src/app/styles/_buttons.scss
marekrozmus: question still valid - from docs: "When two values are specified, the first padding applies to the top and bottom, the second to the left and right."

client/src/modules/list/components/Items/InputBar/InputBar.scss
marekrozmus: I thought it was margin :) now I see it is height. Ignore it :)

adamklepacz:
@marekrozmus  
with @alemikolo we decided to make a function instead of variables, cause if we would like to have more accurate values of margins and padding we'll have to have: 
```
$xxs-margin: 2px;
$xs-margin: 4px;
$s-margin: 8px;
$m-margin: 12px;
$l-margin: 16px;
$xl-margin: 20px;
$xxl-margin: 24px;
$xxxl-margin: 32px;
$xxxxl-margin: 48px;
$xxxxxl-margin: 64px;
```
what doesnt make sense at all.
So now there is a four-by($multiplier) SASS helper function to manage both paddings and margins.

adamklepacz:
@marekrozmus  ready for review

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.scss
```diff
@@ -21,16 +21,16 @@
...
 
   &__description {
     font-size: $m-font;
     color: $text-normal;
-    padding: 8px 0 10px 0;
+    padding: four-by(2) 0;
     line-height: 1.2;
     text-align: left;
-    margin: 0 0 16px 0;
+    margin: 0 0 four-by(4) 0;
```
jakubbogacz: ```margin-bottom: four-by(4)``` wouldn't be enough?

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.scss
```diff
@@ -53,21 +53,21 @@
     cursor: text;
     border: 0;
     outline: 0;
-    padding: 8px 0 10px 0;
-    margin: 0 0 16px 0;
+    padding: four-by(2) 0;
+    margin: 0 0 four-by(4) 0;
```
jakubbogacz: Again ```margin-bottom: four-by(4)```

PR #247, opened 2019-07-25
[EOC-265] [SOCKET] Archive list

adamklepacz:
This branch includes changes from other branches so review only those 7 files:
```
client/src/modules/list/model/actions.js
client/src/modules/list/index.jsx 
client/src/sockets/enums.js 
client/src/sockets/helpers/index.js 
server/common/variables.js
server/sockets/index.js
server/sockets/list.js   --> archiveList function only!!
```


client/src/sockets/helpers/index.js
```diff
@@ -0,0 +1,48 @@
...
+      );
+    }
+    case ListEvents.REMOVE_MEMBER_SUCCESS: {
+      const { cohortId } = data;
+
+      if (cohortId) {
+        dispatch({ type: event, payload: data });
+        history.replace(`cohort/${cohortId}`);
+
+        return window.location.reload();
```
alemikolo: hmm, if this is good for the user to reload to the window if someone else removes them from the list? What about notification: why the user is redirected?

server/sockets/list.js
```diff
@@ -348,20 +367,193 @@ const emitRemoveMemberOnLeaveCohort = socket =>
...
+
+    List.findById(listId)
+      .populate('cohortId')
+      .lean()
+      .exec()
+      .then(doc => {
+        const { viewersIds } = doc;
+
+        viewersIds.forEach(viewerId => {
+          dashboardClients.forEach((socketId, clientId) => {
```
alemikolo: don't iterate through the dashboardClinets and cohortClients - this map may contain a lot of users. The native map's methods `has()` and `get()` are faster.
For each list viewer, you will iterate through the all dashboardClinets and second time through all clients on every cohort view...

client/src/sockets/helpers/index.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+    case ListEvents.REMOVE_BY_SOMEONE: {
+      const { cohortId, listId } = data;
+
+      if (cohortId) {
+        dispatch({ type: ListEvents.DELETE_SUCCESS, payload: listId });
+
+        return history.replace(`/cohort/${cohortId}`);
+      }
+      dispatch({ type: ListEvents.DELETE_SUCCESS, payload: listId });
```
marekrozmus: new line after block

alemikolo:
```
case ListEvents.REMOVE_BY_SOMEONE: {
      const { cohortId, listId } = data;

      dispatch({ type: ListActionTypes.DELETE_SUCCESS, payload: listId });

      return history.replace(cohortId ? `/cohort/${cohortId}` : '/dashboard');
    }
```

client/src/sockets/helpers/index.js
```diff
@@ -0,0 +1,39 @@
...
+
+      return history.replace('/dashboard');
+    }
+    case ListEvents.ARCHIVE_SUCCESS: {
+      const { cohortId, listId } = data;
+
+      dispatch({ type: ListActionTypes.DELETE_SUCCESS, payload: listId });
+      history.replace(cohortId ? `/cohort/${cohortId}` : '/dashboard');
```
alemikolo:
``` 
return history.replace(cohortId ? `/cohort/${cohortId}` : '/dashboard');
```
and remove return from line 34.

server/sockets/helpers/index.js
```diff
@@ -43,37 +46,35 @@ const updateListOnDashboardAndCohortView = (
...
-              // send to users that are on the dashboard view
-              socket.broadcast
-                .to(dashboardClients.get(viewerId))
-                .emit(ListActionTypes.CREATE_SUCCESS, list);
-            }
-          });
-        }
+          if (dashboardClients.has(viewerId)) {
+            const { socketId } = dashboardClients.get(viewerId);
+            // send to users that are on the dashboard view
```
alemikolo: this comment is unnecessary, we have dashboardClients here so we know that event will be sent to users that are on the dashboard

client/src/modules/list/model/reducer.js
```diff
@@ -92,7 +90,7 @@ const lists = (state = {}, action) => {
     case ListActionTypes.FETCH_ARCHIVED_META_DATA_SUCCESS:
       return { ...state, ...action.payload };
     case ListActionTypes.FETCH_META_DATA_SUCCESS:
-      return { ...state, ...action.payload };
+      return { ...action.payload };
```
alemikolo: Why this change? Many list features emiting via sockets won't work...

server/sockets/list.js
```diff
@@ -348,20 +384,245 @@ const emitRemoveMemberOnLeaveCohort = socket =>
...
+            const { socketId } = dashboardClients.get(id);
+
+            // Broadcast to clients on dashboard that have this list
+            socket.broadcast
+              .to(socketId)
+              .emit(ListActionTypes.DELETE_SUCCESS, listId);
+
+            socket.broadcast
+              .to(socketId)
+              .emit(ListActionTypes.FETCH_ARCHIVED_META_DATA_SUCCESS, {
```
alemikolo: check if emitting those to events is necessary if you emit only the second one you update the old list meta-data with the new meta-data so the list should disappear from because of the selector gets only not archived list...

server/sockets/list.js
```diff
@@ -348,20 +384,245 @@ const emitRemoveMemberOnLeaveCohort = socket =>
...
+            const { socketId } = cohortClients.get(id);
+
+            // Broadcast to clients on cohort view that have this list
+            socket.broadcast
+              .to(socketId)
+              .emit(ListActionTypes.DELETE_SUCCESS, listId);
+
+            socket.broadcast
+              .to(socketId)
+              .emit(ListActionTypes.FETCH_ARCHIVED_META_DATA_SUCCESS, {
```
alemikolo: See above and check if emitting those two events is necessary

server/sockets/list.js
```diff
@@ -371,3 +632,11 @@ module.exports = {
...
   updateListHeaderState
 };
+
+// Dla kazdego kto jest na widoku listy, wysylam event ARCHIVE SUCCESS z danymi, list ID, isGuest, cohortId
+// jesli jest guest, to przekierowuje na dashboard i return
+// jesli nie jest guest to sprawdzam czy istnieje cohortiId i przekierwouje na cohorte lub dahsboard
+
+// Dla kazdego kto jest na widoku dahsboard wysylam DELETE_SUCCESS z listId,
+
+// Dla kazdego kto jest na cohort view, wysylam DELETE_SUCCESS z listID
```
alemikolo: As Marek said: use English and maybe move this from comments to PR description?

client/src/sockets/helpers/index.js
```diff
@@ -25,19 +26,13 @@ export const listEventsController = (event, data, dispatch) => {
...
+      const { cohortId, listId } = data;
 
-      history.replace('/dashboard');
-      window.location.reload();
+      // FIXME: CHECK IF IT WORK
+      dispatch({ type: ListActionTypes.DELETE_SUCCESS, payload: listId });
+      history.replace(cohortId ? `/cohort/${cohortId}` : '/dashboard');
 
-      return dispatch({ type: event, payload: data });
+      return;
```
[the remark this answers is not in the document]
adamklepacz: I'm sorry but I fixed this as you said but didn't notice that git push fails - I have some internet connection problems.....

PR #248, opened 2019-07-25
[EOC-353] create profile view

alemikolo:
- Changed avatars shape to wheels
- Changed user/authorization directories structure to (many changes are associated only with the change of names and imports):
```
   modules
      user
        AutBox
        UserProfile
        model
        index.js
```
- Refactored messages ids in locales en.json file to adjust ids to new directories structure
- Created user profile route and view
- Added changed password functionality available from the user profile view. The form was built in the same way as sign-up and sign-in forms.
- Password dots on profile view are only placeholder - real password is never send to front end :-)
Screenshot 2019-07-25_12-42-10-509
Screenshot 2019-07-25_12-43-10-301
Screenshot 2019-07-25_12-45-43-519
chrome-capture
Screenshot 2019-07-25_12-54-44-688


client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
```diff
@@ -0,0 +1,296 @@
...
+  onPasswordConfirmChange = (confirmPasswordValue, isValid) =>
+    this.setState(
+      { confirmPasswordValue, isPasswordConfirmValid: isValid },
+      this.comparePasswords
+    );
+
+  passwordValidator = value => {
+    const { matches } = validator;
+
+    if (!matches(value, /^[^\s]{4,32}$/)) {
```
marekrozmus: this should be put in some common function as we have same check during password recovery

client/src/modules/user/AuthBox/components/SignInForm.jsx
```diff
@@ -119,7 +118,7 @@ class SignInForm extends PureComponent {
           values={{
             link: (
               <Link className="sign-in__link" to="/reset-password">
-                {formatMessage({ id: 'authorization.forgot-password' })}
+                {formatMessage({ id: 'user.auth.forgot-password' })}
```
marekrozmus: why not using component `<FormattedMessage ...` ?

client/src/modules/user/AuthBox/components/SignInForm.jsx
```diff
@@ -136,11 +135,11 @@ class SignInForm extends PureComponent {
     return (
       <p className="sign-in__forgot-password">
         <FormattedMessage
-          id="authorization.forgot-password.question"
+          id="user.auth.forgot-password.question"
           values={{
             link: (
               <Link className="sign-in__link" to="/reset-password">
-                {formatMessage({ id: 'authorization.forgot-password' })}
+                {formatMessage({ id: 'user.auth.forgot-password' })}
```
marekrozmus: same here

client/src/modules/user/UserProfile/UserProfile.jsx
```diff
@@ -0,0 +1,233 @@
...
+            <span className="user-profile__data-value">
+              {activationDate && (
+                <Fragment>
+                  <FormattedDate
+                    value={activationDate}
+                    year="numeric"
+                    month="long"
+                    day="2-digit"
+                  />
+                  {' at '}
```
marekrozmus: this should be localized string with params: `{date} at {time}`

server/controllers/userAuth.js
```diff
@@ -365,9 +365,63 @@ const resendRecoveryLink = (req, resp, next) => {
     });
 };
 
+const getUserDetails = (req, resp) => {
+  if (req.user) {
+    const { activatedAt, createdAt, email, password } = req.user;
+    const activationDate = activatedAt || createdAt;
+    const isPassword = !!password;
```
marekrozmus: do not double negation operator

server/controllers/userAuth.js
```diff
@@ -365,9 +365,63 @@ const resendRecoveryLink = (req, resp, next) => {
...
+  return resp.sendStatus(204);
+};
+
+const changePassword = (req, resp) => {
+  const { password, newPassword, passwordConfirm } = req.body;
+  const errors = {};
+  const { matches } = validator;
+  const { email } = req.user;
+
+  if (!matches(newPassword, /^[^\s]{4,32}$/)) {
```
marekrozmus: again - same pattern checking

client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
```diff
@@ -0,0 +1,293 @@
...
+          newPasswordError: error
+        },
+        newPassword,
+        isNewPasswordValid: isValid
+      },
+      this.comparePasswords
+    );
+  };
+
+  onPasswordConfirmChange = (confirmPasswordValue, isValid) =>
```
adamklepacz: this method related to new passsword so it should be named 
`onNewPasswordConfirmChange`

client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
```diff
@@ -0,0 +1,293 @@
...
+      this.comparePasswords
+    );
+
+  passwordValidator = value => {
+    if (!validatePassword(value)) {
+      return 'user.auth.input.password.invalid';
+    }
+
+    return '';
+  };
```
adamklepacz: this can be reverted 

```
if(validatePassword(value)) {
   return '';
}

 return 'user.auth.input.password.invalid';
```

client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
```diff
@@ -0,0 +1,293 @@
...
+  comparePasswords = () => {
+    const {
+      higherLevelErrors,
+      higherLevelErrors: { confirmPasswordValueError },
+      newPassword,
+      confirmPasswordValue
+    } = this.state;
+    let newError;
+
+    if (confirmPasswordValueError && newPassword === confirmPasswordValue) {
```
adamklepacz: newConfirmPasswordValue

client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
```diff
@@ -0,0 +1,293 @@
...
+
+    return '';
+  };
+
+  comparePasswords = () => {
+    const {
+      higherLevelErrors,
+      higherLevelErrors: { confirmPasswordValueError },
+      newPassword,
+      confirmPasswordValue
```
adamklepacz: newConfirmPasswordValue

client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
```diff
@@ -0,0 +1,293 @@
...
+    let newError;
+
+    if (confirmPasswordValueError && newPassword === confirmPasswordValue) {
+      newError = '';
+    }
+
+    if (
+      !confirmPasswordValueError &&
+      confirmPasswordValue !== undefined &&
+      newPassword !== confirmPasswordValue
```
adamklepacz: newConfirmPasswordValue

client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
```diff
@@ -0,0 +1,293 @@
...
+  isFormValid = () => {
+    const {
+      higherLevelErrors,
+      isNewPasswordValid,
+      isPasswordConfirmValid,
+      isPasswordValid
+    } = this.state;
+    const isError = _some(higherLevelErrors, error => error !== '');
+
+    return this.setState({
```
adamklepacz: why return while setting state?

client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
```diff
@@ -0,0 +1,293 @@
...
+        !isError
+    });
+  };
+
+  handleChangePassword = event => {
+    event.preventDefault();
+    const { password, newPassword, confirmPasswordValue } = this.state;
+
+    this.setState({ pending: true });
+
```
adamklepacz: remove empty line

client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
```diff
@@ -0,0 +1,293 @@
...
+          if (err instanceof ValidationException) {
+            const { isConfirmPasswordError, isNewPasswordError } = err.errors;
+            newState.higherLevelErrors = {
+              confirmPasswordValueError: isConfirmPasswordError
+                ? 'user.auth.input.password.not-match'
+                : '',
+              newPasswordError: isNewPasswordError
+                ? 'user.auth.input.password.invalid'
+                : ''
+            };
```
adamklepacz: add return here and else will be not needed

client/src/modules/user/UserProfile/UserProfile.jsx
```diff
@@ -0,0 +1,237 @@
...
+      </section>
+    );
+  };
+
+  renderAccountInfo = () => {
+    const {
+      currentUser: { activationDate, isPassword }
+    } = this.props;
+    const { isPasswordUpdateFormVisible } = this.state;
+
```
adamklepacz: remove empty line here

client/src/modules/user/UserProfile/UserProfile.jsx
```diff
@@ -0,0 +1,237 @@
...
+      ? 'user.profile.change-password'
+      : 'user.password';
+
+    return (
+      <section className="user-profile__data-container">
+        <h2 className="user-profile__data-header">
+          <FormattedMessage id="user.profile.account" />
+        </h2>
+        <ul className="user-profile__data-list">
+          {isPassword && (
```
adamklepacz: should be `passwordExists` or `passwordIsSet`

client/src/modules/user/model/actions.js
```diff
@@ -107,3 +116,21 @@ export const updatePassword = (token, password, passwordConfirmation) =>
...
+      createNotificationWithTimeout(dispatch, NotificationType.ERROR, {
+        notificationId: 'user.actions.fetch-user-details-error'
+      });
+    });
+
+export const changePassword = (password, newPassword, passwordConfirm) =>
+  postData('/auth/change-password', {
+    password,
+    newPassword,
+    passwordConfirm
```
adamklepacz: newPasswordConfirm

server/controllers/userAuth.js
```diff
@@ -365,9 +365,62 @@ const resendRecoveryLink = (req, resp, next) => {
     });
 };
 
+const getUserDetails = (req, resp) => {
+  if (req.user) {
```
adamklepacz: destructure user from req;

server/controllers/userAuth.js
```diff
@@ -365,9 +365,62 @@ const resendRecoveryLink = (req, resp, next) => {
...
+    const isPassword = password !== undefined;
+
+    return resp.send({ activationDate, email, isPassword });
+  }
+
+  return resp.sendStatus(204);
+};
+
+const changePassword = (req, resp) => {
+  const { password, newPassword, passwordConfirm } = req.body;
```
adamklepacz: newPasswordConfirm

client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
alemikolo: `if-else` isn't an evil ;-). Here we check if ValidationException or any other exception so I think `if-else` is adequate. But there is a reason that I can't use a `return` here. After this `if-else` we set the state and the `return` in `if`  will break this `catch` and the code after `if-else` won't' be performed.

server/controllers/userAuth.js
```diff
@@ -254,11 +255,10 @@ const recoveryPassword = (req, resp) => {
 
 const updatePassword = (req, resp) => {
   const { password: updatedPassword, passwordConfirmation } = req.body;
-  const { matches } = validator;
   const { token } = req.params;
 
   const sanitizedToken = sanitize(token);
-  const validationError = !matches(updatedPassword, /^[^\s]{4,32}$/);
+  const validationError = !validatePassword(updatedPassword);
   const passwordsNoMatch =
```
marekrozmus: you don't need to match passwords if the password is not validated. You can send response right away.

server/controllers/userAuth.js
```diff
@@ -365,9 +365,64 @@ const resendRecoveryLink = (req, resp, next) => {
...
+  const errors = {};
+  const { email } = req.user;
+
+  if (!validatePassword(newPassword)) {
+    errors.isNewPasswordError = true;
+  }
+
+  if (newPassword !== newPasswordConfirm) {
+    errors.isNewConfirmPasswordError = true;
+  }
```
marekrozmus: could we just change those two `ifs` to
```
errors.isNewPasswordError = !validatePassword(newPassword)
errors.isNewConfirmPasswordError = newPassword !== newPasswordConfirm
```

client/src/modules/user/AuthBox/components/PasswordRecoveryForm.jsx
```diff
@@ -1,15 +1,15 @@
...
 import _flowRight from 'lodash/flowRight';
 import _debounce from 'lodash/debounce';
 import _trim from 'lodash/trim';
 
 import { RouterMatchPropType, IntlPropType } from 'common/constants/propTypes';
-import { updatePassword } from '../model/actions';
+import { updatePassword } from 'modules/user/model/actions';
 import ValidationInput from './ValidationInput';
 import PendingButton from 'common/components/PendingButton';
+import { validatePassword as _validatePassword } from 'common/utils/helpers';
```
marekrozmus: maybe instead of using `_` in name we could name it like: `validatePasswordStructure` or sth similar do see that this helper function is doing only that?

client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
```diff
@@ -0,0 +1,301 @@
...
+      this.comparePasswords
+    );
+
+  passwordValidator = value => {
+    if (validatePasswordStructure(value)) {
+      return '';
+    }
+
+    return 'user.auth.input.password.invalid';
+  };
```
marekrozmus: it can be just ternary:
```javascript
=> validatePasswordStructure(value) ? '' : 'user.auth.input.password.invalid';
```

server/controllers/userAuth.js
```diff
@@ -254,15 +255,15 @@ const recoveryPassword = (req, resp) => {
...
-  const validationError = !matches(updatedPassword, /^[^\s]{4,32}$/);
-  const passwordsNoMatch =
-    _trim(updatedPassword) !== _trim(passwordConfirmation);
 
-  if (validationError || passwordsNoMatch) {
+  if (!validatePasswordStructure(updatedPassword)) {
+    return resp.sendStatus(400);
+  }
+
+  if (_trim(updatedPassword) !== _trim(passwordConfirmation)) {
```
marekrozmus: Such `ifs` can be combined again :) It won't evaluate second || argument if first is true.
So:
```
if (!validatePasswordStructure(updatedPassword) || _trim(updatedPassword) !== _trim(passwordConfirmation))
```

client/src/app/components/Layout/Toolbar/components/UserBar/UserBar.jsx
```diff
@@ -28,18 +29,30 @@ class UserBar extends Component {
...
+      currentUser: { avatarUrl, name }
     } = this.props;
 
     return (
       <ul className="user-bar__menu">
         <li className="user-bar__menu-item">
-          {`Logged as:  ${name}`}
-          <UserIcon />
+          {name}
+          <span>
```
Ciunkos: Is the span really needed here? Does it serve any layout purpose?

server/controllers/userAuth.js
```diff
@@ -365,9 +365,59 @@ const resendRecoveryLink = (req, resp, next) => {
...
+
+  if (user) {
+    const { activatedAt, createdAt, email, password } = user;
+    const activationDate = activatedAt || createdAt;
+    const isPasswordSet = password !== undefined;
+
+    return resp.send({ activationDate, email, isPasswordSet });
+  }
+
+  return resp.sendStatus(204);
```
Ciunkos: Is this by design to return 'No content' instead of 'Not found' (404) and why?

server/controllers/userAuth.js
```diff
@@ -365,9 +365,59 @@ const resendRecoveryLink = (req, resp, next) => {
...
+const changePassword = (req, resp) => {
+  const { password, newPassword, newPasswordConfirm } = req.body;
+  const errors = {};
+  const { email } = req.user;
+
+  errors.isNewPasswordError = !validatePasswordStructure(newPassword);
+  errors.isNewConfirmPasswordError = newPassword !== newPasswordConfirm;
+
+  if (_some(errors, error => error)) {
+    return resp.status(406).send(errors);
```
Ciunkos: 404 - Not Acceptable is used for content negotiation. Consider returning plain 400 Bad request error - https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406

server/controllers/userAuth.js
```diff
@@ -365,9 +365,59 @@ const resendRecoveryLink = (req, resp, next) => {
...
+    const activationDate = activatedAt || createdAt;
+    const isPasswordSet = password !== undefined;
+
+    return resp.send({ activationDate, email, isPasswordSet });
+  }
+
+  return resp.sendStatus(204);
+};
+
+const changePassword = (req, resp) => {
```
Ciunkos: After changing password, as a user, I would expect other sessions (token) to be terminated and logged out. Will this happen here as well?

server/controllers/userAuth.js
```diff
@@ -254,15 +255,14 @@ const recoveryPassword = (req, resp) => {
...
 
   const sanitizedToken = sanitize(token);
-  const validationError = !matches(updatedPassword, /^[^\s]{4,32}$/);
-  const passwordsNoMatch =
-    _trim(updatedPassword) !== _trim(passwordConfirmation);
 
-  if (validationError || passwordsNoMatch) {
+  if (
+    !validatePasswordStructure(updatedPassword) ||
+    _trim(updatedPassword) !== _trim(passwordConfirmation)
```
Ciunkos: Please either trim the input all way up before any processing or do not trim at all and just throw validation error when it contains leading/trailing whitespace.

server/controllers/userAuth.js
```diff
@@ -38,7 +39,7 @@ const signUp = (req, resp, next) => {
     errors.emailError = true;
   }
 
-  if (!matches(password, /^[^\s]{4,32}$/)) {
+  if (!validatePasswordStructure(password)) {
```
Ciunkos: Consider renaming to just `validatePassword`.

client/src/modules/user/model/reducer.js
```diff
@@ -6,6 +6,8 @@ const currentUser = (state = null, action) => {
       return action.payload;
     case AuthorizationActionTypes.LOGOUT_SUCCESS:
       return null;
+    case AuthorizationActionTypes.FETCH_SUCCESS:
+      return { ...state, ...action.payload };
```
Ciunkos: The `state` may be null, please guard against it with `return { ...(state || {}), ...action.payload };`

client/src/modules/user/model/actions.js
```diff
@@ -107,3 +116,21 @@ export const updatePassword = (token, password, passwordConfirmation) =>
 
 export const resendRecoveryLink = token =>
   postData(`/auth/resend-recovery-link/${token}`);
+
+export const fetchUserDetails = userName => dispatch =>
+  getData('/auth/user-details')
```
Ciunkos: What about `getJson = (...args) => getData(...args).then(response => response.json())` to simplify many of the calls which expect JSON output?

client/src/modules/user/model/actions.js
```diff
@@ -19,6 +20,15 @@ const loginSuccess = data => ({
   payload: data
 });
 
+const fetchUserDetailsSuccess = data => ({
```
Ciunkos: Consider renaming the parameter from `data` to `payload` to indicate that is passed without any processing and to use shorthand syntax.

client/src/modules/user/UserProfile/UserProfile.jsx
```diff
@@ -0,0 +1,236 @@
...
+  };
+
+  handlePasswordChangeVisibility = event => {
+    event.preventDefault();
+    this.setState(({ isPasswordUpdateFormVisible }) => ({
+      isPasswordUpdateFormVisible: !isPasswordUpdateFormVisible
+    }));
+  };
+
+  renderHeader = () => {
```
Ciunkos: All these `renderX` can be extracted as separate components that are pure and easy to maintain. They would just accept simple props like `name`, `avatarUrl` etc. It is better to be done now than later.

client/src/modules/user/AuthBox/components/SignUpForm.jsx
```diff
@@ -110,35 +113,16 @@ class SignUpForm extends PureComponent {
       this.comparePasswords
     );
 
-  nameValidator = value => {
-    const { isLength } = validator;
+  nameValidator = value =>
```
Ciunkos: There are some similar functions that work in this way `xValidator = value => validator(value) ? '' : errorMessageId`. Consider this:
```javascript
validateName = validateWith(,
  value => validator.isLength(value, { min: 1, max: 32 })),
  'user.auth.input.email.invalid'
)
```
Where `validateWith = validator => errorMessageId => value => validator(value) ? '' : errorMessageId`. Then you can create a set of validation functions and validators for common fields that can be reused across the app.

client/src/modules/user/AuthBox/components/SignInForm.jsx
```diff
@@ -128,26 +122,20 @@ class SignInForm extends PureComponent {
...
-              <Link className="sign-in__link" to="/reset-password">
-                {formatMessage({ id: 'authorization.forgot-password' })}
-              </Link>
-            )
-          }}
-        />
-      </p>
-    );
-  };
+  renderForgotPassword = () => (
```
Ciunkos: Please see my comment about the extraction of subcomponents.

client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
```diff
@@ -0,0 +1,296 @@
...
+
+  onNewPasswordConfirmChange = (newConfirmPasswordValue, isValid) =>
+    this.setState(
+      { newConfirmPasswordValue, isPasswordConfirmValid: isValid },
+      this.comparePasswords
+    );
+
+  passwordValidator = value =>
+    validatePasswordStructure(value) ? '' : 'user.auth.input.password.invalid';
```
Ciunkos: Please see my comment about common validators.

client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
```diff
@@ -0,0 +1,296 @@
...
+      newPassword: '',
+      password: '',
+      passwordChanged: false,
+      pending: false,
+      changePasswordErrorId: ''
+    };
+  }
+
+  componentDidUpdate() {
+    this.isFormValid();
```
Ciunkos: The "is" prefix indicates that this should be pure and just return a boolean indicating if the form is valid or not. Please rename to `validateForm` or something to clearly indicate that this is an effectful function call.

client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
```diff
@@ -0,0 +1,296 @@
+import React, { Fragment, PureComponent } from 'react';
+import { FormattedMessage, injectIntl } from 'react-intl';
+import PropTypes from 'prop-types';
+import _some from 'lodash/some';
+
+import AuthInput from './AuthInput';
```
Ciunkos: I suggest that relative imports should be at the bottom of the list as they are most likely the most related ones to the code below and the most specific for the particular modules.

client/src/modules/cohort/model/actions.js
```diff
@@ -107,32 +107,32 @@ const removeMemberFailure = () => ({
   type: CohortActionTypes.REMOVE_MEMBER_FAILURE
 });
 
-const removeMemberSuccess = (cohortId, userId) => ({
+const removeMemberSuccess = payload => ({
```
Ciunkos: IMHO, you loose "typings" here - the expected parameters are now only defined in the call places. I would revert that.

client/src/modules/cohort/model/actions.js
```diff
@@ -79,24 +79,24 @@ const fetchCohortDetailsFailure = errMessage => ({
   payload: errMessage
 });
 
-const fetchCohortDetailsSuccess = (data, _id) => ({
+const fetchCohortDetailsSuccess = payload => ({
```
Ciunkos: The same issue with "typings" being lost - please see my comment below.

client/src/locales/en.json
```diff
@@ -67,45 +66,57 @@
...
+  "user.auth.forgot-password": "Forgot password",
+  "user.auth.input.current-password": "Current password",
+  "user.auth.input.username.invalid": "Please use min 1 and max 32 chars.",
+  "user.auth.input.username.empty": "Please enter your name.",
+  "user.auth.input.email.invalid": "Please enter a valid email address.",
+  "user.auth.input.password.confirm": "Confirm password",
+  "user.auth.input.new-password.confirm": "Confirm new password",
+  "user.auth.input.password.invalid": "Please use min 4 and max 32 characters without spaces.",
+  "user.auth.input.password.not-match": "Passwords do not match.",
+  "user.auth.sign-up.result-link": "sign in/sign up page",
```
Ciunkos: Why not a sentence case? :)

client/src/locales/en.json
```diff
@@ -67,45 +66,57 @@
...
+  "user.auth-box.sign-up": "Sign up to app:",
+  "user.auth-box.create-account": "Create account",
+  "user.auth.sign-in": "Sign in",
+  "user.auth.sign-up": "Sign up",
+  "user.auth.change": "Change",
+  "user.auth.sign-up.button.resend-link": "Resend link",
+  "user.auth.forgot-password.question": "If you forgot the password please click {link} to reset it.",
+  "user.auth.forgot-password": "Forgot password",
+  "user.auth.input.current-password": "Current password",
+  "user.auth.input.username.invalid": "Please use min 1 and max 32 chars.",
```
Ciunkos: I think this should be parameterized with `{min}` and `{max}` values.

client/src/locales/en.json
```diff
@@ -67,45 +66,57 @@
...
+  "user.auth.input.username.empty": "Please enter your name.",
+  "user.auth.input.email.invalid": "Please enter a valid email address.",
+  "user.auth.input.password.confirm": "Confirm password",
+  "user.auth.input.new-password.confirm": "Confirm new password",
+  "user.auth.input.password.invalid": "Please use min 4 and max 32 characters without spaces.",
```
Ciunkos: This as well.

client/src/locales/en.json
```diff
@@ -67,45 +66,57 @@
...
+  "user.auth.input.password.not-match": "Passwords do not match.",
+  "user.auth.sign-up.result-link": "sign in/sign up page",
+  "user.auth.reset-password.heading": "Reset your password",
+  "user.auth.reset-password.email-label": "Enter email",
+  "user.auth.reset-password.button-content": "Reset password",
+  "user.auth.pass-recovery-form.heading": "Set new password",
+  "user.actions.fetch-user-details-error": "Failed to fetch your profile data.",
+  "user.actions.logout-fail": "Logout failed.",
+  "user.actions.login": "Login success.",
+  "user.actions.link-expired.recovery-expired": "Password recovery link has expired.",
```
Ciunkos: I would avoid dashes for translation if you even want to use javascript APIs to extract and use them as `user.actions.link-expired.recovery-expired` is not a valid property chain but `user.actions.link_expired.recovery_expired` is. However, you can still use the bracket notation `user.actions['link-expired']['recovery-expired']`.

client/src/common/utils/helpers.js
```diff
@@ -17,3 +17,5 @@ export const makeAbortablePromise = promise => {
 };
 
 export const dateFromString = string => new Date(string).toLocaleString();
+
+export const validatePasswordStructure = value => value.match(/^[^\s]{4,32}$/);
```
Ciunkos: Just `validatePassword` would be OK for me.

server/controllers/userAuth.js
adamklepacz: Added as EOC_406 Issue in Jira

client/src/modules/user/UserProfile/UserProfile.jsx
adamklepacz: It's a lot of work and not a scope of this task so I created a separated task for this that can be found in JIRA under EOC_407

client/src/modules/user/model/actions.js
```diff
@@ -1,13 +1,19 @@
...
 import { MessageType as NotificationType } from 'common/constants/enums';
 import { createNotificationWithTimeout } from 'modules/notification/model/actions';
 import { ValidationException } from 'common/exceptions/ValidationException';
 import history from 'common/utils/history';
 
 export const AuthorizationActionTypes = Object.freeze({
-  LOGIN_FAILURE: 'LOGIN_FAILURE',
-  LOGIN_SUCCESS: 'LOGIN_SUCCESS',
-  LOGOUT_FAILURE: 'LOGOUT_FAILURE'
+  FETCH_SUCCESS: 'user/FETCH_SUCCESS',
```
Ciunkos: Please use the new action type generator functions.

client/src/modules/user/AuthBox/components/SignUpForm.jsx
adamklepacz: Added as a separated task to JIRA under EOC_408 task.

client/src/modules/user/AuthBox/components/SignInForm.jsx
adamklepacz: added to JIRA under EOC_407 task

client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
adamklepacz: added to JIRA under EOC_408 task

server/controllers/userAuth.js
```diff
@@ -365,9 +367,59 @@ const resendRecoveryLink = (req, resp, next) => {
...
+    const activationDate = activatedAt || createdAt;
+    const isPasswordSet = password !== undefined;
+
+    return resp.send({ activationDate, email, isPasswordSet });
+  }
+
+  return resp.sendStatus(400);
+};
+
+const changePassword = (req, resp) => {
```
Ciunkos: Please use `req, res` canonical names. Or full ones, `request` and `response`.

client/src/locales/en.json
adamklepacz: Message id is passed as a externaleErrorId prop to AuthInput component, which is used in many places. I would have to add something like errorParams and pass an object with those params  to use them within AuthInpu component. 
But not every components will have them, so I'll have to make some ugly checking if errorParams object exist etc, then....create some `Lets` to store those values if they exist and pass them down to `formatMessage()` function 

SignUpForm_jsx_—_eoc

AuthInput_jsx_—_eoc


I've tried to do it like above, but somehow it still doesn't work 

**Should I go with this approach??** 
Zrzut ekranu 2019-08-6 o 07 58 26

client/src/modules/user/model/actions.js
adamklepacz: Those actions will be replaced during a merge, cause the generator it on master right now

PR #250, opened 2019-07-26
[EOC-257] Deleting members from cohort

alemikolo:
On delete member from cohort "A":
1. The **user performing** an action gets data via the old (fetch) way
2. **Removed cohort member**:
- if they are on the cohort "A" view they are redirected to cohorts view and cohort "A" is deleted from store
- if they are on cohorts view they lost cohort "A"
- if they are on the dashboard they lost all shared lists in which they were only viewers
- if they are on cohort "A" shared lists that they were removed from they are redirected to the dashboard
- if they are on lists in which they are still viewer (limited or shared in which the were at least a member) the get update prop `isGuest: true`
3. **Cohort members**:
- if they are on cohort "A" view: the cohort member list lost removed member
- if they are on cohorts view they get update cohort "A" meta-data ('membersCount')
4. **Cohort lists members**:
- if they are on shared lists that the user was removed from they lost that user from list's members
- if they were on lists in which the removed user (from cohort) is still a viewer this member is displayed as a Guest


server/sockets/list.js
```diff
@@ -477,6 +477,103 @@ const changeListType = (socket, dashboardClients, cohortClients, listClients) =>
...
+          docs.forEach(doc => {
+            const { type, viewersIds } = doc;
+            const isListViewer = checkIfArrayContainsUserId(viewersIds, userId);
+            const listId = doc._id.toString();
+
+            if (isListViewer) {
+              listIdsUserRemained.push(listId);
+            }
+
+            if (!isListViewer && type === ListType.SHARED) {
```
marekrozmus: this should be in `else if` clause as we are checking opposite condition

client/src/modules/list/model/reducer.js
```diff
@@ -123,6 +136,14 @@ const lists = (state = {}, action) => {
       return { ...state, [action.payload.listId]: action.payload.data };
     case ListActionTypes.REMOVE_ARCHIVED_META_DATA:
       return _keyBy(_filter(state, list => !list.isArchived), '_id');
+    case ListActionTypes.REMOVE_BY_IDS: {
+      const { payload: listToRemoved } = action;
```
adamklepacz: listToRemove

server/common/variables.js
```diff
@@ -100,9 +102,12 @@ const ListActionTypes = Object.freeze({
   FETCH_META_DATA_SUCCESS: 'list/FETCH_META_DATA_SUCCESS',
   LEAVE_ON_TYPE_CHANGE_SUCCESS: 'list/LEAVE_ON_TYPE_CHANGE_SUCCESS',
   LEAVE_SUCCESS: 'list/LEAVE_SUCCESS',
+  REMOVE_BY_IDS: 'list/REMOVE_BY_IDS',
+  REMOVE_BY_SOMEONE: 'list/REMOVE_BY_SOMEONE',
   REMOVE_MEMBER_ROLE_SUCCESS: 'list/REMOVE_MEMBER_ROLE_SUCCESS',
   REMOVE_MEMBER_SUCCESS: 'list/REMOVE_MEMBER_SUCCESS',
   REMOVE_OWNER_ROLE_SUCCESS: 'list/REMOVE_OWNER_ROLE_SUCCESS',
+  MEMBER_UPDATE_SUCCESS: 'list/MEMBER_UPDATE_SUCCESS',
```
adamklepacz: sort

server/sockets/list.js
```diff
@@ -477,6 +477,101 @@ const changeListType = (socket, dashboardClients, cohortClients, listClients) =>
...
+            }
+          });
+
+          if (listIdsUserWasRemovedFrom.length > 0) {
+            if (dashboardClients.has(userId)) {
+              const { socketId } = dashboardClients.get(userId);
+
+              socket.broadcast
+                .to(socketId)
+                .emit(ListActionTypes.REMOVE_BY_IDS, listIdsUserWasRemovedFrom);
```
adamklepacz: this was unclear for me from the very beginning. Here I just figure it out that this action is responsible for removing the user from lists where he was, so it should be named something like ListActionTypes.REMOVE_USER_FROM_LISTS

server/sockets/list.js
```diff
@@ -477,6 +477,101 @@ const changeListType = (socket, dashboardClients, cohortClients, listClients) =>
...
+                  socket.broadcast
+                    .to(socketId)
+                    .emit(ListActionTypes.MEMBER_UPDATE_SUCCESS, {
+                      isCurrentUserUpdated: true,
+                      isGuest: true,
+                      listId,
+                      userId
+                    });
+                }
+              }
```
adamklepacz: line of a break

server/sockets/list.js
alemikolo: This action removes lists. If someone removes you from the cohort and you are on the dashboard you have to lose this cohort's list. This event sends an array of lists' ids which have to be removed from the dashboard, we don't use 'LIST' word in List action types names so I think this action's type can be named as `ListActionTypes.REMOVE_BY_IDS`

server/common/variables.js
```diff
@@ -100,6 +102,9 @@ const ListActionTypes = Object.freeze({
   FETCH_META_DATA_SUCCESS: 'list/FETCH_META_DATA_SUCCESS',
   LEAVE_ON_TYPE_CHANGE_SUCCESS: 'list/LEAVE_ON_TYPE_CHANGE_SUCCESS',
   LEAVE_SUCCESS: 'list/LEAVE_SUCCESS',
+  MEMBER_UPDATE_SUCCESS: 'list/MEMBER_UPDATE_SUCCESS',
+  REMOVE_BY_IDS: 'list/REMOVE_BY_IDS',
+  REMOVE_BY_SOMEONE: 'list/REMOVE_BY_SOMEONE',
```
adamklepacz: shouldn't be REMOVED_BY_SOMEONE ??

client/src/sockets/helpers/index.js
```diff
@@ -12,6 +12,29 @@ export const listEventsController = (event, data, dispatch) => {
         `/${isCohortMember ? `cohort/${cohortId}` : 'dashboard'}`
       );
     }
+    case ListEvents.REMOVED_BY_SOMEONE: {
+      const { cohortId, isCohortMember, listId } = data;
+
+      dispatch({ type: ListEvents.DELETE_SUCCESS, payload: listId });
+
+      return history.replace(
```
Ciunkos: Please split expressions into digestible forms:
```javascript
const dashboardUrl = 'dashboard';

const cohortUrl = `cohort/${cohortId}`;
const goToCohort = cohortId && isCohortMember;
const url = goToCohort ? goToCohort : dashboardUrl;

return history.replace(url);
```

server/sockets/cohort.js
```diff
@@ -138,10 +138,42 @@ const updateCohortHeaderStatus = socket => {
   });
 };
 
+const removeCohortMember = (socket, allCohortsClients, cohortClients) =>
+  socket.on(CohortActionTypes.REMOVE_MEMBER_SUCCESS, data => {
+    const { cohortId, userId } = data;
+
+    socket.broadcast
+      .to(`cohort-${cohortId}`)
```
Ciunkos: Please extract helper functions to reduce duplication and possibility or errors:

```javascript
const cohortChannel = cohortId => `cohort-${cohortId}`;

...

const { cohortId, userId } = data;
const channel = cohortChannel(cohortId);

socket
  .broadcast
  .to(channel)
  .emit(CohortActionTypes.REMOVE_MEMBER_SUCCESS, data);
```

server/sockets/list.js
```diff
@@ -477,6 +477,102 @@ const changeListType = (socket, dashboardClients, cohortClients, listClients) =>
...
+              const { socketId } = dashboardClients.get(userId);
+
+              socket.broadcast
+                .to(socketId)
+                .emit(ListActionTypes.REMOVE_BY_IDS, listIdsUserWasRemovedFrom);
+            }
+
+            listIdsUserWasRemovedFrom.forEach(listId => {
+              socket.broadcast
+                .to(`sack-${listId}`)
```
Ciunkos: Consider using channel url helpers like `listChannel = listId =>`. Please also use consistent naming, even if it differs from the display ones.

server/sockets/list.js
```diff
@@ -477,6 +477,102 @@ const changeListType = (socket, dashboardClients, cohortClients, listClients) =>
...
+          const listIdsUserRemained = [];
+
+          docs.forEach(doc => {
+            const { type, viewersIds } = doc;
+            const isListViewer = checkIfArrayContainsUserId(viewersIds, userId);
+            const listId = doc._id.toString();
+
+            if (isListViewer) {
+              listIdsUserRemained.push(listId);
+            } else if (!isListViewer && type === ListType.SHARED) {
```
Ciunkos: `!isListViewer && ` is redundant is this is always true in this case.

server/sockets/list.js
```diff
@@ -477,6 +477,102 @@ const changeListType = (socket, dashboardClients, cohortClients, listClients) =>
...
+        if (docs) {
+          const {
+            cohortId: { memberIds: cohortMemberIds }
+          } = docs[0];
+          const listIdsUserWasRemovedFrom = [];
+          const listIdsUserRemained = [];
+
+          docs.forEach(doc => {
+            const { type, viewersIds } = doc;
+            const isListViewer = checkIfArrayContainsUserId(viewersIds, userId);
```
Ciunkos: Consider separating isListViewer implementation and importing just that. It would be nice to have only `const listViewer = isListViewer(viewersIds)(userId)` - the internal implementation doesn't matter.

server/sockets/list.js
```diff
@@ -477,6 +477,102 @@ const changeListType = (socket, dashboardClients, cohortClients, listClients) =>
...
+
+    List.find({ cohortId })
+      .populate('cohortId', 'memberIds')
+      .lean()
+      .exec()
+      .then(docs => {
+        if (docs) {
+          const {
+            cohortId: { memberIds: cohortMemberIds }
+          } = docs[0];
```
Ciunkos: Wouldn't `findOne` be a better fit for that?

client/src/sockets/enums.js
```diff
@@ -38,19 +38,24 @@ export const ListEvents = Object.freeze({
...
 
 export const CohortEvents = Object.freeze({
   ADD_MEMBER_SUCCESS: CohortActionTypes.ADD_MEMBER_SUCCESS,
   ADD_OWNER_ROLE_SUCCESS: CohortActionTypes.ADD_OWNER_ROLE_SUCCESS,
+  DELETE_SUCCESS: CohortActionTypes.DELETE_SUCCESS,
   FETCH_META_DATA_SUCCESS: CohortActionTypes.FETCH_META_DATA_SUCCESS,
   LEAVE_SUCCESS: CohortActionTypes.LEAVE_SUCCESS,
   REMOVE_MEMBER_SUCCESS: CohortActionTypes.REMOVE_MEMBER_SUCCESS,
   REMOVE_OWNER_ROLE_SUCCESS: CohortActionTypes.REMOVE_OWNER_ROLE_SUCCESS,
+  REMOVED_BY_SOMEONE: 'cohort/REMOVED_BY_SOMEONE',
```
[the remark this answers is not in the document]
alemikolo: Because it is used only by socket's events, but to be more consist I'll move it to actions' types (it will not hurt them in any way).

client/src/sockets/enums.js
```diff
@@ -38,19 +38,24 @@ export const ListEvents = Object.freeze({
   FETCH_META_DATA_SUCCESS: ListActionTypes.FETCH_META_DATA_SUCCESS,
   LEAVE_ON_TYPE_CHANGE_SUCCESS: 'list/LEAVE_ON_TYPE_CHANGE_SUCCESS',
   LEAVE_SUCCESS: ListActionTypes.LEAVE_SUCCESS,
+  MEMBER_UPDATE_SUCCESS: ListActionTypes.MEMBER_UPDATE_SUCCESS,
+  REMOVE_BY_IDS: ListActionTypes.REMOVE_BY_IDS,
   REMOVE_MEMBER_ROLE_SUCCESS: ListActionTypes.REMOVE_MEMBER_ROLE_SUCCESS,
   REMOVE_MEMBER_SUCCESS: ListActionTypes.REMOVE_MEMBER_SUCCESS,
   REMOVE_OWNER_ROLE_SUCCESS: ListActionTypes.REMOVE_OWNER_ROLE_SUCCESS,
+  REMOVED_BY_SOMEONE: 'list/REMOVED_BY_SOMEONE',
```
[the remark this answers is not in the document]
alemikolo: Because it is used only by socket's events, but to be more consist I'll move it to actions' types (it will not hurt them in any way).

server/sockets/list.js
alemikolo: Unfortunately, in this case, I have to use `find. I have to find all cohort's lists that removed user was a viewer in. But I have to know some cohort data which is populated during querying for lists. But for all cohort's list cohort data is the same so I get it from the first list in a collection.

client/src/modules/list/model/reducer.js
```diff
@@ -73,6 +73,19 @@ const membersReducer = (state = {}, action) => {
         [userId]: { ...state[userId], isOwner: false, isMember: false }
       };
     }
+    case ListActionTypes.MEMBER_UPDATE_SUCCESS: {
+      const {
+        payload: { userId, listId, isCurrentUserUpdated, ...data }
+      } = action;
+      const previousMember = state[userId];
+      const dataToUpdate = _pickBy(data, el => el !== undefined);
```
Ciunkos: The pattern is replicated many times (dataToUpdate few lines below, for example), consider using object filter helper like this:
```javascript
const filter = f => xs => 
  Object.fromEntries(
    Object.entries(xs)
      .filter(([key, value]) => f(value, key, xs))
  );
``` 

Note: Object.fromEntries might need a polyfill, however, reduce-based implementation is not that hard.

Then you can define and reuse common helpers:

```javascript
const defined = x => x !== undefined;
const filterDefined  = filter(defined);

...

const dataToUpdate  = filterDefined(data);
```

client/src/sockets/helpers/index.js
```diff
@@ -1,16 +1,37 @@
-import { ListEvents } from 'sockets/enums';
+import { CohortEvents, ListEvents } from 'sockets/enums';
 import history from 'common/utils/history';
+import { Routes } from 'common/constants/enums';
```
Ciunkos: I think it would be nice to extract helper functions as well from there so they can be used without needing to import both Routes and the routeGenerator. For example, `cohortRoute = id => routeGenerator(Routes.COHORT, id)`.

client/src/sockets/helpers/index.js
```diff
@@ -1,16 +1,37 @@
...
+import { routeGenerator } from 'common/utils/helpers';
 
 export const listEventsController = (event, data, dispatch) => {
```
Ciunkos: Maybe you should consider to keep the redux-like naming and protocol? 🤔 `(action, payload) => dispatch => { ... }`

client/src/sockets/receiveEvents.js
```diff
@@ -27,7 +27,7 @@ export const receiveEvents = (dispatch, socket) => {
   );
 
   Object.values(CohortEvents).forEach(event =>
-    socket.on(event, data => dispatch({ type: event, payload: data }))
+    socket.on(event, data => cohortEventsController(event, data, dispatch))
```
Ciunkos: If you kept the redux-like protocol, the previous call would fit without any changes.

server/common/variables.js
```diff
@@ -75,10 +75,12 @@ const ItemStatusType = Object.freeze({
 const CohortActionTypes = Object.freeze({
```
Ciunkos: Consider using the enumerable helpers for the server types as well.

server/sockets/list.js
Ciunkos: Ok, however, I am still a little bit confused how the cohorts and lists work :)

PR #251, opened 2019-07-29
[EOC-267] Archive cohort

alemikolo:
On archive cohort:
- the user performing an action gets data via old (fetch) way,
- all users on this cohort view are redirected to cohorts view, the cohort is deleted from the store,
- all users on cohorts view lost archived cohort
- all users on the dashboard lost all lists from archived cohort
- users on this cohort's lists' views are redirected to cohorts and lost this list and cohort from store


client/src/modules/list/model/actionTypes.js
```diff
@@ -35,6 +35,7 @@ export const ListActionTypes = Object.freeze({
   LEAVE_FAILURE: 'list/LEAVE_FAILURE',
   LEAVE_SUCCESS: 'list/LEAVE_SUCCESS',
   REMOVE_ARCHIVED_META_DATA: 'list/REMOVE_ARCHIVED_META_DATA',
+  REMOVE_BY_IDS: 'list/REMOVE_BY_IDS',
```
adamklepacz: this is probably the same issue as on the other branch that I commented yesterday
If this action means that user is gonna be removed from many lists this should be named like REMOVE_FROM_LISTS or even REMOVE_FROM_MANY 

Decide which one or give your proposition

server/sockets/cohort.js
```diff
@@ -138,9 +138,43 @@ const updateCohortHeaderStatus = socket => {
   });
 };
 
+const archiveCohort = (socket, allCohortsClients) => {
```
adamklepacz: unnecessary braces

server/sockets/helpers/index.js
```diff
@@ -76,7 +76,30 @@ const updateListOnDashboardAndCohortView = (
...
+  lists.forEach(list => {
+    const { _id, viewersIds } = list;
+    const listId = _id.toString();
+
+    viewersIds.forEach(id => {
+      const viewerId = id.toString();
+
+      if (!listsByViewers[viewerId]) {
+        listsByViewers[viewerId] = [];
+      }
```
adamklepacz: line of a break here

client/src/modules/list/model/actionTypes.js
alemikolo: This event removes lists. If someone removes you from the cohort and you are on the dashboard you have to lose this cohort's list. This event sends an array of lists' ids which have to be removed from the dashboard, we don't use 'LIST' word in List action types names so I think this action's type can be named as ListActionTypes.REMOVE_BY_IDS

server/sockets/list.js
```diff
@@ -477,6 +480,43 @@ const changeListType = (socket, dashboardClients, cohortClients, listClients) =>
...
+      .lean()
+      .exec()
+      .then(docs => {
+        if (docs) {
+          const listIds = docs.map(list => list._id.toString());
+          const listsByViewers = getListsByViewers(docs);
+
+          listIds.forEach(listId => {
+            socket.broadcast
+              .to(`sack-${listId}`)
```
Ciunkos: Please use channel generator functions.

server/sockets/list.js
```diff
@@ -477,6 +480,43 @@ const changeListType = (socket, dashboardClients, cohortClients, listClients) =>
       });
   });
 
+const removeListsOnArchiveCohort = (socket, dashboardClients) =>
+  socket.on(CohortActionTypes.ARCHIVE_SUCCESS, data => {
+    const { cohortId } = data;
+
+    List.find({
```
Ciunkos: I can see the pattern being replicated many times. What about a helper?

```javascript
resolveMultiple => Model => selector => callback =>
  Model
    .find(selector)
    .lean()
    .exec()
    .then(docs => {
      if (docs) {
        const ids= docs.map(({ _id })=> _id.toString());
        callback(docs, ids);
      }
    })

resolveSingle => Model => selector => callback =>  
  Model
    .find(selector)
    .lean()
    .exec()
    .then(docs => {
      if (doc) {
        callback(doc, doc._id.toString());
      }
    })

resolveMultiple(List)({ cohortId })((docs, ids) => {
  const listsByViewers = getListsByViewers(docs);

  ids.forEach(listId => { ... });
});
```

client/src/sockets/enums.js
```diff
@@ -38,15 +38,19 @@ export const ListEvents = Object.freeze({
   FETCH_META_DATA_SUCCESS: ListActionTypes.FETCH_META_DATA_SUCCESS,
   LEAVE_ON_TYPE_CHANGE_SUCCESS: 'list/LEAVE_ON_TYPE_CHANGE_SUCCESS',
   LEAVE_SUCCESS: ListActionTypes.LEAVE_SUCCESS,
+  REMOVE_BY_IDS: ListActionTypes.REMOVE_BY_IDS,
   REMOVE_MEMBER_ROLE_SUCCESS: ListActionTypes.REMOVE_MEMBER_ROLE_SUCCESS,
   REMOVE_MEMBER_SUCCESS: ListActionTypes.REMOVE_MEMBER_SUCCESS,
+  REMOVE_ON_ARCHIVE_COHORT: 'list/REMOVE_ON_ARCHIVE_COHORT',
```

client/src/sockets/enums.js
```diff
@@ -38,15 +38,19 @@ export const ListEvents = Object.freeze({
   FETCH_META_DATA_SUCCESS: ListActionTypes.FETCH_META_DATA_SUCCESS,
   LEAVE_ON_TYPE_CHANGE_SUCCESS: 'list/LEAVE_ON_TYPE_CHANGE_SUCCESS',
```

server/sockets/list.js
alemikolo: Great idea, but I think we should refactor emitting sockets and move emitting events from frontend actions to the backend controllers (we discussed it on Wednesday about the security of our current approach) and in controllers, the sockets will have access to all needed data so additional queries to DB would be unnecessary (as we   do it now). But I save your proposal for future purposes :-)

PR #252, opened 2019-07-29
[EOC-268] Delete cohort

alemikolo:
On delete cohort:
- the user performing an action gets data via old (fetch) way,
- all users on this cohort view are redirected to cohorts view, the cohort is deleted from the store,
- all users on cohorts view lost this cohort

We can delete the cohort only when it is archived. When the cohort is archived there is no way to access its lists so there is no need any other events. Users lost cohort's lists on archiving cohort.


client/src/modules/cohort/model/actions.js
```diff
@@ -233,7 +235,7 @@ export const archiveCohort = (cohortId, cohortName) => dispatch =>
         data: cohortName
       });
       history.replace('/cohorts');
-      socket.emit(CohortActionTypes.ARCHIVE_SUCCESS, { cohortId });
+      // socket.emit(CohortActionTypes.ARCHIVE_SUCCESS, { cohortId });
```
adamklepacz: ??

client/src/sockets/enums.js
```diff
@@ -41,19 +41,19 @@ export const ListEvents = Object.freeze({
   REMOVE_BY_IDS: ListActionTypes.REMOVE_BY_IDS,
   REMOVE_MEMBER_ROLE_SUCCESS: ListActionTypes.REMOVE_MEMBER_ROLE_SUCCESS,
   REMOVE_MEMBER_SUCCESS: ListActionTypes.REMOVE_MEMBER_SUCCESS,
-  REMOVE_ON_ARCHIVE_COHORT: 'list/REMOVE_ON_ARCHIVE_COHORT',
+  REMOVE_WHEN_COHORT_UNAVAILABLE: 'list/REMOVE_WHEN_COHORT_UNAVAILABLE',
```
adamklepacz: what those ''REMOVE_WHEN_COHORT_UNAVAILABLE' actions mean??

server/sockets/cohort.js
```diff
@@ -155,26 +151,25 @@ const archiveCohort = (socket, allCohortsClients) => {
...
-                .emit(CohortActionTypes.DELETE_SUCCESS, cohortId);
-            }
-          });
+          removeCohort(socket, cohortId, allCohortsClients, memberIds);
         }
       });
   });
 };
 
+const deleteCohort = (socket, allCohortsClients) => {
```
adamklepacz: unnecessary braces

client/src/sockets/enums.js
alemikolo: remove the cohort from store when cohort is unavaiable that means when is archved or deleted.(REMOVE_ON_COHORT_DELETION_ARCHIVIZATION?)
DELETE_SUCCES is emitting in many other situations without redirecting so I had to use another action I tried to use ARCHIVED_SUCCESS in deletion case but I look strange. But I agree that this name can be misleading.

client/src/sockets/enums.js
```diff
@@ -41,19 +41,19 @@ export const ListEvents = Object.freeze({
...
 
 export const CohortEvents = Object.freeze({
   ADD_MEMBER_SUCCESS: CohortActionTypes.ADD_MEMBER_SUCCESS,
   ADD_OWNER_ROLE_SUCCESS: CohortActionTypes.ADD_OWNER_ROLE_SUCCESS,
-  ARCHIVE_SUCCESS: CohortActionTypes.ARCHIVE_SUCCESS,
   DELETE_SUCCESS: CohortActionTypes.DELETE_SUCCESS,
   FETCH_META_DATA_SUCCESS: CohortActionTypes.FETCH_META_DATA_SUCCESS,
   LEAVE_SUCCESS: CohortActionTypes.LEAVE_SUCCESS,
   REMOVE_MEMBER_SUCCESS: CohortActionTypes.REMOVE_MEMBER_SUCCESS,
+  REMOVE_WHEN_COHORT_UNAVAILABLE: 'cohort/REMOVE_WHEN_COHORT_UNAVAILABLE',
```
marekrozmus: why not in `CohortActionTypes`?

server/controllers/cohort.js
```diff
@@ -242,7 +245,7 @@ const deleteCohortById = (req, resp) => {
         return resp.sendStatus(400);
       }
 
-      resp.send();
+      resp.send(cohortMembers);
```
Ciunkos: Please see my comment about the cohort details response.

client/src/modules/cohort/model/actions.js
```diff
@@ -203,13 +203,15 @@ export const updateCohort = (cohortName, cohortId, data) => dispatch =>
 
 export const deleteCohort = (cohortId, cohortName) => dispatch =>
   deleteData(`/api/cohorts/${cohortId}/delete`)
-    .then(() => {
+    .then(resp => resp.json())
+    .then(members => {
```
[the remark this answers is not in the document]
alemikolo: I know that it's confusing and unclear. Usually, we don't send data on deletion. But we emit socket events on the frontend. So if the cohort is deleted we want to send some info to the members of this cohort and redirect them from this cohort, etc. But socket event is emitted after the cohort is deleted so all cohort data is lost so we can't get this data and send an event to cohort members because we don't know them. So on cohort deletion, I send cohort's members ids to know who should I send events to. This problem disappears when we move emitting events to controllers and we will not have to do such strange things...

server/controllers/cohort.js
alemikolo: https://github.com/sandstreamdev/eoc/pull/252#discussion_r310105688

PR #253, opened 2019-07-29
[EOC-269] [SOCKET] Adding lists inside cohort

alemikolo:
The newly created shared list is sent to cohort members which are on the cohort's view or on the dashboard.


client/src/modules/list/model/actions.js
```diff
@@ -201,16 +201,20 @@ export const fetchListData = listId => dispatch =>
...
     .then(json => {
       dispatch(createListSuccess(json));
       createNotificationWithTimeout(dispatch, NotificationType.SUCCESS, {
         notificationId: 'list.actions.create-list',
         data: data.name
       });
-      socket.emit(ListActionTypes.CREATE_SUCCESS, { json });
+      if (type === ListType.SHARED) {
+        socket.emit(ListActionTypes.CREATE_SUCCESS, json);
+      }
```
adamklepacz: what about LIMITED lists??

client/src/modules/list/model/actions.js
```diff
@@ -201,16 +201,20 @@ export const fetchListData = listId => dispatch =>
...
+  return postData('/api/lists/create', data)
     .then(resp => resp.json())
     .then(json => {
       dispatch(createListSuccess(json));
       createNotificationWithTimeout(dispatch, NotificationType.SUCCESS, {
         notificationId: 'list.actions.create-list',
         data: data.name
       });
-      socket.emit(ListActionTypes.CREATE_SUCCESS, { json });
+      if (type === ListType.SHARED) {
```
adamklepacz: add a line of a break before if statement

client/src/modules/list/model/actions.js
alemikolo: limited lists are visible only for the creator so we don't have to send it to other users. If we add someone to limited list the will get this list via socket. Only shared lists are visible for all cohort members and need to be sent via sockets.

PR #254, opened 2019-07-30
[EOC-266] Deleting list

adamklepacz:


server/sockets/list.js
```diff
@@ -606,6 +606,31 @@ const archiveList = (socket, dashboardClients, cohortClients, listClients) =>
...
+
+    dashboardClients.forEach(client => {
+      const { socketId } = client;
+
+      socket.broadcast
+        .to(socketId)
+        .emit(ListActionTypes.DELETE_SUCCESS, listId);
+    });
+
+    cohortClients.forEach(client => {
```
alemikolo: this should be done only when this is cohort's list so
```
if (cohortId) {}
```

alemikolo: You iterate through all cohortClients and send this event to all regardless of whether they are members of this list or not. You should send this event to list viewers which are on this cohort view

server/sockets/list.js
```diff
@@ -606,6 +606,31 @@ const archiveList = (socket, dashboardClients, cohortClients, listClients) =>
...
+      const { socketId } = client;
+
+      socket.broadcast
```
alemikolo: Check if the viewer is on this cohort view (viewId === cohortId)

client/src/sockets/enums.js
```diff
@@ -35,6 +35,7 @@ export const ListEvents = Object.freeze({
   ADD_VIEWER_SUCCESS: ListActionTypes.ADD_VIEWER_SUCCESS,
   ARCHIVE_SUCCESS: ListActionTypes.ARCHIVE_SUCCESS,
   CHANGE_TYPE_SUCCESS: ListActionTypes.CHANGE_TYPE_SUCCESS,
+  DELETE_AND_REDIRECT: 'list/DELETE_AND_REDIRECT',
```

adamklepacz: beacuse this is specific to socket and we don't need action type for this.
The same approach is in the line 43 and 45 in this file. 
Check this out

PR #255, opened 2019-07-31
[EOC-380] Restoring cohort

alemikolo:
on restoring cohort:
- te user performing an action gets data via the old (fetch) way
- all cohort members on cohorts view get cohort meta-data
- cohort owners which are on archived cohort view get cohort details and cohorts lists (where they are viewers)
-all viewers of lists from this cohort get those list's meta-data (dashboard)


client/src/modules/cohort/model/reducer.js
```diff
@@ -88,8 +88,14 @@ const cohorts = (state = {}, action) => {
     case CohortActionTypes.REMOVE_ARCHIVED_META_DATA:
       return _keyBy(_filter(state, cohort => !cohort.isArchived), '_id');
     case CohortActionTypes.RESTORE_SUCCESS:
-    case CohortActionTypes.FETCH_DETAILS_SUCCESS:
-      return { ...state, [action.payload._id]: action.payload.data };
+    case CohortActionTypes.FETCH_DETAILS_SUCCESS: {
+      const {
+        payload: data,
+        payload: { _id }
```
marekrozmus: double `payload`?

server/sockets/helpers/index.js
```diff
@@ -116,9 +116,32 @@ const removeCohort = (socket, cohortId, clients, members) => {
...
+    const { _id, viewersIds } = list;
+    const listId = _id.toString();
+
+    viewersIds.forEach(id => {
+      const viewerId = id.toString();
+
+      if (!listsByViewers[viewerId]) {
+        listsByViewers[viewerId] = {};
+      }
+      if (!listsByViewers[viewerId][listId]) {
```
marekrozmus: new line after block

client/src/modules/cohort/model/actions.js
```diff
@@ -275,11 +272,7 @@ export const fetchCohortDetails = cohortId => dispatch =>
   getData(`/api/cohorts/${cohortId}/data`)
     .then(resp => resp.json())
     .then(json => {
```
adamklepacz: uneccessary braces

client/src/modules/cohort/model/reducer.js
```diff
@@ -88,8 +88,14 @@ const cohorts = (state = {}, action) => {
...
+      } = action;
```
adamklepacz: why not just
`const {payload: {data, _id}} = action;`

server/common/utils/helpers.js
```diff
@@ -64,7 +65,8 @@ const responseWithList = (list, userId) => {
   };
 
   if (cohortId) {
-    listToSend.cohortId = cohortId;
+    listToSend.cohortId =
+      typeof cohortId === 'string' ? cohortId : cohortId._id;
```
adamklepacz: can't we just always send id as a string ??

`listToSend.cohortId = cohortId.toString();`

server/common/utils/helpers.test.js
```diff
@@ -400,3 +403,20 @@ describe('function responseWithComment', () => {
     );
   });
 });
+
+describe('function responseWithCohortDetails', () => {
+  const { _id: userId } = usersMock[0];
+  const result = responseWithCohortDetails(cohortDetailsMock, userId);
+
+  it('returns cohorts details with desired properties', () => {
```
adamklepacz: braces not necessary

server/common/utils/helpers.test.js
```diff
@@ -400,3 +403,20 @@ describe('function responseWithComment', () => {
...
+    expectedCohortDetailsProperties.map(property =>
+      expect(result).toHaveProperty(property)
+    );
+  });
+
+  const notExpected = ['favIds', 'ownerIds', 'memberIds'];
+
+  it('returns cohorts details without sensitive data', () => {
```
adamklepacz: check if braces are necessary here

server/sockets/cohort.js
```diff
@@ -163,13 +166,59 @@ const deleteCohort = (socket, allCohortsClients) =>
     removeCohort(socket, cohortId, allCohortsClients, members);
   });
 
+const restoreCohort = (socket, allCohortsClients, cohortClients) =>
+  socket.on(CohortActionTypes.RESTORE_SUCCESS, data => {
+    const { cohortId } = data;
```
adamklepacz: add a line of a break

server/sockets/list.js
```diff
@@ -517,6 +518,50 @@ const removeListsOnArchiveCohort = (socket, dashboardClients) =>
...
+
+    List.find({ cohortId })
+      .populate('cohortId', 'ownerIds')
+      .lean()
+      .exec()
+      .then(docs => {
+        if (docs) {
+          const {
+            cohortId: { ownerIds: cohortOwners }
+          } = docs[0];
```
adamklepacz: maybe use `findOne`, so you don't have to use this [0] index notation.

server/sockets/list.js
```diff
@@ -517,6 +518,50 @@ const removeListsOnArchiveCohort = (socket, dashboardClients) =>
...
+          const listsByUsers = getListsDataByViewers(docs);
+
+          cohortOwners.forEach(id => {
+            const cohortOwnerId = id.toString();
+
+            if (cohortClients.has(cohortOwnerId)) {
+              const { socketId, viewId } = cohortClients.get(cohortOwnerId);
+              if (viewId === cohortId) {
```
adamklepacz: is this check necessary, doesn't the line 539 claims that owner is currently on that archived cohort view?

server/sockets/list.js
```diff
@@ -517,6 +518,50 @@ const removeListsOnArchiveCohort = (socket, dashboardClients) =>
...
+          const {
+            cohortId: { ownerIds: cohortOwners }
+          } = docs[0];
+          const listsByUsers = getListsDataByViewers(docs);
+
+          cohortOwners.forEach(id => {
+            const cohortOwnerId = id.toString();
+
+            if (cohortClients.has(cohortOwnerId)) {
+              const { socketId, viewId } = cohortClients.get(cohortOwnerId);
```
adamklepacz: line o a break after this line

server/tests/__mocks__/cohortsMock.js
```diff
@@ -22,6 +22,33 @@ const cohortsMock = [
     __v: { $numberInt: '0' }
   }
 ];
+const cohortDetailsMock = {
```
adamklepacz: line of a break before this const

server/common/utils/helpers.js
alemikolo: It can't be done in the way you proposed because once cohortId is a string, second time is an object (when you populate cohort data). To use this helper in both situation you have to check cohortId type if it is a string you can assign it to listToSend object directly if not you have to get _id from cohortId object.

server/common/utils/helpers.test.js
alemikolo: I can't remove braces because `it`  must return Promise or undefined

server/common/utils/helpers.test.js
alemikolo: I can't remove braces because it must return Promise or undefined

server/sockets/list.js
alemikolo: when restoring the cohort we have to send all cohorts' list meta-data to the clients, so we have to find all cohorts list and use `find`.
But there is a case that cohort's owners could be on archived cohort view so when the cohort is restoring they should get lists' meta-data also.  But we have to know cohort owners so I populate cohort data an get those owners from the first list (all list will have the same cohort data)

server/sockets/list.js
alemikolo: It is necessary because if the owner of cohort "A" will be on cohort "B" view their id will be stored in `cohortClients` also. But there is no need to send the list for them.

client/src/modules/cohort/model/actions.js
```diff
@@ -54,9 +54,9 @@ const archiveCohortFailure = errMessage => ({
   payload: errMessage
 });
 
-const restoreCohortSuccess = (data, _id) => ({
+const restoreCohortSuccess = data => ({
```
Ciunkos: Consider naming it just a `payload`.

client/src/modules/cohort/model/actions.js
```diff
@@ -79,9 +79,9 @@ const fetchCohortDetailsFailure = errMessage => ({
   payload: errMessage
 });
 
-const fetchCohortDetailsSuccess = (data, _id) => ({
+const fetchCohortDetailsSuccess = data => ({
```
Ciunkos: Here as well.

client/src/modules/cohort/model/selectors.js
```diff
@@ -54,5 +54,5 @@ export const getArchivedCohorts = createSelector(
 
 export const getMembers = createSelector(
   getCohortDetails,
-  cohort => cohort && cohort.members
+  cohort => (cohort && cohort.members) || {}
```
Ciunkos: This can cause unnecessary re-renderings. You can extract `const emptyObject = {}` and then use it as a fallback to avoid constant re-renderings.

server/common/utils/helpers.js
Ciunkos: I think, in this case this should no longer be named `cohortId` when it is truly the `cohort` object but under confusing name. Please try to avoid that situation, maybe send the `cohort` and `cohortId` separately?

server/common/utils/helpers.js
```diff
@@ -285,6 +287,35 @@ const responseWithComments = comments =>
...
+    responseWithCohortMembers(membersCollection, ownerIds),
+    '_id'
+  );
+
+  return {
+    _id,
+    createdAt,
+    description,
+    isArchived,
+    isMember: true,
```
Ciunkos: Is this always true? Maybe you should check if `members` set includes the given `userId` to be sure. Otherwise, rename the function to clearly mark this must be only called by the cohort member `responseWithMembersCohortDetails`, etc..

server/sockets/list.js
Ciunkos: That discussion only assures me there is something wrong with this design. Maybe you need a dedicated function to pull cohorts list metadata? The intent will be much cleaner then.

PR #256, opened 2019-07-31
[EOC-379] Restoring list

adamklepacz:
This Pull Request includes also some changes from EOC-266 Deleting list branch. 

So please review only code related to restoring the list:
client/src/modules/list/model/actions.js   **lines from 320 - 338**
client/src/sockets/enums.js 
client/src/sockets/helpers/index.js **lines from 39 -48** 
server/common/variables.js
server/sockets/index.js 
server/sockets/list.js  - **restoreList function only**


client/src/sockets/enums.js
```diff
@@ -46,6 +46,8 @@ export const ListEvents = Object.freeze({
   REMOVE_MEMBER_ROLE_SUCCESS: ListActionTypes.REMOVE_MEMBER_ROLE_SUCCESS,
   REMOVE_MEMBER_SUCCESS: ListActionTypes.REMOVE_MEMBER_SUCCESS,
   REMOVE_OWNER_ROLE_SUCCESS: ListActionTypes.REMOVE_OWNER_ROLE_SUCCESS,
+  RESTORE_AND_REDIRECT: 'list/RESTORE_AND_REDIRECT',
```

adamklepacz: beacuse this is specific to socket and we don't need action type for this.
The same approach is in the line 43 and 45 in this file.
Check this out

alemikolo: Because it is used only by socket's events, but to be more consist I'll move it to actions' types (it will not hurt them in any way). @marekrozmus which approach should we use?

Ciunkos: Ultimately, I think you should have a single list of actions/events that work across sockets/redux-stores and it doesn't matter where and how they are dispatched.

client/src/sockets/helpers/index.js
```diff
@@ -36,6 +36,16 @@ export const listEventsController = (event, data, dispatch) => {
...
     }
+    case ListEvents.RESTORE_AND_REDIRECT: {
+      const { listId, listData } = data;
+
+      dispatch({
+        type: ListActionTypes.RESTORE_SUCCESS,
+        payload: { data: listData, listId }
+      });
+
+      return history.replace(`/sack/${listId}`);
```
Ciunkos: Please use new route helpers when they become available.

PR #258, opened 2019-07-31
[EOC-381] Activities investigation - concept only for review, not for merge!

adamklepacz:
This branch is not gonna be merged - it's a result of my investigation related to activities module and how to update it via socket. 

So if any action Socket listeners on backend run, we need to check who we should send UPDATE_ACTIVITIES event. 
So for lists - to the viewers and for cohorts - to the members. 
When the user opens activities block/view his ID is saved to ActivitiesClients map, that is how I know that he has this view/block open.  

So if he is a viewer of the list and his activities block is open I'm sending UPDATE_ACTIVITIES event which dispatches an action that updates `shouldUpdate` property in redux store.

Then in ActivitiesList.jsx component I'm checking if this prop was updated if so I'm fetching activities for that user and he gets updated activities entries.


adamklepacz:
- activities pagination - updating should include this feature also   - **this will work with this approach** 

- could this solution be done not only for activity view but for every client who is using the app right now regardless of the view (on login wi will store user id in Map eocClients)? If some action will concern the user, their lists or cohorts, their's activities icon would change somehow? This approach base on your proposal but concern all users not only those who have activities opened. This could also solve the problem of possible notifications in the application  - **this is a good idea but this is a bigger think and could be implemented in a separated task and of course it will work with current approach but can not replace current approach, beacuse if we want to store eocClients and show them red notification that something was changed we shouldn't send them data about noficiation before they actually open activities box so this implementation gonna be whole new task**

adamklepacz:
Okay I'm closing this for now

PR #259, opened 2019-08-01
[EOC-388] Remove lists and cohorts meta data from store on view change

adamklepacz:


client/src/modules/cohort/components/Cohorts.jsx
```diff
@@ -49,9 +50,11 @@ class Cohorts extends Component {
 
   componentWillUnmount() {
     const {
+      clearMetaData,
       currentUser: { id: userId }
     } = this.props;
 
+    clearMetaData();
     leaveView(Routes.COHORTS, userId);
```
Ciunkos: Maybe this should be the responsibility of `leaveView` to clear the related metadata? This way, you are most likely to not forget to clear the data. By using `willUnmount` it is easy to miss some places where the cleanup should be called.

adamklepacz: but those two functions are not related in any way. 
`clearMetaData()` is related to the redux store and cleans up cohorts metadata,
`leaveView()` is related to SOCKETS, to distinguish if the user is one the current view.

so they both responsible for two totally different things.  
Merging them somehow together will be against single responsibility principle where the function should manage only one thing.

Ciunkos: I must oppose as this is, at least for me, clearly an effect of leaving the view. After entering the view, there is some metadata fetch, then it is cleared after leaving. I think the `leaveView` should be composed with these two things to keep the SRP and still be happy. Better yet, the metadata should be cleared while reducing LEAVE_VIEW action that is also dispatched through the sockets.

```javascript
const leaveView = () => {
   socket.dispatch(LEAVE_VIEW, { route, userId });
   store.dispatch(LEAVE_VIEW / CLEAR_METADATA);
}
```

adamklepacz: Okay, I've created a separate task in Jira to refactor that in coming sprints :)

PR #260, opened 2019-08-01
[EOC-383] Unlock editing cohort after a timeout 

adamklepacz:


client/src/modules/cohort/model/actions.js
```diff
@@ -410,16 +410,26 @@ export const leaveCohort = (cohortId, userId, userName) => dispatch =>
...
+    userId
   });
 
-export const unlockCohortHeader = (cohortId, { nameLock, descriptionLock }) =>
+export const unlockCohortHeader = (
+  cohortId,
+  userId,
+  { nameLock, descriptionLock }
+) =>
   socket.emit(CohortHeaderStatusTypes.UNLOCK, {
```
Ciunkos: Consider using a single locking mechanism that locks particular `resourceId` and is handled in a single place. Then, this can be used to handle name and description locks of multiple things like cohorts and list in a coherent way.

server/sockets/cohort.js
```diff
@@ -121,21 +121,47 @@ const updateCohort = (socket, allCohortsViewClients) => {
...
+      if (descriptionLock !== undefined) {
+        updatedData.descriptionLock = false;
+      }
+
+      socket.broadcast
+        .to(`cohort-${cohortId}`)
+        .emit(CohortHeaderStatusTypes.UNLOCK, updatedData);
+
+      cohortClientLocks.delete(userId);
+    }, 300000);
```
Ciunkos: Please extract `LOCK_TIMEOUT` constant.

client/src/modules/cohort/model/actions.js
adamklepacz: Saved to a separated task in Jira for coming sprints

PR #261, opened 2019-08-01
[EOC-389] Unlock editing item after a timeout

adamklepacz:


server/sockets/list.js
```diff
@@ -49,17 +49,43 @@ const restoreItem = socket => {
...
+      if (descriptionLock !== undefined) {
+        updatedData.descriptionLock = false;
+      }
+
+      socket.broadcast
+        .to(`sack-${listId}`)
+        .emit(ItemStatusType.UNLOCK, updatedData);
+
+      itemClientLocks.delete(userId);
+    }, 300000);
```
Ciunkos: Please use the new `LOCK_TIMEOUT` constant when merging.

PR #262, opened 2019-08-01
[EOC-378] Wrap long item name inside notification box 

adamklepacz:
**BUG:**
Zrzut ekranu 2019-07-24 o 12 41 02

**FIXED TO:** 
Zrzut ekranu 2019-08-1 o 19 48 58

### DIFF
--- a/client/src/common/components/MessageBox.jsx
+++ b/client/src/common/components/MessageBox.jsx
@@ -6,7 +6,7 @@ import { MessageType } from 'common/constants/enums';
 import { MessagePropType } from 'common/constants/propTypes';
 import { InfoIcon } from 'assets/images/icons';
 
-const MessageBox = ({ message, type }) => (
+const MessageBox = ({ message, type, children }) => (
   <div
     className={classNames('message-box', {
       'message-box--green': type === MessageType.SUCCESS,
@@ -15,14 +15,18 @@ const MessageBox = ({ message, type }) => (
       'message-box--gray': type === MessageType.INFO
     })}
   >
-    <span>
+    <span className="message-box__icon">
       <InfoIcon />
     </span>
-    <p className="message-box__content">{message}</p>
+    <p className="message-box__content">
+      {message}
+      {children}
+    </p>
   </div>
 );
 
 MessageBox.propTypes = {
+  children: PropTypes.node,
   message: PropTypes.string,
   type: MessagePropType.isRequired
 };
--- a/client/src/common/components/MessageBox.jsx
+++ b/client/src/common/components/MessageBox.jsx
@@ -18,10 +18,7 @@ const MessageBox = ({ message, type, children }) => (
     <span className="message-box__icon">
       <InfoIcon />
     </span>
-    <p className="message-box__content">
-      {message}
-      {children}
-    </p>
+    <p className="message-box__content">{(message, children)}</p>
   </div>
 );
 
--- a/client/src/common/components/MessageBox.scss
+++ b/client/src/common/components/MessageBox.scss
@@ -16,6 +16,19 @@
   line-height: 1.4;
 
   span {
+    display: flex;
+    align-items: center;
+  }
+
+  strong {
+    display: inline-block;
+    white-space: nowrap;
+    max-width: 200px;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+
+  &__icon {
     position: relative;
     display: block;
     display: flex;
@@ -30,7 +43,7 @@
   }
 
   &--red {
-    span {
+    .message-box__icon {
       background-color: $message-bg-red;
       border-color: $message-bg-red;
     }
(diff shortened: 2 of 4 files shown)


client/src/common/components/MessageBox.jsx
```diff
@@ -15,14 +15,15 @@ const MessageBox = ({ message, type }) => (
       'message-box--gray': type === MessageType.INFO
     })}
   >
-    <span>
+    <span className="message-box__icon">
       <InfoIcon />
     </span>
-    <p className="message-box__content">{message}</p>
+    <p className="message-box__content">{(message, children)}</p>
```
Ciunkos: Please use `message || children` as the comma operator can be tricky to understand. If the component is used only in the single place below, you can drop the `message` parameter altogether.

client/src/common/components/MessageBox.jsx
```diff
@@ -15,14 +15,15 @@ const MessageBox = ({ message, type }) => (
...
   </div>
 );
 
 MessageBox.propTypes = {
+  children: PropTypes.node,
```
Ciunkos: Can this be a string as well?

Ciunkos: Please see:
https://flow.org/en/docs/react/children/
https://react-cn.github.io/react/tips/children-props-type.html

I would go for `any` 😉

adamklepacz: But what about this warning??
Zrzut ekranu 2019-08-2 o 12 09 49

Ciunkos: Either disable the rule or reduce the valid prop types to:

```javascript
PropTypes.oneOfType([
  PropTypes.node.isRequired,
  PropTypes.string.isRequired,
  PropTypes.number.isRequired
])
```

client/src/common/components/MessageBox.jsx
```diff
@@ -6,7 +6,7 @@ import { MessageType } from 'common/constants/enums';
 import { MessagePropType } from 'common/constants/propTypes';
 import { InfoIcon } from 'assets/images/icons';
 
-const MessageBox = ({ message, type }) => (
+const MessageBox = ({ message, type, children }) => (
```
Ciunkos: I think you should drop `message` prop and always use the `children`. This requires changes across all usages. Also, please consider sorting the props.

client/src/modules/list/components/Items/ItemsList.jsx
```diff
@@ -78,24 +77,19 @@ class ItemsList extends PureComponent {
...
-            message={
-              archived
-                ? formatMessage({ id: 'list.items-list.message-no-arch-items' })
-                : formatMessage({ id: 'list.items-list.message-no-items' })
-            }
-            type={MessageType.INFO}
-          />
+          <MessageBox type={MessageType.INFO}>
+            {archived ? (
+              <FormattedMessage id="list.items-list.message-no-arch-items" />
```
Ciunkos: In this case, I would vary the `id` prop, not the whole components.

client/src/modules/list/components/Items/ItemsList.jsx
```diff
@@ -121,9 +115,8 @@ class ItemsList extends PureComponent {
 
 ItemsList.propTypes = {
   archived: PropTypes.bool,
-  intl: IntlPropType.isRequired,
   isMember: PropTypes.bool,
   items: PropTypes.arrayOf(PropTypes.object)
 };
 
-export default _flowRight(withRouter, injectIntl)(ItemsList);
+export default _flowRight(withRouter)(ItemsList);
```
Ciunkos: No need for flowRight/pipe anymore. Can be safely removed.

client/src/modules/list/components/Items/ItemsList.jsx
```diff
@@ -78,24 +77,19 @@ class ItemsList extends PureComponent {
...
-            }
-            type={MessageType.INFO}
-          />
+          <MessageBox type={MessageType.INFO}>
+            {archived ? (
+              <FormattedMessage id="list.items-list.message-no-arch-items" />
+            ) : (
+              <FormattedMessage id="list.items-list.message-no-items" />
+            )}
+          </MessageBox>
```
marekrozmus: As Przemo mentioned please use:
```javascript
<FormattedMessage id={archived ? list.items-list.message-no-arch-items : list.items-list.message-no-items } />
```

client/src/modules/list/components/Items/ItemsList.jsx
```diff
@@ -78,24 +76,21 @@ class ItemsList extends PureComponent {
...
-              archived
-                ? formatMessage({ id: 'list.items-list.message-no-arch-items' })
-                : formatMessage({ id: 'list.items-list.message-no-items' })
-            }
-            type={MessageType.INFO}
-          />
+          <MessageBox type={MessageType.INFO}>
+            <FormattedMessage
+              id={
+                archived
```
Ciunkos: This can be extracted before the return as `const emptyMessageId = archived ? 'list.items-list.message-no-arch-items' : 'list.items-list.message-no-items';` to reduce the nesting depth.

PR #263, opened 2019-08-02
[EOC-306] Add more descriptive information to archived item for member

adamklepacz:
Zrzut ekranu 2019-08-2 o 13 48 19


client/src/common/components/ArchivedMessage.scss
```diff
@@ -7,6 +7,7 @@
   background: $bg-color;
   padding: four-by(4) four-by(8);
   box-shadow: 0 1px 3px rgba($shadow-color, 0.15);
+  max-width: 465px;
```
Ciunkos: How was the value chosen? Does it follow your design or it is a random value that looks ok in this case?

client/src/locales/en.json
```diff
@@ -10,7 +10,8 @@
   "app.toolbar.go-to-dashboard": "Go back to dashboard",
   "app.toolbar.notification-tooltip": "See notifications",
   "common.archived-message.restoring": "Restoring '{name}' {item}...",
-  "common.archived-message.was-archived": "The '{name}' {item} was archived.",
+  "common.archived-message.was-archived": "The '{name}' {item} was archived. ",
+  "common.archived-message.was-archived-member": "The '{name}' {item} was archived. To get access to the data please contact the owner to restore '{name}' {item}.",
```
Ciunkos: `... restore the '{name}' ...`

Ciunkos: Please split out the common prefix, then you can use separate messages and display the additional message if needed. By combining and duplicating locales, the translation costs increase.

client/src/common/components/ArchivedMessage.jsx
```diff
@@ -55,7 +55,11 @@ class ArchivedMessage extends PureComponent {
               />
             ) : (
               <FormattedMessage
```
Ciunkos: Please split out the common message and the additional content. Please try this:

```javascript
const successMessage = (
  <>
    <FormattedMessage id="common.archived-message.was-archived" values={{ name, item }} />
    {!isOwner&& <FormattedMessage id="common.archived-message.was-archived-member" values={{ name, item }} />}
  </>
);
```

client/src/common/components/ArchivedMessage.scss
adamklepacz: A random value that looks okay in this case, we don't have any design rules to follow. 

The whole design will be re-implemented anyway in coming sprints.

Ciunkos: Ok

PR #264, opened 2019-08-06
[EOC-400] To big avatar placeholder in user bar dropdown menu

adamklepacz:


client/src/common/components/Avatar.scss
```diff
@@ -14,8 +14,7 @@
   }
 
   svg {
-    width: 100%;
-    height: 100%;
+    height: 16px;
```
Ciunkos: If constant dimensions are expected I ten to enforce them all:
```css
width: 16px;
min-width: 16px;
max-width: 16px;
height: 16px;
min-height: 16px;
max-height: 16px;
flex-shrink: 0;
```

You can create a SASS helper for this.

client/src/common/components/Avatar.scss
```diff
@@ -14,7 +14,7 @@
   }
 
   svg {
-    height: 16px;
+    @include constant-height(16px);
```
Ciunkos: Please rename to `size` or `constant-size`.

PR #265, opened 2019-08-06
[EOC-309] Item - do not delete, flag as deleted

adamklepacz:
Item is now flagged as isDeleted and not removed from the database.

`deleteItem() ` controller is now unnecessary because deleting item is managed via updateItem controller.

### DIFF
--- a/client/src/modules/list/components/Items/model/actions.js
+++ b/client/src/modules/list/components/Items/model/actions.js
@@ -322,7 +322,8 @@ export const fetchArchivedItems = (listId, listName) => dispatch =>
   getJson(`/api/lists/${listId}/archived-items`)
     .then(json => {
       const dataMap = _keyBy(json, '_id');
-      dispatch(fetchArchivedItemsSuccess({ listId, dataMap }));
+
+      dispatch(fetchArchivedItemsSuccess({ listId, data: dataMap }));
     })
     .catch(() => {
       dispatch(fetchArchivedItemsFailure());
@@ -357,7 +358,10 @@ export const restoreItem = (listId, itemId, name) => dispatch =>
     });
 
 export const deleteItem = (listId, itemId, name) => dispatch =>
-  patchData(`/api/lists/${listId}/delete-item/${itemId}`)
+  patchData(`/api/lists/${listId}/update-item`, {
+    isDeleted: true,
+    itemId
+  })
     .then(() => {
       const action = deleteItemSuccess({ listId, itemId });
       const { type, payload } = action;
--- a/client/src/modules/list/components/Items/model/actions.js
+++ b/client/src/modules/list/components/Items/model/actions.js
@@ -1,6 +1,11 @@
 import _keyBy from 'lodash/keyBy';
 
-import { getJson, patchData, postData } from 'common/utils/fetchMethods';
+import {
+  deleteData,
+  getJson,
+  patchData,
+  postData
+} from 'common/utils/fetchMethods';
 import {
   CommentActionTypes,
   ItemActionTypes,
@@ -358,10 +363,7 @@ export const restoreItem = (listId, itemId, name) => dispatch =>
     });
 
 export const deleteItem = (listId, itemId, name) => dispatch =>
-  patchData(`/api/lists/${listId}/update-item`, {
-    isDeleted: true,
-    itemId
-  })
+  deleteData(`/api/lists/${listId}/delete-item/${itemId}`)
     .then(() => {
       const action = deleteItemSuccess({ listId, itemId });
       const { type, payload } = action;
--- a/client/src/modules/list/components/Items/model/actions.js
+++ b/client/src/modules/list/components/Items/model/actions.js
@@ -321,9 +321,9 @@ export const archiveItem = (listId, itemId, name) => dispatch =>
 export const fetchArchivedItems = (listId, listName) => dispatch =>
   getJson(`/api/lists/${listId}/archived-items`)
     .then(json => {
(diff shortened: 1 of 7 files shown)


client/src/modules/list/components/Items/model/actions.js
```diff
@@ -322,7 +322,8 @@ export const fetchArchivedItems = (listId, listName) => dispatch =>
   getJson(`/api/lists/${listId}/archived-items`)
     .then(json => {
       const dataMap = _keyBy(json, '_id');
```
Ciunkos: Why not just name it `data`?

server/controllers/list.js
```diff
@@ -1030,14 +1031,19 @@ const updateListItem = (req, resp) => {
           : ActivityType.ITEM_RESTORE;
       }
 
+      if (isDeleted !== undefined) {
```
Ciunkos: What about `isArchived: true, isDeleted: false`? Then the `editedItemActivity` would be `ActivityType.ITEM_DELETE`. Is that expected?

server/controllers/list.js
```diff
@@ -1210,49 +1216,15 @@ const getArchivedItems = (req, resp) => {
       }
 
       const { items } = list;
-      const archivedItems = items.filter(item => item.isArchived);
+      const archivedItems = items.filter(
```
Ciunkos: Please split filter to `liveItems = items.filter(({ isDeleted }) => !isDeleted)`, then `archivedItems = liveItems.filter(({ isArchived }) => isArchived)`

server/controllers/list.js
adamklepacz: When `isArchived` is true we are archiving, when false we are restoring. So here it is expected.
```
        itemToUpdate.isArchived = isArchived;
        editedItemActivity = isArchived
          ? ActivityType.ITEM_ARCHIVE
          : ActivityType.ITEM_RESTORE;
      }
```


For `isDeleted` case, you're right it can be done like below because there will be no false state for 'restoring from deleted'. 
```
  if (isDeleted) {
        itemToUpdate.isDeleted = isDeleted;
        editedItemActivity = ActivityType.ITEM_DELETE;
   }
```

Ciunkos: As this can not be undone?, I think you should expose `DELETE` endpoint to handle this case separately.

adamklepacz: Deleting item can not be undone 

I've prepared separated endpoint for this. Check next commit, please :)

client/src/modules/list/components/Items/model/actions.js
```diff
@@ -357,7 +363,7 @@ export const restoreItem = (listId, itemId, name) => dispatch =>
     });
 
 export const deleteItem = (listId, itemId, name) => dispatch =>
-  patchData(`/api/lists/${listId}/delete-item/${itemId}`)
+  deleteData(`/api/lists/${listId}/delete-item/${itemId}`)
```
Ciunkos: In order to be more REST-ful, you should not include action names in the path. For example, proper path in this case would be `DELETE /api/lists/${listId}/items/${itemId}` or `DELETE /api/lists/${listId}/${itemId}`.

server/controllers/list.js
```diff
@@ -1210,47 +1210,51 @@ const getArchivedItems = (req, resp) => {
...
+    'items._id': sanitizedItemId,
+    memberIds: userId
+  })
     .exec()
-    .then(list => {
-      if (!list) {
-        return resp.sendStatus(400);
+    .then(doc => {
+      if (!doc) {
+        throw new BadRequestException();
```
Ciunkos: Maybe `NotFoundException`? 404 would be better in this case.

server/controllers/list.js
```diff
@@ -1210,47 +1210,51 @@ const getArchivedItems = (req, resp) => {
       }
 
       const { items } = list;
-      const archivedItems = items.filter(item => item.isArchived);
+      const activeItems = items.filter(item => !item.isDeleted);
+      const archivedItems = activeItems.filter(item => item.isArchived);
```
marekrozmus: why not just one filter:
`const archivedItems = items.filter(item => item.isArchived && !item.isDeleted)`
?

adamklepacz: There was one, Przemo said "Split filters"  😄
https://github.com/sandstreamdev/eoc/pull/265#discussion_r310995105

server/controllers/list.js
```diff
@@ -1034,10 +1034,10 @@ const updateListItem = (req, resp) => {
     })
     .then(doc => {
       if (!doc) {
-        return resp.sendStatus(400);
+        return res.sendStatus(400);
```
Ciunkos: How about 404 not found?

server/common/utils/helpers.test.js
```diff
@@ -139,10 +142,10 @@ describe('function responseWithItem', () => {
     );
   });
 
-  const notExpected = 'voterIds';
+  const notExpected = ['voterIds', 'isDeleted'];
```
Ciunkos: Consider sorting this.

server/common/utils/helpers.test.js
```diff
@@ -122,9 +122,12 @@ describe('function responseWithItems ', () => {
     );
   });
 
-  const notExpected = 'voterIds';
-  it('returns items without sensitive data', () => {
-    result.map(item => expect(item).not.toHaveProperty(notExpected));
+  const notExpected = ['voterIds', 'isDeleted'];
```
Ciunkos: And this as well.

PR #266, opened 2019-08-06
[EOC-310] List do not delete flag as deleted

adamklepacz:


server/controllers/list.js
```diff
@@ -253,7 +218,8 @@ const getListData = (req, resp) => {
 
   List.findOne({
     _id: sanitizedListId,
-    viewersIds: userId
+    viewersIds: userId,
+    isDeleted: false
```
Ciunkos: Please sort 😉

server/controllers/list.js
```diff
@@ -487,6 +454,12 @@ const updateListById = (req, resp) => {
           : ActivityType.LIST_RESTORE;
       }
 
+      if (isDeleted) {
+        listActivity = ActivityType.LIST_DELETE;
+
+        Comment.updateMany({ listId: sanitizedListId }, { isDeleted }).exec();
```
Ciunkos: Where is the `listActivity` mode assigned?

server/common/utils/helpers.test.js
```diff
@@ -103,9 +103,9 @@ describe('function responseWithListsMetaData', () => {
     );
   });
 
-  const notExpected = ['favIds', 'items'];
+  const notExpected = ['favIds', 'items', 'isDeleted'];
```
Ciunkos: Consider sorting this as well.

PR #268, opened 2019-08-07
[EOC-BUG] Add longer pingTimeout to prevent from disconnection

adamklepacz:
This error causes those issues with Sockets that disconnects randomly on Chrome. 

Zrzut ekranu 2019-08-7 o 12 21 53

https://github.com/socketio/socket.io/issues/3259#issuecomment-448058937

Referring to a comment linked above, socket.io library changed default ping timeout from 60000ms in (v2.0.4) to 5000ms in (v2.1.0+), which is not enough for some browsers and they get disconnected randomly. 

Overwriting this parameter fixed the issue, at least on localhost 😄

### DIFF
--- a/server/sockets/index.js
+++ b/server/sockets/index.js
@@ -55,7 +55,7 @@ const {
 } = require('./cohort');
 
 const socketListenTo = server => {
-  const ioInstance = io(server, { forceNew: true });
+  const ioInstance = io(server, { forceNew: true, pingTimeout: 60000 });
 
   ioInstance.use(
     passportSocketIo.authorize({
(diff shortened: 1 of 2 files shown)


server/sockets/index.js
```diff
@@ -55,7 +55,7 @@ const {
 } = require('./cohort');
 
 const socketListenTo = server => {
-  const ioInstance = io(server, { forceNew: true });
+  const ioInstance = io(server, { forceNew: true, pingTimeout: 60000 });
```
Ciunkos: Please extract this as `SOCKET_TIMEOUT_MS = 60000;`

PR #269, opened 2019-08-07
[EOC-311] Cohort do not delete flag as deleted

adamklepacz:


server/common/utils/helpers.test.js
```diff
@@ -155,7 +155,7 @@ describe('function responseWithCohorts', () => {
     );
   });
 
-  const notExpected = ['favIds', 'ownerIds', 'memberIds'];
+  const notExpected = ['favIds', 'isDeleted', 'ownerIds', 'memberIds'];
```
Ciunkos: Consider sorting this.

server/common/utils/helpers.test.js
```diff
@@ -174,7 +174,7 @@ describe('function responseWithCohort', () => {
...
```
Ciunkos: This as well.

server/controllers/cohort.js
```diff
@@ -218,11 +219,21 @@ const deleteCohortById = (req, resp) => {
       if (lists) {
         const listsIds = lists.map(lists => lists._id);
 
-        return Comment.deleteMany({ listId: { $in: listsIds } });
+        return Comment.updateMany(
+          { listId: { $in: listsIds } },
```
Ciunkos: Shouldn't the listIds be sanitized as well?

adamklepacz: Those listsIds are coming from the database, not from the request of the client, so in my opinion, it doesn't have to be sanitized. Those ids are generated by MongoDB itself, so there will be no query injections.

Ciunkos: Ok

PR #270, opened 2019-08-08
[EOC-370] Add history or last edit by user name to the item

adamklepacz:
When a new item is added there is no field "Last edit by:"
If someone edited name or description this field appears so other users know who was the last editor.


client/src/locales/en.json
```diff
@@ -145,60 +145,61 @@
   "dashboard.index.add": "Add new sack",
   "list.archived-items-container.hide": "Hide archived items",
   "list.archived-items-container.show": "Show archived items",
-  "list.input-bar.placeholder": "What is missing?",
+  "list.index.arch-sack": "Archive the '{name}' sack",
+  "list.index.archivization": "'{name}' sack archivization...",
+  "list.index.change-type-question": "Are you sure you want to change the '{name}' sack to {value}?",
+  "list.index.changing-type": "Changing '{name}' sack's type to {value}...",
+  "list.index.hide-members": "Hide sacks members",
```
Ciunkos: Shouldn't it be "Hide sack's members"?

client/src/locales/en.json
```diff
@@ -299,6 +300,7 @@
   "activity.item.clear-vote": "{performer} removed vote from '{item}' item.",
   "activity.item.clear-vote.no-item": "{performer} removed vote from item.",
   "activity.item.edit-name": "{performer} changed {value} 'item's name to '{item}'.",
```
Ciunkos:
```suggestion
  "activity.item.edit-name": "{performer} changed {value} item's name to '{item}'.",
```

client/src/locales/en.json
```diff
@@ -299,6 +300,7 @@
...
+  "activity.item.edit-by": "{performer} edited item: {item}.",
```
Ciunkos:
```suggestion
  "activity.item.edit-by": "{performer} edited item: '{item}'.",
```

client/src/locales/en.json
```diff
@@ -299,6 +300,7 @@
...
   "activity.item.edit-name.no-item": "{performer} changed {value} 'item's name.",
```
Ciunkos: Please fix all occurrences of `'item's`

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -362,7 +362,8 @@ class ListItem extends PureComponent {
         descriptionLock = false,
         isOrdered,
         name,
-        nameLock = false
+        nameLock = false,
+        editedBy
```
Ciunkos: Please sort.

client/src/modules/list/components/Items/model/reducer.js
```diff
@@ -102,7 +103,8 @@ const items = (state = {}, action) => {
         [itemId]: {
           ...previousItem,
           description: newDescription,
-          name: name || previousItem.name
+          name: name || previousItem.name,
+          editedBy
```
Ciunkos: Please sort.

server/controllers/list.js
```diff
@@ -947,81 +947,104 @@ const updateListItem = (req, resp) => {
     user: { _id: userId }
   } = req;
   const { id: listId } = req.params;
+  const sanitizedUserId = sanitize(userId);
```
Ciunkos: No need for sanitization as this is provided by the auth middleware.

server/controllers/list.js
```diff
@@ -947,81 +947,104 @@ const updateListItem = (req, resp) => {
...
+
+          if (isArchived !== undefined) {
+            itemToUpdate.isArchived = isArchived;
+            editedItemActivity = isArchived
+              ? ActivityType.ITEM_ARCHIVE
+              : ActivityType.ITEM_RESTORE;
+          }
+
+          if (editedBy) {
+            itemToUpdate.editedBy = editedBy;
```
Ciunkos: `itemToUpdate.editedBy = userId;` - no need for additional DB query.

server/models/item.model.js
```diff
@@ -7,6 +7,7 @@ const ItemSchema = new Schema(
   {
     authorId: { type: ObjectId, ref: 'User', required: true },
     description: { type: String, default: '' },
+    editedBy: { type: String },
```
Ciunkos: This should be an optional ref to `User`

adamklepacz:
@Ciunkos  @marekrozmus  please do not review. I need to to finish this up on Friday. I'll drop you a line when ready to review.

adamklepacz:
@Ciunkos @marekrozmus  Ready for review

server/common/utils/helpers.js
```diff
@@ -110,13 +110,27 @@ const responseWithListsMetaData = (lists, userId) =>
 
 const responseWithItems = (userId, items) =>
   _map(items, item => {
-    const { authorId: author, isArchived, voterIds, ...rest } = item;
+    const {
+      authorId: author,
+      editedBy: editor,
```
Ciunkos: Please keep the original name here, but respond with:

```javascript
const { displayName: editorDisplayName  } = editedBy || {}
return {
  ...
  editedBy: editorDisplayName,
  ...
};
```

server/controllers/list.js
```diff
@@ -1003,6 +1003,9 @@ const updateListItem = (req, resp) => {
           : ActivityType.ITEM_RESTORE;
       }
 
+      itemToUpdate.editedBy = userId;
+      editedItemActivity = ActivityType.ITEM_EDIT_BY;
```
Ciunkos: Please just name the action `ITEM_EDIT` - it must be done by someone.

server/sockets/list.js
```diff
@@ -107,11 +109,25 @@ const updateItemState = (socket, itemClientLocks) => {
...
+    let editedBy;
+
+    User.findById(sanitizedUserId)
+      .then(user => {
+        if (user) {
+          const { displayName } = user;
+          editedBy = displayName;
+        }
+      })
+      .finally(() => {
```
Ciunkos: Please use `then` - `finally` should be used as a cleanup for error handling. In case of missing user - I would throw an error as this is not an expected situation.

client/src/locales/en.json
```diff
@@ -145,60 +145,61 @@
...
+  "list.list-archived-item.author": "Added by: {authorName}",
+  "list.list-archived-item.confirmation": "Do you really want to delete '{name}' item?",
+  "list.list-archived-item.delete": "Delete",
+  "list.list-archived-item.done": "Done",
+  "list.list-archived-item.restore": "Restore",
+  "list.list-archived-item.unhandled": "Unhandled",
+  "list.list-archived-item.votes-count": "votes: {votesCount}",
+  "list.list-description.add-button": "Add description",
+  "list.list-description.title": "Description",
+  "list.list-header.add-button": "Add description",
```
marekrozmus: do you need two "Add description" strings?

server/sockets/list.js
```diff
@@ -107,11 +109,25 @@ const updateItemState = (socket, itemClientLocks) => {
...
+    let editedBy;
+
+    User.findById(sanitizedUserId)
+      .then(user => {
+        if (user) {
+          const { displayName } = user;
+          editedBy = displayName;
+        }
+      })
+      .then(() => {
```
marekrozmus: why this `then`? shouldn't all that code be in `if` block of previous `then`? What happens if user is not found?

adamklepacz: When the item is updated, there is ALWAYS a User that is performing that action, so the user will always be found.

client/src/modules/list/components/Items/ListItemName/ListItemName.jsx
```diff
@@ -67,18 +69,21 @@ class ListItemName extends PureComponent {
...
-
-          this.handleNameInputBlur();
-          onBlur();
-        }
-      );
+      updateListItem(name, listId, itemId, userData, {
+        name: updatedName
+      }).finally(() => {
+        this.setState({
+          pending: false,
```
Ciunkos: Consider sorting this.

server/controllers/list.js
```diff
@@ -1003,6 +1003,9 @@ const updateListItem = (req, resp) => {
           : ActivityType.ITEM_RESTORE;
       }
 
+      itemToUpdate.editedBy = userId;
+      editedItemActivity = ActivityType.ITEM_EDIT;
```
Ciunkos: `editedItemActivity` will always be `ActivityType.ITEM_EDIT` - is this expected?

server/sockets/list.js
Ciunkos: I think the thens should be combined so it is more clear now - `let` declaration can be moved inside.

server/controllers/list.js
adamklepacz: This activityType should never be registered as EDITING means edition of name or description. And we're already saving those activities, so this one doesn't make any sense.

I removed this one

server/sockets/list.js
```diff
@@ -107,11 +109,21 @@ const updateItemState = (socket, itemClientLocks) => {
...
   socket.on(ItemActionTypes.UPDATE_SUCCESS, data => {
-    const { listId } = data;
+    const { listId, userId } = data;
+    const sanitizedUserId = sanitize(userId);
 
-    socket.broadcast
-      .to(listChannel(listId))
-      .emit(ItemActionTypes.UPDATE_SUCCESS, data);
+    User.findById(sanitizedUserId).then(user => {
+      if (user) {
```
Ciunkos: Please do not hide the error when the user is not found - please let it throw.

adamklepacz: but we are intentionally not catching errors within sockets logic, as sockets are not the main functionality of the app. 

If there is an error we do not handle them in any way, because it will not affect app functionality.

Ciunkos: I think you should at least log it. Anyway, if it is by design, then it is OK for me.

PR #271, opened 2019-08-09
[EOC-290] Add validation of required fields when creating sacks or cohorts

adamklepacz:
Zrzut ekranu 2019-08-9 o 09 14 11
Zrzut ekranu 2019-08-9 o 09 14 02


client/src/common/components/Form.jsx
```diff
@@ -74,6 +75,11 @@ class Form extends PureComponent {
             type="text"
             value={name}
           />
+          {errorMessageId && (
+            <span className="error-message ">
```
marekrozmus: unneeded space before "

client/src/common/components/FormDialog.jsx
```diff
@@ -39,14 +42,30 @@ class FormDialog extends Component {
 
   handleDescriptionChange = description => this.setState({ description });
 
-  handleNameChange = name => this.setState({ name });
+  handleNameChange = name => {
```
marekrozmus: why braces were added?

client/src/common/components/FormDialog.jsx
```diff
@@ -39,14 +42,30 @@ class FormDialog extends Component {
...
   handleConfirm = () => {
     const { defaultDescription, defaultName, onConfirm } = this.props;
-    const { description, name } = this.state;
+    const { description, name, errorMessageId } = this.state;
+
+    this.validateName();
 
-    if (defaultDescription !== description || defaultName !== name) {
-      name && onConfirm(name, description);
+    if (!errorMessageId) {
```
marekrozmus: I'm not sure how this upposed to work:
- you take from state `errorMessageId`
- you call `validateName` which set that exxact part of state
- you check in `if` the value of `errorMessageId` which have previous value (before `validateName`)

And additionally `setState` is async so you cannot be sure if this is already set after `validateName` finishes

client/src/common/components/FormDialog.jsx
```diff
@@ -39,14 +42,30 @@ class FormDialog extends Component {
...
+      if (defaultDescription !== description || defaultName !== name) {
+        return name && onConfirm(name, description);
```
marekrozmus: why this `return`? Do you need result of `onConfirm` anywhere?

client/src/common/components/FormDialog.jsx
adamklepacz: by mistake

client/src/common/components/FormDialog.jsx
adamklepacz: yeah, I was also not 100% sure how this was working as the setState is async.

I've reimplemented the logic, please check it right now

client/src/common/components/Form.jsx
```diff
@@ -1,6 +1,6 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
-import { injectIntl } from 'react-intl';
+import { injectIntl, FormattedMessage } from 'react-intl';
```
Ciunkos: Consider sorting this. In order to minimize nit-picking, consider using https://eslint.org/docs/rules/sort-imports

client/src/common/components/Form.jsx
```diff
@@ -74,6 +75,11 @@ class Form extends PureComponent {
             type="text"
             value={name}
           />
+          {errorMessageId && (
+            <span className="error-message">
+              <FormattedMessage id="common.form.default-warning" />
```
Ciunkos: I think the `errorMessageId` should be used, otherwise, if there is an unknown error - only then it should fall back to the default error message.

client/src/common/components/FormDialog.jsx
```diff
@@ -39,17 +42,31 @@ class FormDialog extends Component {
...
 
   handleConfirm = () => {
     const { defaultDescription, defaultName, onConfirm } = this.props;
-    const { description, name } = this.state;
+    const { description, name, errorMessageId } = this.state;
 
-    if (defaultDescription !== description || defaultName !== name) {
-      name && onConfirm(name, description);
+    if (!errorMessageId) {
+      if (defaultDescription !== description || defaultName !== name) {
```
Ciunkos: Please refactor this into:
```javascript
const descriptionDiffers = defaultDescription !== description;
const nameDiffers = defaultName !== name;
const differs = descriptionDiffers || nameDiffers;

if (name && differs) {
  onConfirm(name, description);
}
```

client/src/locales/en.json
```diff
@@ -24,8 +24,9 @@
   "common.cookie-consent.message": "We use cookies to ensure you get the best experience. By using this website, you consent to use of these cookies.",
   "common.cookie-consent.privacy-1": "For more information, please see our",
   "common.cookie-consent.privacy-2": "privacy policy.",
-  "common.form.name": "Name",
+  "common.form.default-warning": "This field is required.",
```
Ciunkos: I would go for `common.form.required` or `common.form.required-warning`.

client/src/common/components/Form.jsx
adamklepacz: I will add this rule on a separate branch(task), because there will too many diffs to check on this PR

client/src/common/components/Form.jsx
```diff
@@ -1,6 +1,6 @@
+import { FormattedMessage, injectIntl } from 'react-intl';
```
Ciunkos: Consider moving it below common react and prop-types packages.

client/src/common/components/FormDialog.jsx
```diff
@@ -39,17 +42,33 @@ class FormDialog extends Component {
...
   handleDescriptionChange = description => this.setState({ description });
 
-  handleNameChange = name => this.setState({ name });
+  handleNameChange = name => this.setState({ name }, this.validateName);
+
+  validateName = callback => {
+    const { name } = this.state;
+
+    const errorMessageId = validateWith(value =>
+      validator.isLength(value, { min: 1, max: 32 })
```
Ciunkos: Please display "Field is too long" for messages that exceed the max length - currently it would display the "Field is required" message for a text of 35 characters.

client/src/common/components/FormDialog.jsx
```diff
@@ -39,17 +42,40 @@ class FormDialog extends Component {
...
+    const { name } = this.state;
+    let errorMessageId;
+
+    errorMessageId = validateWith(value => !validator.isEmpty(value))(
+      'common.form.required-warning'
+    )(name);
+
+    if (name) {
+      errorMessageId = validateWith(value =>
+        validator.isLength(value, { min: 1, max: 21 })
```
Ciunkos: Typo?

```suggestion
        validator.isLength(value, { min: 1, max: 32 })
```

PR #272, opened 2019-08-09
[EOC-217] Change SVG icons to components

adamklepacz:
I have also removed unused ones.


client/src/assets/images/icons/index.js
```diff
@@ -473,3 +473,37 @@ export const ErrorIcon = () => (
     />
   </svg>
 );
+
+export const AppIcon = () => (
+  <svg
+    aria-hidden="true"
```
Ciunkos: Consider extracting this as:
```javascript
const Icon = props => (
  <svg
    className="svg-inline--fa fa-coffee fa-w-20"
    data-prefix="fas"
    role="img"
    viewBox="0 0 640 512"
    xmlns="http://www.w3.org/2000/svg"
    {..props}
  />
);
```

Then you can use it like this:

```javascript
export const AppIcon = () => (
  <Icon data-icon="coffee">
    <path
      d="M192 384h192c53 0 96-43 96-96h32c70.6 0 128-57.4 128-128S582.6 32 512 32H120c-13.3 0-24 10.7-24 24v232c0 53 43 96 96 96zM512 96c35.3 0 64 28.7 64 64s-28.7 64-64 64h-32V96h32zm47.7 384H48.3c-47.6 0-61-64-36-64h583.3c25 0 11.8 64-35.9 64z"
      fill="#ef9b1f"
    />
  </Icon>
)
```

This reduces duplication a lot and makes all the icons use consistent sizes and parameters. IMHO, `aria-hidden="true"` should be deleted as this is a false statement and may confuse accessibility tools - please see https://css-tricks.com/accessible-svgs/#article-header-id-8

client/src/common/components/AppLogo.scss
```diff
@@ -2,8 +2,11 @@
   display: flex;
   align-items: center;
 
-  &__icon {
+  span.app-logo__icon {
     height: 16px;
```
Ciunkos: Please use `constant-size` helper to avoid SVG sizes to be reduced below the expected size.

client/src/common/components/SortBox.scss
```diff
@@ -45,8 +45,8 @@
     width: 40px;
     cursor: pointer;
 
-    img {
-      width: 14px;
+    svg {
+      height: 14px;
```
Ciunkos: Please use the `constant-size` helper.

adamklepacz:
> Where are all those deleted images? They are not use anymore?

As in the description: 
> I have also removed **unused** ones.

client/src/assets/images/icons/index.js
```diff
@@ -1,425 +1,164 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
-export const CohortIcon = () => (
+const Icon = ({ children, ...props }) => (
```
Ciunkos: No need for children extraction, if you spread props over the component it will render children as well.

client/src/assets/images/icons/index.js
```diff
@@ -1,425 +1,164 @@
...
   <svg
-    aria-hidden="true"
-    className="svg-inline--fa fa-users fa-w-20"
-    data-icon="users"
+    className="svg-inline--fa fa-w-20"
     data-prefix="fas"
     role="img"
```
Ciunkos: Consider using `aria-role` as well.

adamklepacz: aria-role is invalid attribute.
Eslint also warns about this: 
Zrzut ekranu 2019-08-12 o 11 24 51

https://www.w3.org/WAI/PF/aria-1.1/states_and_properties

Ciunkos: Ok

client/src/common/components/SortBox.jsx
```diff
@@ -58,17 +58,15 @@ class SortBox extends Component {
           <button
             className={orderButtonClass}
             onClick={this.handleSortOrderChange}
+            title={formatMessage(
+              {
+                id: 'common.sort-box.sort'
+              },
+              { sortOrder }
```
marekrozmus: `sortOrder` got always english word `ascending` or `descending` taken from constant so it is not translated.

PR #274, opened 2019-08-12
[EOC-413] While creating shared list on cohort view creating second as limited causes creation of shared

adamklepacz:


client/src/modules/cohort/index.jsx
```diff
@@ -172,7 +172,7 @@ class Cohort extends PureComponent {
     this.setState({ pendingForListCreation: true });
 
     createList(data).finally(() => {
-      this.setState({ pendingForListCreation: false });
+      this.setState({ pendingForListCreation: false, type: ListType.LIMITED });
```
Ciunkos: Maybe you should set the state to the initial state all at once? Is that expected behavior here?

PR #276, opened 2019-08-14
[EOC-422] Database model migration - old lists and cohorts are not working on production server  - investigation

adamklepacz:
This is the result of my investigation about migrating mongoDB models.
This branch is not gonna be merged, it just a draft of how it may look like. 

I haven't found any working library for this and any tutorial that makes sense so I decided to write something by myself. 
The flow goes like this:

1)  Running the script:` npm run migrate:list`
2) The script is connecting to the Database
3) The script is looking inside List model what properties model has,
4) The script is checking in all List collection if the document doesn't have given property for example
`{isDeleted, {$extists: false}}`
5) The script sets property if the document doesn't have one
6) After successful migration, the script disconnects from the database

`connectDatabase` and `disconnectDatabase` function will be moved to `common/helpers` as now they are gonna be used across many files.

This is an example of List model. While implementing, the same logic will be needed for all models.
**Please point what can be improved and what can be done better.**
**In Jira, task EOC_422 you can find screen recording of how it works.**


server/tasks/migrateListModel.js
```diff
@@ -0,0 +1,51 @@
...
+        const queryField = queryNonExistingSchemaField(field);
+        const setField = setNonExistingSchemaField(field, ListSchema);
+
+        return List.updateMany(queryField, setField)
+          .lean()
+          .exec()
+          .then(() => console.log('Successfully migrated list model...'))
+          .catch(err => {
+            throw err;
+          })
```
marekrozmus: unneeded catch clause if the error is rethrown

adamklepacz:
@marekrozmus 
Yes, this will only work for List. 
This is just an investigation. For other models, we have to do the same thing. 
This is only an example of `up migration. `
We can divide the implementation of this task for several tasks - for every single model (migration up and down).

Ciunkos:
After a while of thinking there are few ideas that came to my mind:
- No down migrations - if you want to revert a migration, you create a new one that reverts the previous ones - just like revert commits.
- Semver based migrations - you create a migration like `migrations/1.0.0.js` - you apply them in the version ascending order, for example: `1.0.0, 1.0.1, 1.0.2, 1.1.0, 1.2.0, 2.0.0, 2.0.1` etc.
- You keep the current schema version within the database to be able to read the current version. After within the migration you update the version field.
- You seed the database with the newest version.
- You run pending migrations when starting up the db service - you look at the current version and apply the migrations in the semver order.

For example:
Current db version is 1.0.0.

Migration (1.0.1.js);
```
export const version = '1.0.1';

export default database => {
  database.lists.addField('test', null);
};
```

Runner:
```javascript
const apply = (migration, version)=> {
  migration(database);
  database.version = version;
}
```

Database up:
```javascript
const up = () => {
  const database = connect();
  const version = database.version();
  const migrations = pendingMigrations(version);
  migrations.forEach(migration => apply(migration));
  return database;
}
```

alemikolo:
What about 'node-migration': https://www.npmjs.com/package/migrate, additional tutorial: https://101node.io/blog/how-to-automate-database-migrations-mongodb-using-nodejs/ or migrate-mongo': https://www.npmjs.com/package/migrate-mongo ?

adamklepacz:
Task marked as done in Jira. I'm closing this PR

PR #283, opened 2019-08-20
[EOC-377] Add routes for reseting password, confirmation account for logged user

adamklepacz:
Lines from 145 to 159 look identical to lines from 102 to 116, so I've tried to not repeat them by creating a method,
```
renderRoutes() => (
  <Fragment>
     Repeated routes here...
 </Fragment>
)
```
 and call it in those two places. Unfortunately, **<Framgment>** can't be used within `<Switch>` component and app crashed. 
So finally I left it as it is.

### DIFF
--- a/client/src/app/components/Layout/Layout.jsx
+++ b/client/src/app/components/Layout/Layout.jsx
@@ -142,6 +142,21 @@ export class Layout extends PureComponent {
         </Toolbar>
         <Switch>
           <Redirect from="/" exact to="/dashboard" />
+          <Route component={SuccessMessage} path="/account-created" />
+          <Route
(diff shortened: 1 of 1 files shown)


client/src/app/components/Layout/Layout.jsx
```diff
@@ -142,6 +142,21 @@ export class Layout extends PureComponent {
         </Toolbar>
         <Switch>
           <Redirect from="/" exact to="/dashboard" />
+          <Route component={SuccessMessage} path="/account-created" />
```
Ciunkos: How about just redirecting the user to the dashboard with a pop-up message?

PR #284, opened 2019-08-21
[EOC-433] Add backend validation for name for creating list

adamklepacz:
This task is related to EOC-420 which is already merged.
This task returned from QA.


server/controllers/list.js
```diff
@@ -35,6 +35,10 @@ const createList = (req, resp) => {
   } = req;
   const isSharedList = type === ListType.SHARED;
 
+  if (name !== undefined && !validator.isLength(name, { min: 1, max: 32 })) {
```
marekrozmus: what if name === undefined?

PR #285, opened 2019-08-21
[EOC-421] Tile view on dashboard goes back to list view on page refresh

alemikolo:
This branch does not include migrations. They are implemented in separate branch EOC-424.


client/src/app/components/Layout/Layout.jsx
```diff
@@ -70,24 +84,35 @@ export class Layout extends PureComponent {
...
+
+    const settings = {
+      viewType: newViewType
+    };
+
+    this.handleSwitchView(newViewType);
+    this.setState({ pendingForViewType: true });
+
+    updateSettings(settings)
+      .catch(() => {
```
marekrozmus: remove braces

client/src/modules/user/model/actions.js
```diff
@@ -138,3 +148,14 @@ export const changePassword = (password, newPassword, newPasswordConfirm) =>
...
+
+export const updateSettings = settings => dispatch =>
+  postData('/auth/update-settings', { settings })
+    .then(() => dispatch(updateSettingsSuccess(settings)))
+    .catch(() => {
+      dispatch(updateSettingsFailure());
+      createNotificationWithTimeout(dispatch, NotificationType.ERROR, {
+        notificationId: 'user.actions.update-settings-failed'
+      });
+      throw new Error();
```
marekrozmus: throw error from catch

client/src/app/components/Layout/Layout.jsx
```diff
@@ -70,24 +84,33 @@ export class Layout extends PureComponent {
...
+
+    const settings = {
+      viewType: newViewType
+    };
+
+    this.handleSwitchView(newViewType);
+    this.setState({ pendingForViewType: true });
+
+    updateSettings(settings)
+      .catch(() => this.handleSwitchView(viewType))
```
adamklepacz: why do you handling something on catch??

server/common/utils/userUtils.js
```diff
@@ -5,6 +5,7 @@ const User = require('../../models/user.model');
 const List = require('../../models/list.model');
 const Cohort = require('../../models/cohort.model');
 const Comment = require('../../models/comment.model');
+const Settings = require('../../models/settings.model');
 
 // Find or create user
```
adamklepacz: remove comment

server/common/utils/userUtils.js
```diff
@@ -108,10 +113,23 @@ const removeDemoUserData = id =>
 
 const validatePassword = value => value.match(/^[^\s]{4,32}$/);
 
+const responseWithUserData = user => {
+  const { avatarUrl, _id: id, displayName: name, settings: config } = user;
+  const data = { avatarUrl, id, name };
+
+  if (config) {
+    const { _id, createdAt, updatedAt, ...settings } = config;
```
adamklepacz: line of a break

server/controllers/userAuth.js
```diff
@@ -4,15 +4,19 @@ const _some = require('lodash/some');
...
+  responseWithUserData,
+  validatePassword
+} = require('../common/utils/userUtils');
+const Settings = require('../models/settings.model');
 
 const sendUser = (req, resp) => {
-  const { avatarUrl, _id: id, displayName: name } = req.user;
-
-  resp.send({ avatarUrl, id, name });
+  return resp.send(responseWithUserData(req.user));
```
adamklepacz: unnecessary braces and return

client/src/modules/user/model/actions.js
```diff
@@ -138,3 +148,14 @@ export const changePassword = (password, newPassword, newPasswordConfirm) =>
     newPassword,
     newPasswordConfirm
   });
+
+export const updateSettings = settings => dispatch =>
+  postData('/auth/update-settings', { settings })
```
Ciunkos: Consider just `POST /auth/settings`. Please discuss with the team change of the `auth` route to `user` or `profile` as this is no longer `auth` only endpoint.

client/src/modules/user/model/reducer.js
```diff
@@ -8,6 +8,12 @@ const currentUser = (state = null, action) => {
       return null;
     case AuthorizationActionTypes.FETCH_SUCCESS:
       return { ...(state || {}), ...action.payload };
+    case AuthorizationActionTypes.UPDATE_SETTINGS_SUCCESS: {
+      const { payload: updatedSettings } = action;
+      const { settings } = state;
```
Ciunkos: Please use fallback values to guard against `null` values of `state`/`state.settings`.

server/common/variables.js
```diff
@@ -130,6 +130,11 @@ const ListHeaderStatusTypes = Object.freeze({
 const LOCK_TIMEOUT = 300000;
 const SOCKET_TIMEOUT = 60000;
 
+const ViewType = {
```
Ciunkos: Please use the `enumerable('viewType')('LIST', 'TILES')`.

client/src/app/components/Layout/Layout.jsx
alemikolo: We have an optimistic update here, we update state and change view instantly. And if something goes wrong we rollback changes.

server/controllers/userAuth.js
```diff
@@ -415,6 +418,33 @@ const changePassword = (req, res) => {
...
+  User.findById(userId)
+    .exec()
+    .then(doc => {
+      if (!doc) {
+        throw new NotFoundException();
+      }
+
+      const { settings } = doc;
+
+      _forEach(editedSettings, (value, key) => {
```
Ciunkos: Please extract this as sanitizieObject so it can be reused.

client/src/app/components/Layout/Layout.jsx
adamklepacz: so you can run this also in finally without catch

alemikolo: no,  because in `catch` we rollback the changes if something went wrong, `finally` is invoked regardless promise is resolved or reject.

adamklepacz: now I get IT :)

client/src/modules/user/model/reducer.js
```diff
@@ -8,6 +8,19 @@ const currentUser = (state = null, action) => {
       return null;
     case AuthorizationActionTypes.FETCH_SUCCESS:
       return { ...(state || {}), ...action.payload };
+    case AuthorizationActionTypes.UPDATE_SETTINGS_SUCCESS: {
+      const { payload: updatedSettings } = action;
+      if (state) {
+        const { settings } = state;
```
Ciunkos: Please assign default here - `const { settings = {} } = state || {}`.

server/common/utils/helpers.js
```diff
@@ -326,9 +328,32 @@ const responseWithCohortDetails = (doc, userId) => {
   };
 };
 
+const enumerable = namespace => (...keys) =>
+  Object.freeze(
+    fromEntries(keys.map(key => [key, [namespace, key].join('/')]))
+  );
+
+const forEach = f => o =>
```
Ciunkos: Consider renaming this to `mapObject`.

server/common/utils/helpers.js
```diff
@@ -331,6 +333,23 @@ const enumerable = namespace => (...keys) =>
     fromEntries(keys.map(key => [key, [namespace, key].join('/')]))
   );
 
+const forEach = f => o =>
```
adamklepacz: what are f and o?
Please use words as arguments, not letters as it's not possible to understand what it is

client/src/modules/user/model/actions.js
alemikolo: We will discuss this.

PR #287, opened 2019-08-22
[EOC-401] Improvement refactor clear meta data method

adamklepacz:


client/src/modules/cohort/components/Cohorts.jsx
```diff
@@ -22,7 +21,7 @@ import CollectionView from 'common/components/CollectionView';
 import FormDialog from 'common/components/FormDialog';
 import Breadcrumbs from 'common/components/Breadcrumbs';
 import { ColorType, Routes, ViewType } from 'common/constants/enums';
-import { enterView, leaveView } from 'sockets';
+import { enterView, leaveView } from 'common/model/actions';
```
alemikolo: hmm, we have a general model directory in the app folder. Only wonder if we could we put this new action file there instead of creating a new model in the common directory?

client/src/common/model/actions.js
```diff
@@ -0,0 +1,56 @@
...
+});
+
+const clearCohortMetaDataSuccess = () => ({
+  type: CohortActionTypes.CLEAR_META_DATA_SUCCESS
+});
+
+export const enterView = (route, userId) => dispatch => {
+  const view = _upperFirst(route);
+
+  socket.emit(`enter${view}View`, userId);
```
Ciunkos: Please try not to use computed ids. Rather, send the view as a parameter of generalized action:
```socket.emit('enterView', { userId, view });```

File: client/src/common/model/actions.js [hunk 3]
```diff
@@ -0,0 +1,56 @@
...
+};
+
+export const leaveView = (route, userId) => dispatch => {
+  const view = _upperFirst(route);
```
Ciunkos: Please react to `leaveView` action within the reducers. For example:

List reducer:
```javascript
case LEAVE_VIEW: {
  if (action.payload.view === 'list') {
      return {};
  } else {
      return state;
  }
}
```

For other views, do it in a similar way.

client/src/common/model/actions.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  socket.emit(`leave${view}View`, userId);
+};
+
+export const joinRoom = (route, id, userId) => dispatch => {
+  const room = _upperFirst(route);
+  const data = { room: `${route}-${id}`, userId, viewId: id };
+
+  socket.emit(`join${room}Room`, data);
```
Ciunkos: Please apply a similar pattern as for leaveView action.

client/src/common/model/actions.js
```diff
@@ -0,0 +1,38 @@
...
+import socket from 'sockets';
+import { CommonActionTypes } from 'common/model/actionTypes';
+
+const clearMetaDataSuccess = view => ({
+  type: CommonActionTypes.LEAVE_VIEW,
+  payload: view
+});
+
+export const enterView = (route, userId) => dispatch => {
+  const view = _upperFirst(route);
```
Ciunkos: Please do not compute names - please use raw view names.

client/src/modules/cohort/model/reducer.js
```diff
@@ -130,8 +133,16 @@ const cohorts = (state = {}, action) => {
 
       return { ...state, [cohortId]: cohort };
     }
-    case CohortActionTypes.CLEAR_META_DATA_SUCCESS:
-      return {};
+    case CommonActionTypes.LEAVE_VIEW: {
+      const { payload } = action;
+
+      if (payload === _upperFirst(Routes.COHORTS)) {
```
Ciunkos: Please use raw names.

client/src/common/model/actionTypes.js
```diff
@@ -0,0 +1,4 @@
+export const CommonActionTypes = Object.freeze({
```
Ciunkos: Consider using `enumerable` helper with an empty namespace.

client/src/common/model/actions.js
```diff
@@ -0,0 +1,29 @@
+import socket from 'sockets';
+import { CommonActionTypes } from 'common/model/actionTypes';
+
+const clearMetaDataSuccess = view => ({
+  type: CommonActionTypes.LEAVE_VIEW,
+  payload: view
```
Ciunkos: Consider making the payload always an object to make extension easy when for example, `userId` is needed.

client/src/common/model/actions.js
```diff
@@ -0,0 +1,29 @@
+import socket from 'sockets';
+import { CommonActionTypes } from 'common/model/actionTypes';
+
+const clearMetaDataSuccess = view => ({
```
Ciunkos: Please name this action `leaveViewAction` or something similar.

PR #289, opened 2019-08-26
[EOC-402] Refactor events emiting from actions to controllers for list

adamklepacz:
There is some commented code on this branch, that disabled sockets logic for 
marking item as done /unhandled('updateItem' on backend, toggleItem action on front-end).
This action and controller needs to be refactored from tooglleItem to 
`markAsDone(), markAsUnhandled(),` because toggling causes problems right now, and socket functionality doesn't work properly. 

This will be fixed and refactored on **EOC-404 [IMPROVEMENT] Refactor events emitting from actions to controllers for ITEMS**


server/sockets/list.js
```diff
@@ -782,113 +776,93 @@ const archiveList = (socket, dashboardClients, cohortClients, listClients) =>
...
-      });
-  });
+  io.sockets
+    .to(`sack-${listId}`)
+    .emit(ListActionTypes.DELETE_AND_REDIRECT, data);
 
-const deleteList = (socket, dashboardClients, cohortClients) =>
-  socket.on(ListActionTypes.DELETE_SUCCESS, data => {
-    const { listId, cohortId } = data;
+  dashboardClients.forEach(client => {
```
marekrozmus: could be one line:
`...forEach(({ socketId }) => ...`

server/sockets/index.js
```diff
@@ -70,16 +71,18 @@ const socketListenTo = server => {
...
+  }
 
-  const cohortViewClients = new Map();
-  const allCohortsViewClients = new Map();
-  const dashboardViewClients = new Map();
-  const listViewClients = new Map();
-  const cohortClientLocks = new Map();
-  const itemClientLocks = new Map();
-  const listClientLocks = new Map();
+  return socketInstance;
```
alemikolo: could we return socket instead of socketInstance to not use io.socket everywhere but only socket?

adamklepacz: I've tried to export just `return socketInstance.sockets ` but it doesn't work.

server/controllers/list.js
```diff
@@ -449,21 +477,47 @@ const updateListById = (req, resp) => {
...
+          archiveList(
+            socketInstance,
+            dashboardClients,
+            cohortClients,
+            listClients
+          )(data);
+        }
+
+        listActivity = ActivityType.LIST_RESTORE;
+        restoreList(
```
Ciunkos: Please `return` from the actions so no next actions are fired after them.

server/controllers/list.js
```diff
@@ -719,9 +784,10 @@ const removeOwnerRole = (req, resp) => {
...
 
-      resp.send({
-        message: 'User has no owner role.'
-      });
+      const data = { listId, userId };
+
+      removeOwnerRoleInList(socketInstance, listClients)(data);
+      resp.send();
 
       saveActivity(
```
Ciunkos: Please move it to the `fireAndForget` handler.

server/controllers/list.js
```diff
@@ -991,6 +1072,13 @@ const updateListItem = (req, res) => {
         editedItemActivity = isOrdered
           ? ActivityType.ITEM_DONE
           : ActivityType.ITEM_UNHANDLED;
+
+        // THIS METHOD NEED TO RECOGNIZE CLIENT SOMEHOW,
```
Ciunkos: Please add `// TODO:` to make searching for them easier.

server/controllers/list.js
Ciunkos: Please `then` all actions and return responses.

server/controllers/list.js
```diff
@@ -1157,7 +1245,15 @@ const changeType = (req, resp) => {
...
+
+      changeListType(
+        socketInstance,
+        dashboardClients,
+        cohortClients,
+        listClients
+      )(data);
+      resp.send({ members, type });
 
       saveActivity(
```
Ciunkos: `const fireAndForget = promise => promise.catch((error) => console.error(error));` `fireAndForget( saveActivity(...))`

server/controllers/list.js
```diff
@@ -1157,7 +1245,15 @@ const changeType = (req, resp) => {
...
       );
 
-      resp.send({ members, type, removedViewers });
+      const data = { listId, type, removedViewers };
+
+      changeListType(
+        socketInstance,
+        dashboardClients,
+        cohortClients,
+        listClients
```
Ciunkos: `return changeListType(...).then(() => ({ members, type })`. `.then(payload => resp.send(payload))`

server/controllers/list.js
```diff
@@ -719,18 +824,22 @@ const removeOwnerRole = (req, resp) => {
...
+        saveActivity(
+          ActivityType.LIST_SET_AS_MEMBER,
+          ownerId,
+          null,
+          sanitizedListId,
+          list.cohortId,
+          userId
+        )
       );
+      resp.send();
```
Ciunkos: Please add a new line for consistency.

server/index.js
```diff
@@ -7,8 +7,18 @@ const passport = require('passport');
 const cookieParser = require('cookie-parser');
 const session = require('express-session');
 const path = require('path');
+/* eslint-disable import/order */
```
Ciunkos: Why this rule is disabled? Wouldn't it be better to just sort the imports?

adamklepacz: Added as EOC-433 task in JIRA

PR #290, opened 2019-08-26
[EOC-408] Improvement refactor xValidator methods within components

adamklepacz:


client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
```diff
@@ -86,7 +90,9 @@ class PasswordChangeForm extends PureComponent {
     );
 
   passwordValidator = value =>
-    validatePassword(value) ? '' : 'user.auth.input.password.invalid';
+    validateWith(value => validatePassword(value))(
```
Ciunkos: `value => validatePassword(value)` can be replaced with `validatePassword`

client/src/modules/user/AuthBox/components/PasswordRecoveryForm.jsx
```diff
@@ -89,52 +89,35 @@ class PasswordRecoveryForm extends PureComponent {
 
   passwordValidator = password => {
     const { errors } = this.state;
+    const errorMessageId = validateWith(value => validatePassword(value))(
```
Ciunkos: Here as well.

PR #291, opened 2019-08-26
[EOC-406] Terminate other user sessions after password changes

alemikolo:
To show notification regarding no authorization or end of user's session I had to restore error capturing in action's catch statement and pass this error to the notification, and there, by checking its type display a notification with timer feature or not.

I had a problem with a time value to the notification message. Regardless of the way I was passing it, there were no spaces between the time value and the rest of the notification's message. I have no idea why this was happening. I found some workaround and it's working but I don't know if this should be written in that way: File NotificationWithTimer.jsx, line 41-49.

session-terminate

### DIFF
--- a/client/src/common/components/Activities/model/actions.js
+++ b/client/src/common/components/Activities/model/actions.js
@@ -36,7 +36,7 @@ export const fetchActivities = page => dispatch =>
       createNotificationWithTimeout(
         dispatch,
         NotificationType.ERROR,
-        { notificationId: 'Failed to fetch activities. Please try again.' },
+        { notificationId: 'activity.actions.fetch-fails' },
         err
       );
     });
--- a/client/src/common/components/Activities/model/actions.js
+++ b/client/src/common/components/Activities/model/actions.js
@@ -31,11 +31,12 @@ export const fetchActivities = page => dispatch =>
         })
       );
     })
-    .catch(() => {
+    .catch(err => {
       dispatch(fetchActivitiesFailure());
       createNotificationWithTimeout(
         dispatch,
         NotificationType.ERROR,
-        'Failed to fetch activities. Please try again.'
+        { notificationId: 'Failed to fetch activities. Please try again.' },
+        err
       );
     });
--- a/client/src/common/constants/variables.js
+++ b/client/src/common/constants/variables.js
@@ -1,7 +1,7 @@
+export const asyncPostfixes = ['FAILURE', 'SUCCESS'];
 export const COMPANY_NAME = 'Sandstream Development Sp. z o.o.';
 export const COMPANY_PAGE_URL = 'https://sandstream.pl/';
 export const NOTIFICATION_TIMEOUT = 5000;
-export const REDIRECT_TIMEOUT = 10000;
 export const PENDING_DELAY = 1000;
 export const PROJECT_NAME = 'EOC';
-export const asyncPostfixes = ['FAILURE', 'SUCCESS'];
+export const REDIRECT_TIMEOUT = 10000;
--- a/client/src/common/constants/variables.js
+++ b/client/src/common/constants/variables.js
@@ -1,6 +1,7 @@
 export const COMPANY_NAME = 'Sandstream Development Sp. z o.o.';
 export const COMPANY_PAGE_URL = 'https://sandstream.pl/';
 export const NOTIFICATION_TIMEOUT = 5000;
+export const REDIRECT_TIMEOUT = 10000;
 export const PENDING_DELAY = 1000;
 export const PROJECT_NAME = 'EOC';
 export const asyncPostfixes = ['FAILURE', 'SUCCESS'];
--- a/client/src/locales/en.json
+++ b/client/src/locales/en.json
@@ -106,6 +106,7 @@
   "user.auth.reset-password.email-label": "Enter email",
   "user.auth.reset-password.button-content": "Reset password",
   "user.auth.pass-recovery-form.heading": "Set new password",
+  "user.auth.session-ended": "Your session has ended, you will be redirected to the login page in 5s.",
   "user.actions.fetch-user-details-error": "Failed to fetch your profile data.",
   "user.actions.logout-fail": "Logout failed.",
   "user.actions.login": "Login success.",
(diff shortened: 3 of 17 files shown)


client/src/common/components/Activities/model/actions.js
```diff
@@ -31,11 +31,12 @@ export const fetchActivities = page => dispatch =>
...
       );
     })
-    .catch(() => {
+    .catch(err => {
       dispatch(fetchActivitiesFailure());
       createNotificationWithTimeout(
         dispatch,
         NotificationType.ERROR,
-        'Failed to fetch activities. Please try again.'
+        { notificationId: 'Failed to fetch activities. Please try again.' },
```
adamklepacz: shouldn't this be an id, not an actual message?

client/src/common/constants/variables.js
```diff
@@ -1,6 +1,7 @@
 export const COMPANY_NAME = 'Sandstream Development Sp. z o.o.';
 export const COMPANY_PAGE_URL = 'https://sandstream.pl/';
 export const NOTIFICATION_TIMEOUT = 5000;
+export const REDIRECT_TIMEOUT = 10000;
```
adamklepacz: sort alphabetically

client/src/locales/en.json
```diff
@@ -106,6 +106,7 @@
   "user.auth.reset-password.email-label": "Enter email",
   "user.auth.reset-password.button-content": "Reset password",
   "user.auth.pass-recovery-form.heading": "Set new password",
+  "user.auth.session-ended": "Your session has ended, you will be redirected to the login page in {data} s.",
```
adamklepacz: will it countdown? :D

client/src/modules/notification/components/NotificationWithTimer.jsx
```diff
@@ -0,0 +1,60 @@
...
+      time
+    };
+  }
+
+  componentDidMount() {
+    this.createInterval();
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.timer);
```
adamklepacz: I think this.timer should be declared in the constructor with some default value

server/controllers/userAuth.js
```diff
@@ -411,6 +411,25 @@ const changePassword = (req, res) => {
...
+      const regexp = new RegExp(userId);
+
+      return db
+        .collection('sessions')
+        .find({
+          _id: { $ne: sessionID },
+          session: regexp
+        })
+        .forEach(session => {
+          const { _id: id } = session;
```
adamklepacz: line of a break

adamklepacz: do you need to map it from _id to id?

server/controllers/userAuth.js
```diff
@@ -411,6 +411,26 @@ const changePassword = (req, res) => {
...
+      } = req;
+      const regexp = new RegExp(userId);
+
+      return db
+        .collection('sessions')
+        .find({
+          _id: { $ne: sessionID },
+          session: regexp
+        })
+        .forEach(session => {
```
marekrozmus: `({ _id }) => store.destroy(_id)`

client/src/modules/notification/index.jsx
```diff
@@ -2,36 +2,37 @@ import React, { Fragment } from 'react';
...
               return (
                 <li className="notification__list-item" key={id}>
-                  <MessageBox type={item.type}>
-                    <FormattedMessage
+                  {redirect ? (
+                    <NotificationWithTimer
                       id={notificationId}
-                      values={{ data: <strong>{itemName}</strong> }}
+                      redirect={redirect}
+                      time={REDIRECT_TIMEOUT / 1000}
```
marekrozmus: `timeout`

client/src/modules/notification/model/actions.js
```diff
@@ -16,12 +20,21 @@ const removeNotification = payload => ({
 export const createNotificationWithTimeout = (
   dispatch,
   type,
-  notification,
+  message,
+  err = null,
   timeout = NOTIFICATION_TIMEOUT
 ) => {
+  const forbiddenError = err instanceof ForbiddenException;
+
```
adamklepacz: remove thath empty line

client/src/modules/notification/model/actions.js
```diff
@@ -16,12 +20,21 @@ const removeNotification = payload => ({
...
   const id = uniqueId('notification_');
-  dispatch(addNotification({ id, type, notification }));
+  const delay = forbiddenError ? REDIRECT_TIMEOUT : timeout;
+  const notification = forbiddenError
+    ? { notificationId: 'user.auth.session-ended' }
```
adamklepacz: what if we gonna have other notifications with counter??
wouldn't be better to pass notifictaionID as parameter, as you do in regular notification via message parameter ??

server/controllers/userAuth.js
```diff
@@ -411,6 +411,22 @@ const changePassword = (req, res) => {
...
+        sessionStore: store,
+        sessionStore: { db }
+      } = req;
+      const regexp = new RegExp(userId);
+
+      return db
+        .collection('sessions')
+        .find({
+          _id: { $ne: sessionID },
+          session: regexp
```
adamklepacz: will this query find userId in his session ?? wouldn't it destroy all the existing sessions in the database, even those ones that not belong to him??

cause session prop in DB looks like this 
`{"cookie":{"originalMaxAge":null,"expires":null,"httpOnly":true,"path":"/"},"passport":{"user":"5c9dc5a619ae7924074940f4"}}`

client/src/modules/notification/model/actions.js
alemikolo: If we gonna have other notifications with I think we pass notificationId or some other prop from specific actions to the notification dispatcher... In `createNotificationWithTimeout` we will be able to decide what kind of notification should we display.
But in this case, in all actions in the app, I would have to check the type of the error and pass right notifactionId or dispatch different action. I didn't want to do all these things in all app's actions. So I decided to pass only error from catch to notification `createNotificationWithTimeout` and all the logic write in one place.

server/controllers/userAuth.js
alemikolo: We create a regexp with userId in line 420: `const regexp = new RegExp(userId)`. Mongo recognizes regexp and uses it to find this user id in the session's string. Other user's session remain.

PR #293, opened 2019-08-27
[EOC-439] Profile view. Password changed successfully message is displayed after putting wrong current password filed. 

alemikolo:


server/controllers/userAuth.js
```diff
@@ -411,6 +416,8 @@ const changePassword = (req, res) => {
 
         return doc.save();
       }
+
+      throw new Error();
```
Ciunkos: Please send validation exception in this case - consider inverting if so the check if for `!validHash` and throws an error from inside.

client/src/modules/user/AuthBox/components/PasswordChangeForm.jsx
```diff
@@ -158,24 +161,27 @@ class PasswordChangeForm extends PureComponent {
...
+                  : '',
+                newPasswordError: isNewPasswordError
+                  ? 'user.auth.input.password.invalid'
+                  : ''
+              };
+            } else {
+              newState.changePasswordErrorId =
+                'common.auth.input.wrong-password';
+            }
+          } else
```
Ciunkos: Please add braces.

server/controllers/userAuth.js
```diff
@@ -403,17 +409,26 @@ const changePassword = (req, res) => {
...
+
+      return doc.save();
     })
     .then(() => res.send())
-    .catch(() => res.sendStatus(400));
+    .catch(err => {
+      if (err instanceof ValidationException) {
+        const { status } = err;
+
+        res.status(status).send({ reason: BadRequestReason.VALIDATION });
```
Ciunkos: Please `return` so the next call is not made.

client/src/locales/en.json
```diff
@@ -100,6 +100,7 @@
   "user.auth.input.password.confirm": "Confirm password",
   "user.auth.input.new-password.confirm": "Confirm new password",
   "user.auth.input.password.invalid": "Please use min 4 and max 32 characters without spaces.",
+  "common.auth.input.wrong-password": "Please enter correct current password.",
```
adamklepacz: just `Please enter the correct password. `

current password label is displayed above the input

PR #296, opened 2019-08-28
[EOC-434] Bug restore item via socket not working properly

adamklepacz:


client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -400,19 +404,13 @@ class ListItem extends PureComponent {
...
-    } = this.state;
+    const { areDetailsVisible, disableToggleButton, isNameEdited } = this.state;
     const isEdited = nameLock || descriptionLock;
 
     return (
       <li
         className={classNames('list-item', {
-          'list-item--restore': !done && isOrdered,
-          'list-item--done': done || isOrdered,
+          'list-item--done': isOrdered,
```
Ciunkos: Shouldn't it be named `done`?

client/src/modules/list/components/Items/model/reducer.js
```diff
@@ -89,21 +73,21 @@ const items = (state = {}, action) => {
...
+        previousDescription = previousItem.description;
+      }
 
       return {
         ...state,
         [itemId]: {
-          ...previousItem,
-          description: newDescription,
+          ...(previousItem || action.payload),
+          description: description || previousDescription,
```
Ciunkos: Why the previous description is kept?

server/controllers/list.js
```diff
@@ -39,13 +39,15 @@ const {
   addMemberRoleInList,
   addOwnerRoleInList,
   archiveList,
+  // changeItemOrderState,
```
Ciunkos: Commented out code left.

server/controllers/list.js
```diff
@@ -1124,43 +1120,59 @@ const updateListItem = (req, res) => {
...
+        )(doc);
       }
 
       if (isArchived !== undefined) {
         itemToUpdate.isArchived = isArchived;
-        editedItemActivity = isArchived
-          ? ActivityType.ITEM_ARCHIVE
-          : ActivityType.ITEM_RESTORE;
-      }
+        itemToUpdate.editedBy = userId;
```
Ciunkos: Why not just assign it unconditionally?

client/src/modules/list/components/Items/ListItem/ListItem.jsx
adamklepacz: On backend `isOrdered`, here was `done` - it was a bit confusing, so I decided to make it consistent and just use `isOrdered` everywhere

client/src/modules/list/components/Items/model/reducer.js
adamklepacz: Because when you update the just the name and description will come as an empty string or undefined from backend it will actually overwrite the current description.

That's why we have to check if the previous value exists.

server/controllers/list.js
adamklepacz: Because inside every if I have `return` statement at the end,  so if I put this piece of code outside of all if`s it will never be executed.

client/src/modules/list/components/Items/ListItem/ListItem.jsx
Ciunkos: Ok, however, `done` is better IMHO as the `isOrdered` could be confused as if the item is sorted or not. Anyway, that would require a migration.

client/src/modules/list/components/Items/model/reducer.js
Ciunkos: That behavior is a little bit confusing - what if I wanted to delete description from 'Test' to an empty text ''?

server/controllers/list.js
Ciunkos: You could assign it before the first condition is ever run.

client/src/modules/list/components/Items/ListItem/ListItem.jsx
alemikolo: hmm, maybe we should think about some refactoring? And change isOrdered to isDone which is more universal... (in the separate task of course)

client/src/modules/list/components/Items/model/reducer.js
alemikolo: Could we create some helper/helpers to filter ous undefined props or 'empty' strings and the use the filtred object to avoid those multiple checks?

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -34,52 +34,53 @@ class ListItem extends PureComponent {
...
-    const shouldChangeAuthor = isNotSameAuthor && isOrdered;
-
-    if (shouldChangeAuthor) {
-      return toggle(itemName, isOrdered, _id, listId, userId, name).finally(
-        this.handleItemUnlock
-      );
-    }
-
-    toggle(itemName, isOrdered, _id, listId).finally(this.handleItemUnlock);
+    return updateListItem(itemName, listId, _id, userData, data).finally(() => {
```
marekrozmus: remove braces

server/controllers/list.js
```diff
@@ -1124,43 +1119,59 @@ const updateListItem = (req, res) => {
...
 
-      itemToUpdate.editedBy = userId;
+        if (isArchived) {
+          editedItemActivity = ActivityType.ITEM_ARCHIVE;
+        }
 
-      return doc.save();
+        if (!isArchived) {
+          editedItemActivity = ActivityType.ITEM_RESTORE;
+        }
```
marekrozmus: why two `ifs`? Use `if...else` or one line:
`editedItemActivity = isArchived ? ActivityType.ITEM_ARCHIVE : ActivityType.ITEM_RESTORE`

client/src/modules/list/components/Items/ListItem/ListItem.jsx
adamklepacz: added as new taski in JIRA

client/src/modules/list/components/Items/model/reducer.js
adamklepacz: we have already helper for that ->` isDefined()`

adamklepacz: > That behavior is a little bit confusing - what if I wanted to delete description from 'Test' to an empty text ''?

You right, now it works correctly :)

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -78,9 +78,9 @@ class ListItem extends PureComponent {
 
     this.handleItemLock();
 
-    return updateListItem(itemName, listId, _id, userData, data).finally(() => {
-      this.setState({ disableToggleButton: false }, this.handleItemUnlock);
-    });
+    return updateListItem(itemName, listId, _id, userData, data).finally(() =>
+      this.setState({ disableToggleButton: false }, this.handleItemUnlock)
```
alemikolo: setState in finally may cause error/warning: "can't call setState on unmounting components (...)", what if updated item is unmounting? Or we don't care about this warning any more?

client/src/modules/list/components/Items/model/reducer.js
```diff
@@ -75,18 +75,18 @@ const items = (state = {}, action) => {
...
+      const newDescription = isDefined(description)
+        ? description
+        : previousDescription;
 
       return {
         ...state,
         [itemId]: {
           ...(previousItem || action.payload),
-          description: description || previousDescription,
+          description: newDescription,
```
alemikolo: hmm, I think about something like this: 

this are our function in helpers:
```
const defined = x => x !== undefined;

const filter = f => object =>
  fromEntries(
    Object.entries(object).filter(([key, value]) => f(value, key, object))
  );

```
I think about those additional helpers:
```
const updated = x => definded(x) && x !== '';
export const filterUpdated = filter(updated);


```
and in reducer:

```
const { _id: itemId, ...rest } = action.payload;
const dataToUpdate = filterUpdated(rest);

return {...state, [itemId]: { ...state[itemId, ...dataToUpdate }
```
If we'll add another prop to Item and this prop will be editable, nothing will change and we won't have to check another prop manually... (I hope so)

server/controllers/list.js
```diff
@@ -1135,39 +1142,50 @@ const updateListItem = (req, res) => {
...
+        editedItemActivity = isArchived
+          ? ActivityType.ITEM_ARCHIVE
+          : ActivityType.ITEM_RESTORE;
 
         const data = { listId, userId, itemId, isArchived };
 
         return returnPayload(
-          updateItem(socketInstance, dashboardClients, cohortClients)(data)
+          updateItem(
+            socketInstance,
```
alemikolo: hmm returnPayload is invoke 4 times (and only data is channging (I think?) Can we set Activity type and data inside `ifs` and call this metod once at the end of then?

client/src/modules/list/components/Items/ListItem/ListItem.jsx
adamklepacz: I've been testing and I didn't have this warning any time

client/src/modules/list/components/Items/model/reducer.js
adamklepacz: Yes, it looks nice but did you tested this with a single value coming?
`{name: 'test'} `??

Because I've been trying to use filterUndefined() helper function, but values returned from that funcitons were weird like` {1: {n}, 2:{a}, 3:{m}, 4: {e}}`

server/controllers/list.js
adamklepacz: return payload is invoked only once on a single request. Data is different in every single if so I think it should stay as it is

client/src/modules/list/components/Items/model/reducer.js
alemikolo: it works in cohort case:
```
const cohorts = (state = {}, action) => {
  switch (action.type) {
    case CohortActionTypes.CREATE_SUCCESS:
      return { ...state, [action.payload._id]: { ...action.payload } };
    case CohortActionTypes.UPDATE_SUCCESS: {
      const { cohortId, ...data } = action.payload;
      const previousCohort = state[cohortId];
      const dataToUpdate = filterDefined(data);

      const updatedCohort = {
        ...previousCohort,
        ...dataToUpdate
      };

      return { ...state, [cohortId]: updatedCohort };
    }
    case CohortActionTypes.ARCHIVE_SUCCESS:
```

adamklepacz: yes it works because there is always an object with data comming witn name and description.
so it filters object perfectly. But if the string is passed it doesnt work as expected. 

If you want me to use it, test it

alemikolo: and in locks case: 
```
case CohortHeaderStatusTypes.UNLOCK: {
      const {
        payload: { cohortId, locks }
      } = action;
      const { locks: prevLocks } = state[cohortId];
      const updatedLocks = filterDefined(locks);

      return {
        ...state,
        [cohortId]: {
          ...state[cohortId],
          locks: { ...prevLocks, ...updatedLocks }
        }
      };
    }
```

server/controllers/list.js
alemikolo: Yes, its invoked once per request but there is unnecessary code repetition I think that only data and activity type could be set in ifs. And after those ifs we could write the call of this function once and pass the data that was set before in ifs... If you think it is impossible -resolve this comment

client/src/modules/list/components/Items/model/reducer.js
alemikolo: Yes, you're right, it's alway an object. You merged many methods to UpdateItem and after this merge reducer became too much expanded (especially in 404. branch) I would prefer always send an object and have simple and clean reducer but decide and resolve comments

PR #300, opened 2019-08-30
[EOC-409] [IMPROVEMENT] Parametrize password requirments localized messages

adamklepacz:


client/src/common/utils/helpers.js
```diff
@@ -20,7 +20,14 @@ export const makeAbortablePromise = promise => {
 
 export const dateFromString = string => new Date(string).toLocaleString();
 
-export const validatePassword = value => value.match(/^[^\s]{4,32}$/);
+export const validatePassword = value =>
```
Ciunkos: Please just check the length of the trimmed input:
```javascript
const trimmed = value.trim();
return trimmed.length >= MIN && trimmed.length <= MAX;
```

adamklepacz: We don't want the user to put spaces in their password.

PR #301, opened 2019-08-30
[EOC-415] Add sort imports rule to eslint configuration 

adamklepacz:
After discussion with Przemek, it turned out that we already use the rule that he wanted 😄 😂

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -83,7 +83,7 @@
       "error",
       {
         "groups": [
-          ["external", "builtin"],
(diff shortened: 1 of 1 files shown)


.eslintrc
```diff
@@ -83,7 +83,7 @@
       "error",
       {
         "groups": [
-          ["builtin", "external"],
+          ["external", "builtin"],
```
Ciunkos: `["builtin", "external"]` seems to be better.

PR #302, opened 2019-08-30
[EOC-403] Refactor events emitting from actions to controllers for cohort

alemikolo:


server/controllers/cohort.js
```diff
@@ -122,7 +130,7 @@ const updateCohortById = (req, resp) => {
         return resp.sendStatus(400);
       }
 
-      resp.send();
+      cohort = doc;
```
adamklepacz: you don have to do it like this.
You can use helper function 
```
returnPayload(socketActions.updateCohort(
          socketInstance,
          allCohortsViewClients
        )(data))(doc);
```
and in the last `.then` you will get this doc as payload 
```

.then(doc/payload => { 
 fireAndForget(
        saveActivity(
          cohortActivity,
          userId,
          null,
          null,
          sanitizedCohortId,
          null,
          payload.name
        )
      );	      );
}
```

Please refactor this in all places

server/controllers/cohort.js
```diff
@@ -133,27 +141,63 @@ const updateCohortById = (req, resp) => {
...
+          allCohortsViewClients
+        )(data);
       }
 
       if (isArchived !== undefined) {
-        cohortActivity = isArchived
-          ? ActivityType.COHORT_ARCHIVE
-          : ActivityType.COHORT_RESTORE;
-      }
+        const data = { cohortId };
```
adamklepacz: you can declare this data once, above all if's as let data;
and ovewrite this data in every single if

server/controllers/cohort.js
```diff
@@ -133,27 +141,63 @@ const updateCohortById = (req, resp) => {
...
+
+        const data = { cohortId, name };
+
+        return socketActions.updateCohort(
+          socketInstance,
+          allCohortsViewClients
+        )(data);
       }
 
       if (isArchived !== undefined) {
```
adamklepacz: please use helper `isDefined`

server/controllers/cohort.js
```diff
@@ -122,7 +130,7 @@ const updateCohortById = (req, resp) => {
         return resp.sendStatus(400);
       }
 
-      resp.send();
+      cohort = doc;
 
       if (description !== undefined) {
```
adamklepacz: please add helper isDefined

server/controllers/cohort.js
```diff
@@ -133,27 +141,63 @@ const updateCohortById = (req, resp) => {
...
+
+        return socketActions.restoreCohort(
+          socketInstance,
+          allCohortsViewClients,
+          cohortClients,
+          dashboardClients
+        )(data);
+      }
+    })
+    .then(() => {
```
adamklepacz: after usage of `returnPayload` helper you will get` payload/doc` here

server/controllers/cohort.js
```diff
@@ -206,7 +250,7 @@ const deleteCohortById = (req, resp) => {
     user: { _id: userId }
   } = req;
   const sanitizedCohortId = sanitize(cohortId);
-  let cohortMembers;
+  let cohort;
```
adamklepacz: use `returnPayload` helper

server/controllers/cohort.js
```diff
@@ -441,16 +514,15 @@ const addMember = (req, resp) => {
         );
       }
 
-      const { _id, avatarUrl, displayName } = user;
+      const { _id } = user;
```
adamklepacz: map to userId

server/controllers/cohort.js
```diff
@@ -441,16 +514,15 @@ const addMember = (req, resp) => {
...
         throw new BadRequestException(
           'cohort.actions.add-member-already-member'
         );
       }
-
       currentCohort.memberIds.push(_id);
-      newMember = { avatarUrl, _id, displayName };
+      newMember = user;
 
       return currentCohort.save();
```
adamklepacz: you can also use helper `returnPayload` here to return the actual user data

server/controllers/cohort.js
```diff
@@ -471,19 +543,33 @@ const addMember = (req, resp) => {
...
-        return saveActivity(
-          ActivityType.COHORT_ADD_USER,
-          userId,
-          null,
-          null,
-          sanitizedCohortId,
-          newMember._id
+        userToSend = responseWithCohortMember(newMember, ownerIds);
+
+        return socketActions.addMember(
```
adamklepacz: as well as here.

`returnPayload(promise)(userData);`

server/controllers/cohort.js
```diff
@@ -471,19 +543,33 @@ const addMember = (req, resp) => {
...
+          socketInstance,
+          allCohortsViewClients,
+          dashboardClients
+        )({
+          cohortId,
+          member: userToSend
+        });
+      }
+    })
+    .then(() => {
```
adamklepacz: and here you will get `userData`

server/controllers/cohort.js
alemikolo: In this case, I can't use `returnPayload`, because I have to rebuild this helper and refactor all its usage. In this case, I need cohort data and data return by the promise I passed to this helper.

server/controllers/cohort.js
alemikolo: In this case, I can't use returnPayload, because I have to rebuild this helper and refactor all its usage. In this case, I need cohort data and data return by the promise I passed to this helper.

server/controllers/cohort.js
```diff
@@ -193,13 +198,13 @@ const updateCohortById = (req, resp) => {
...
           null,
-          cohort.name
+          doc.name
         )
       );
 
       resp.send();
     })
-    .catch(() => resp.sendStatus(400));
+    .catch(err => resp.sendStatus(400));
```
adamklepacz: unnecesarry err

server/controllers/cohort.js
```diff
@@ -133,27 +142,67 @@ const updateCohortById = (req, resp) => {
...
+          userId,
+          null,
+          null,
+          sanitizedCohortId,
+          null,
+          doc.name
+        )
       );
+
+      resp.send();
```
Ciunkos: Please return this (assume it is a task/promise - it may not be in this case).

server/controllers/cohort.js
```diff
@@ -441,49 +517,65 @@ const addMember = (req, resp) => {
...
+              viewersIds: { $nin: [newMemberId] }
+            },
+            { $push: { viewersIds: newMemberId } }
+          ).exec()
+        )(user);
       }
     })
-    .then(() => {
-      if (newMember) {
+    .then(user => {
```
Ciunkos: Please use consistent naming - eg. pick one - user or member.

server/sockets/cohort.js
```diff
@@ -187,136 +254,311 @@ const updateCohortHeaderStatus = (socket, cohortClientLocks) => {
...
+      .to(socketId)
+      .emit(CohortActionTypes.DELETE_SUCCESS, { cohortId });
+  }
 
-              socket.broadcast
-                .to(socketId)
-                .emit(CohortActionTypes.FETCH_META_DATA_SUCCESS, {
-                  [cohortId]: cohort
-                });
+  List.find({ cohortId })
```
Ciunkos: `findOne`?

server/sockets/cohort.js
```diff
@@ -187,136 +254,311 @@ const updateCohortHeaderStatus = (socket, cohortClientLocks) => {
...
+        });
+
+        return List.find({ cohortId, isArchived: false })
+          .populate('cohortId', 'ownerIds')
+          .lean()
+          .exec();
+      }
+    })
+    .then(docs => {
+      if (docs && docs.length > 0) {
```
Ciunkos: Only a single list processed? Shouldn't all of them be iterated?

server/sockets/cohort.js
alemikolo: we want to get all cohort's lists

server/sockets/cohort.js
alemikolo: `dosc` is an array of documents, if there is at least one doc (`docs && docs.length > 0`) we process this array

server/sockets/cohort.js
```diff
@@ -1,136 +1,201 @@
...
+  const { cohortId, userId } = data;
 
-const addOwnerRoleInCohort = (socket, clients) => {
-  socket.on(CohortActionTypes.ADD_OWNER_ROLE_SUCCESS, data => {
-    const { cohortId, userId } = data;
+  io.sockets
+    .to(cohortChannel(cohortId))
+    .emit(CohortActionTypes.REMOVE_MEMBER_SUCCESS, data);
+
+  if (allCohortsClients.size > 0) {
```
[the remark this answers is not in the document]
alemikolo: part of this code was he same  so we made helper for that (`emitCohortMetaData`):
```
  if (allCohortsClients.size > 0) {
    emitCohortMetaData(cohortId, allCohortsClients, io);
  }
```
the rest of the code is different, but I made a helper for changing role in cohort because of code was exactly the same

PR #304, opened 2019-09-02
[EOC-447] Set new password show additional info for which account

adamklepacz:


client/src/modules/user/AuthBox/components/PasswordRecoveryForm.jsx
```diff
@@ -25,18 +29,47 @@ class PasswordRecoveryForm extends PureComponent {
...
+      userName: null
     };
 
     this.debouncedPasswordValidation = _debounce(this.passwordValidator, 500);
     this.debouncedComparePasswords = _debounce(this.comparePasswords, 500);
+    this.pendingPromise = null;
+  }
+
+  componentDidMount() {
+    this.handleUserName();
```
marekrozmus: why is this function named with `handle` prefix?

server/controllers/userAuth.js
```diff
@@ -470,11 +470,32 @@ const updateSettings = (req, res) => {
...
+  } = req;
+
+  User.findOne({ resetToken: token })
+    .lean()
+    .exec()
+    .then(user => {
+      if (user) {
+        const { displayName } = user;
+
+        return resp.send({ displayName }).status(200);
```
alemikolo: status(200) is nod needed

PR #305, opened 2019-09-02
[EOC-445] Move list view setting to local storage

alemikolo:


client/src/common/utils/localStorage.js
```diff
@@ -0,0 +1,22 @@
+export const loadSettings = () => {
+  try {
+    const serializedState = localStorage.getItem('settings');
+
+    if (serializedState === null) {
+      return undefined;
+    }
+
+    return JSON.parse(serializedState);
```
marekrozmus: change to single line:
`return serializedState === null ? undefined : JSON.parse(serializedState)`
or just one if:
```
if (serializedState !== null) {
    return JSON.parse(serializedState);
}
```

client/src/common/utils/localStorage.js
```diff
@@ -0,0 +1,22 @@
...
+  } catch (err) {
+    return undefined;
```
marekrozmus: you don't need to return undefined - function does that by default

client/src/app/components/Layout/Layout.jsx
```diff
@@ -97,12 +88,7 @@ export class Layout extends PureComponent {
       viewType: newViewType
     };
 
-    this.handleSwitchView(newViewType);
-    this.setState({ pendingForViewType: true });
-
-    updateSettings(settings)
-      .catch(() => this.handleSwitchView(viewType))
-      .finally(() => this.setState({ pendingForViewType: false }));
+    this.setState({ viewType: newViewType }, saveSettings(settings));
```
Ciunkos: As the callback is asynchronous, please read settings again in the saveSettings callback.

client/src/app/components/Layout/Layout.jsx
```diff
@@ -211,9 +197,11 @@ Layout.propTypes = {
   location: PropTypes.shape({
     pathname: PropTypes.string.isRequired
   }),
+  settings: PropTypes.shape({
+    viewType: PropTypes.string
```
Ciunkos: The type would be `oneOf([ViewType.LIST, ViewType.TILES]).isRequired`.

client/src/common/utils/localStorage.js
```diff
@@ -0,0 +1,22 @@
+export const loadSettings = () => {
+  try {
+    const serializedState = localStorage.getItem('settings');
+
+    if (serializedState === null) {
+      return undefined;
+    }
+
+    return JSON.parse(serializedState);
+  } catch (err) {
```
Ciunkos: Consider at least loggin the error out.

client/src/app/components/Layout/Layout.jsx
alemikolo: I had to remove settings from prop validation because settings are loaded from localStorage directly in Layout component and they aren't passed as a prop from parrent component now.

client/src/common/utils/localStorage.js
alemikolo: we avoid logging anything on the front, and we don't have now any functionality to save some logs for future debugging

client/src/app/components/Layout/Layout.jsx
```diff
@@ -84,25 +78,19 @@ export class Layout extends PureComponent {
...
-    const { updateSettings } = this.props;
     const { viewType } = this.state;
-
     const newViewType =
       viewType === ViewType.LIST ? ViewType.TILES : ViewType.LIST;
 
-    const settings = {
-      viewType: newViewType
-    };
+    this.setState({ viewType: newViewType }, this.handleSaveSettings);
```
adamklepacz: consider handleViewSettings()

client/src/app/components/Layout/Layout.jsx
```diff
@@ -43,6 +44,7 @@ export class Layout extends PureComponent {
   }
 
   componentDidMount() {
+    const settings = loadSettings();
```
adamklepacz: consider loadViewSettings()

server/migrations/20190902121514-0.0.5.js
```diff
@@ -0,0 +1,31 @@
...
+        { settings: { $exists: false } },
+        { $set: { settings: new Settings() } }
+      );
+
+  await runAsyncTasks(addSettingsToUser);
+};
+
+module.exports = {
+  up,
+  down
```
adamklepacz: sort

PR #306, opened 2019-09-03
[EOC-446] Add ellipsis to item name

adamklepacz:
End_of_Coffee_-_track_your_office_inventory-5
End_of_Coffee_-_track_your_office_inventory-4
2 09 2019 eoc_446


client/src/common/components/Confirmation.jsx
```diff
@@ -30,9 +36,9 @@ const Confirmation = ({ className, disabled, onCancel, onConfirm, title }) => (
 );
 
 Confirmation.propTypes = {
+  children: PropTypes.any,
```
alemikolo: string or node? Or maybe:
```
children: PropTypes.oneOfType([
    PropTypes.string,
    PropTypes.node
  ]),
```

adamklepacz: Przemek suggested my last time to for .any for children proptype

alemikolo: ok

PR #312, opened 2019-09-05
[EOC-289] Cannot close the profile menu

adamklepacz:
Bug fixed for all touch devices.
I have also added some to code to close the menu on view change - what was expected behavior from a long time ago.

### DIFF
--- a/client/src/common/components/Dropdown.jsx
+++ b/client/src/common/components/Dropdown.jsx
@@ -21,9 +21,11 @@ class Dropdown extends PureComponent {
     if (isVisible) {
       document.addEventListener('keydown', this.handleEscapePress);
       document.addEventListener('click', this.handleClickOutside);
+      document.addEventListener('touchend', this.handleClickOutside);
     } else {
       document.removeEventListener('keydown', this.handleEscapePress);
       document.removeEventListener('click', this.handleClickOutside);
+      document.removeEventListener('touchend', this.handleClickOutside);
     }
   }
 
--- a/client/src/common/components/Dropdown.jsx
+++ b/client/src/common/components/Dropdown.jsx
@@ -29,14 +29,14 @@ class Dropdown extends PureComponent {
       document.addEventListener('keydown', this.handleEscapePress);
       document.addEventListener('click', this.handleClickOutside);
       document.addEventListener('touchend', this.handleClickOutside);
+
+      if (pathname !== prevPathname) {
+        this.hide();
+      }
     } else {
       document.removeEventListener('keydown', this.handleEscapePress);
       document.removeEventListener('click', this.handleClickOutside);
       document.removeEventListener('touchend', this.handleClickOutside);
-    }
-
-    if (pathname !== prevPathname) {
-      this.hideDropdown();
     }
   }
 
@@ -45,15 +45,17 @@ class Dropdown extends PureComponent {
     document.removeEventListener('click', this.handleClickOutside);
   }
 
-  hideDropdown = () => this.setState({ isVisible: false });
+  setVisibility = isVisible => this.setState({ isVisible });
 
-  showDropdown = () => this.setState({ isVisible: true });
+  hide = () => this.setVisibility(false);
+
+  show = () => this.setVisibility(true);
 
   handleEscapePress = event => {
     const { code } = event;
 
     if (code === KeyCodes.ESCAPE) {
-      this.hideDropdown();
+      this.hide();
     }
   };
 
@@ -61,7 +63,7 @@ class Dropdown extends PureComponent {
     const isClickedOutside = !this.dropdown.current.contains(event.target);
 
     if (isClickedOutside) {
-      this.hideDropdown();
+      this.hide();
     }
   };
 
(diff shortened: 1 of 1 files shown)


client/src/common/components/Dropdown.jsx
```diff
@@ -15,15 +16,27 @@ class Dropdown extends PureComponent {
...
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps) {
     const { isVisible } = this.state;
+    const {
+      location: { pathname }
+    } = this.props;
+    const {
+      location: { pathname: prevPathname }
+    } = prevProps;
```
alemikolo: I'm still not sure if we should bind the dropdown component with additional libraries. With this solution, we will be able to use it only in apps that use react-router and only there, where changing the route should close the dropdown.

adamklepacz: As I said, and Przemo agreed with my it is a bad practise to control visibility of component in two places (inside component and additional prop from outside).

I know what you mean, and I think if someone want to use this component in future, and he will copy and paste this it will be clear thath he has to remove with router statement and componentDidUpdate method.

And the one last thing, in my opinion this component now is designed in a wrong way.If we want to make generic components, that can be used across many other projects we should do this in a diffrent way. It should aways take visbility information from outside, from parrent component. It should have isVisible property which means that developer has to provide this property from outside.

```

<Button onClick={isVisible ? showDropdown : hideDropdown } />
<Dropdown isVisible={true || false}>
   {children}
</Dropdown>
```

and in this way, dropdown component is as Dan Abramov said "dump" functional component responsible only for rendering data, but he don't care anything about internal state etc.
I know also what was the idea of working component "out of the box", but I don't think it was done properly. 
For now I in my opinion it works as expected. And I totally agree with you @alemikolo , that it's not generic

adamklepacz: The second solution to work this out of the box is to create two dependent components.
DropdownController which will have internal state to control the visibility of the Dropdown component. And a Dropdown component will still be as I said before a "dump" functional component to render data only.

And in this way, we have "out of the box" dropdown solution, Where they can be used together or just Dropdown component independently. 

```
<DropdownController>
    <button onClick={isVisible ? showDropdown : hideDropdown}>

    <Dropdown isVisible={}>
          {children}
    <Dropdown>
<DropdownController>
```

client/src/common/components/Dropdown.jsx
```diff
@@ -15,15 +16,27 @@ class Dropdown extends PureComponent {
...
       document.addEventListener('keydown', this.handleEscapePress);
       document.addEventListener('click', this.handleClickOutside);
+      document.addEventListener('touchend', this.handleClickOutside);
     } else {
       document.removeEventListener('keydown', this.handleEscapePress);
       document.removeEventListener('click', this.handleClickOutside);
+      document.removeEventListener('touchend', this.handleClickOutside);
+    }
+
+    if (pathname !== prevPathname) {
```
Ciunkos: Please include `isVisible` check to not spam the state changes, i.e. `isVisible && parhname !== previousPathname`.

client/src/common/components/Dropdown.jsx
```diff
@@ -32,25 +45,26 @@ class Dropdown extends PureComponent {
     document.removeEventListener('click', this.handleClickOutside);
   }
 
+  hideDropdown = () => this.setState({ isVisible: false });
```
Ciunkos: This can be simplified as:
```javascript
setVisible = isVisible => () => this.setState({ isVisible });

showDropdown = this.setVisible(true);
hideDropdown = this.setVisible(false);
```

client/src/common/components/Dropdown.jsx
```diff
@@ -32,25 +45,26 @@ class Dropdown extends PureComponent {
...
+
+  showDropdown = () => this.setState({ isVisible: true });
```
Ciunkos: This is a Dropdown component, IMHO this should be simplified to `show/hide` as there is nothing else than the dropdown to be shown/hidden.

client/src/common/components/Dropdown.jsx
```diff
@@ -65,7 +79,7 @@ class Dropdown extends PureComponent {
       <div className="dropdown">
         <button
           className={classNames(buttonClassName)}
-          onClick={this.handleDropdownVisibility}
+          onClick={isVisible ? this.hideDropdown : this.showDropdown}
```
Ciunkos: Consider extracting this before the `return (` as `const toggle = isVisible ? this.hide : this.show`.

client/src/common/components/Dropdown.jsx
```diff
@@ -94,7 +108,10 @@ Dropdown.propTypes = {
     .isRequired,
   children: PropTypes.oneOfType([PropTypes.node, PropTypes.string]).isRequired,
   dropdownClassName: PropTypes.string,
-  dropdownName: PropTypes.string
+  dropdownName: PropTypes.string,
+  location: PropTypes.shape({
+    pathname: PropTypes.string.isRequired
+  })
```
Ciunkos: The `location` prop is required as well (destructured without a fallback value) - please add `.isRequired`.

client/src/common/components/Dropdown.jsx
```diff
@@ -94,7 +108,10 @@ Dropdown.propTypes = {
...
 };
 
-export default Dropdown;
+export default withRouter(props => <Dropdown {...props} />);
```
Ciunkos: Just `withRouter(Dropdown)`.

PR #315, opened 2019-09-07
[EOC-308] Moving item between lists

alemikolo:
move-items


server/controllers/list.js
```diff
@@ -1513,6 +1512,165 @@ const leaveList = (req, resp) => {
...
+
+  const query = {
+    memberIds: userId,
+    isArchived: false
+  };
+
+  List.find(query, 'name')
+    .lean()
+    .exec()
+    .then(docs => {
```
marekrozmus: use just one line ternary operator:
`docs ? res.send(docs) : resp.sendStatus(400)`

server/controllers/list.js
```diff
@@ -1513,6 +1512,165 @@ const leaveList = (req, resp) => {
...
+        isArchived,
+        isDeleted,
+        isOrdered,
+        locks: { description: false, name: false },
+        name,
+        purchaserId,
+        status,
+        updatedAt,
+        voterIds
+      });
```
marekrozmus: `const newItem = new Item({ ...items.id(sanitizedItemId).lean(), locks: { description: false, name: false }})`?

alemikolo: It can't be done with `lean()` but I fix this in other way:
```
const { items } = oldList;
      const newItem = new Item({
        ...items.id(sanitizedItemId)._doc,
        locks: { description: false, name: false }
      });
```

client/src/modules/list/components/Items/MoveToListPanel/MoveToListPanel.jsx
```diff
@@ -0,0 +1,207 @@
+import React, { Fragment, PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { withRouter } from 'react-router-dom';
+import _flowRight from 'lodash/flowRight';
+import { FormattedMessage, injectIntl } from 'react-intl';
+
+import Filter from 'common/components/Filter';
+import { fetchListsForItem } from 'modules/list/model/actions';
```
adamklepacz: 1)Please change the name, I would go for 'fetchListsForUser'
beacuse I can see in the controller that you are fetching lists for userId, not for itemId.
2) Can't you just use "fetchListsMetaData" logic?? Instead of creating a second controllers, actions etc ??

client/src/modules/list/model/selectors.js
```diff
@@ -108,3 +108,9 @@ export const getMembers = createSelector(
   getList,
   list => list && list.members
 );
+
+export const getListsForItem = createSelector(
```
adamklepacz: I would go for "getListsNames", as this is what you are returning here

client/src/modules/list/components/Items/MoveToListPanel/MoveToListPanel.jsx
```diff
@@ -0,0 +1,207 @@
...
+    if (lists.length > 0) {
+      this.setState({ listsToDisplay: lists });
+    }
+  }
+
+  componentDidUpdate(prevProps) {
+    const { lists: prevLists } = prevProps;
+    const { lists } = this.props;
+
+    if (prevLists.length === 0 && lists.length > 0) {
```
adamklepacz: this can be simplified to 
if(!prevLists.length && lists.lenght)

client/src/modules/list/components/Items/MoveToListPanel/MoveToListPanel.jsx
```diff
@@ -0,0 +1,207 @@
...
+        params: { id: listId }
+      },
+      moveItemToList,
+      unlockItem
+    } = this.props;
+    const notificationData = { item: itemName, list: listName };
+
+    lockItem();
+
+    return moveItemToList(itemId, listId, newListId, notificationData).finally(
```
adamklepacz: I would go for `moveItemToList(itemId, listId, newListId)(notificationData)`
where two-argument groups represent different context, first is related to moving item logic, second related to notification shown

client/src/modules/list/components/Items/MoveToListPanel/MoveToListPanel.jsx
```diff
@@ -0,0 +1,207 @@
...
+              values={{ item: itemName, list: listName }}
+            />
+          </Confirmation>
+        ) : (
+          <Fragment>
+            <div className="move-to-list-panel__filter-box">
+              <Filter
+                buttonContent={<CloseIcon />}
+                classes="move-to-list-panel__filter"
+                elements={lists}
```
adamklepacz: options

client/src/modules/list/model/reducer.js
```diff
@@ -106,6 +106,8 @@ const lists = (state = {}, action) => {
   switch (action.type) {
     case ListActionTypes.FETCH_ARCHIVED_META_DATA_SUCCESS:
       return { ...state, ...action.payload };
+    case ListActionTypes.FETCH_FOR_ITEM_SUCCESS:
```
adamklepacz: check my previous comments and consider using FETCH_META_DATA

client/src/modules/notification/components/Notification.jsx
```diff
@@ -7,15 +7,22 @@ import { MessageType as NotificationType } from 'common/constants/enums';
 
 const Notification = ({ data, id, redirect, type }) => (
   <MessageBox type={type}>
-    <FormattedMessage id={id} values={{ data: <em>{data}</em> }} />
+    <FormattedMessage
+      id={id}
+      values={typeof data === 'string' ? { data } : { ...data }}
```
adamklepacz: 1) please extract this above the return
2) Check if we can always return object to avoid such things (if this is a lot of work, create separated task and add this comment into description)

server/common/utils/helpers.js
```diff
@@ -388,7 +388,8 @@ const runAsyncTasks = async (...tasks) => {
 /* eslint-disable-next-line no-console */
 const fireAndForget = promise => promise.catch(err => console.error(err));
 
-const returnPayload = promise => payload => promise.then(() => payload);
+const returnPayload = (promise, returnResult = false) => payload =>
+  promise.then(result => (returnResult ? { payload, result } : payload));
```
adamklepacz:
```
const returnPayload = promise => (promiseResult = false) =>  payload =>
  promise.then(result => (returnResult ? { payload, result } : payload));
```

server/controllers/list.js
```diff
@@ -1513,6 +1512,134 @@ const leaveList = (req, resp) => {
...
+    memberIds: userId,
+    isArchived: false
+  };
+
+  List.find(query, 'name')
+    .lean()
+    .exec()
+    .then(docs => (docs ? resp.send(docs) : resp.sendStatus(400)))
+    .catch(() => resp.sendStatus(400));
+};
```
adamklepacz: Check if this controller can be removed while using "fetchMetaData" logic

client/src/modules/list/components/Items/MoveToListPanel/MoveToListPanel.jsx
alemikolo: Ad 2) I tried to use fetchListMetaData logic but I realized that I have to get lists in which the user is at least a member, not viewer because only as a member I can manage list's items

server/common/utils/helpers.js
alemikolo: I restored to the previous version because we will be replacing this method with simple `thens` or `async`-`awaits`

client/src/modules/list/components/Items/MoveToListPanel/MoveToListPanel.jsx
alemikolo: I can't be done like this, because this is `moveItemToList(itemId, listId, newListId)(dispatch)` because of thunk

client/src/modules/notification/components/Notification.jsx
alemikolo: 1. Done.
2. Task created

client/src/common/components/Filter.jsx
```diff
@@ -23,40 +23,47 @@ class Filter extends PureComponent {
...
+
+    this.setState({ query: '' }, () => onFilter(options));
   };
 
   handleInputChange = event => {
     const { value } = event.target;
     const query = _trim(value);
 
-    this.setState(() => ({ query }), this.handleFilterElements);
+    this.setState(() => ({ query }), this.handleFilterOptions);
```
adamklepacz: handleFilteredOptions

client/src/common/components/Filter.jsx
```diff
@@ -79,17 +86,17 @@ class Filter extends PureComponent {
...
 Filter.propTypes = {
   buttonContent: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
   classes: PropTypes.string.isRequired,
-  elements: PropTypes.arrayOf(
+  options: PropTypes.arrayOf(
     PropTypes.shape({ _id: PropTypes.string, name: PropTypes.string })
   ).isRequired,
   placeholder: PropTypes.string,
-  resetButton: PropTypes.bool,
+  clearFilterButton: PropTypes.bool,
```
adamklepacz: sort props

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -138,20 +138,28 @@ class ListItem extends PureComponent {
...
 
-    this.setState(({ isConfirmationVisible }) => ({
-      isConfirmationVisible: !isConfirmationVisible
-    }));
+    this.setState({ isConfirmationVisible: false });
   };
 
-  handleMoveToListPanelVisibility = event => {
+  showConfirmation = event => {
     event.preventDefault();
```
adamklepacz: do we need this preventDeafult() calls? 
do we have any additional effects on those cancels, confirm buttons ??

client/src/modules/list/components/Items/MoveToListPanel/MoveToListPanel.jsx
```diff
@@ -137,7 +137,7 @@ class MoveToListPanel extends PureComponent {
               <Filter
                 buttonContent={<CloseIcon />}
                 classes="move-to-list-panel__filter"
-                elements={lists}
+                options={lists}
                 onFilter={this.handleFilterLists}
```
adamklepacz: sort props

client/src/modules/list/model/actionTypes.js
```diff
@@ -11,7 +11,7 @@ export const ListActionTypes = enumerable('list')(
   ...asyncTypes('FAVOURITES'),
   ...asyncTypes('FETCH_ARCHIVED_META_DATA'),
   ...asyncTypes('FETCH_DATA'),
-  ...asyncTypes('FETCH_FOR_ITEM'),
+  ...asyncTypes('FETCH_AVAILABLE'),
```
adamklepacz: sort

PR #318, opened 2019-09-09
[EOC-456] Cannot resend link to reset password

adamklepacz:
Mariusz found a case where the user clicks the reset email once again after he just Reseted his password.

When user updates password via reset password, his reset token gets revoked, so using the same link to resend link will not work and this is expected behavior. One thing that needed to be improved was the message displayed to the user.

Now when he uses the same link to reset, after he just updated his password, he will get a more accurate message. 
Please check EOC-456 task in Jira, I have attached a screencast file showing how does it work right now.

As Przemo said our token logic can be improved using JWT library, but this has to be discussed more cause it will be a lot of work to re-implement our current logic


client/src/locales/en.json
```diff
@@ -122,6 +123,7 @@
   "user.actions.password-recovery-success": "Password successfully updated. Please go to the {link} and sign in to app.",
   "user.actions.recovery-password-default-error": "Account does not exist or is not activated.",
   "user.actions.reset": "An email with instructions to reset password has been sent to {data} if it is associated with an EOC account. Your existing password has not been changed.",
+  "user.actions.token-revoked": "This reset link link has expired. To reset your password please go to the {link}.",
```
marekrozmus: `link link` - remove extra `link`

adamklepacz:
@marekrozmus  
In a "normal" way where the user resets his password, but one hour passed he will get this view with the "Resend link" button. 
When he clicks it, he will get a new email with a link to update his password.

But Mariusz found an "edge case" which I described in the description of this PR. 
If you have an any better idea for this, let me know

Ciunkos:
I would go for the removal of the "Resend" button and just display - "The link has expired. Please try again". This is just an edge case and I think there are more important fields of the app that can be improved instead of this.

adamklepacz:
> I would go for the removal of the "Resend" button and just display - "The link has expired. Please try again". This is just an edge case and I think there are more important fields of the app that can be improved instead of this.

I think that the best approach. Now it works like this:
- if a token has expired, and he will click the link in the mail - he can "Resend link" as normal,
- if he updates the password or he clicks a very old link - he will only get the message ` The link has expired....`

Zrzut ekranu 2019-09-9 o 17 36 53

server/controllers/userAuth.js
```diff
@@ -465,8 +457,27 @@ const getUserName = (req, resp) => {
...
 
+const checkToken = (req, resp) => {
+  const {
+    params: { token }
+  } = req;
+
+  User.findOne({ resetToken: token })
+    .lean()
+    .exec()
+    .then(user => {
```
marekrozmus: just single line:
`user => user ? resp.send() : resp.sendStatus(400)`

client/src/locales/en.json
```diff
@@ -102,12 +102,13 @@
...
-  "user.auth.reset-password.heading": "Reset your password",
-  "user.auth.reset-password.email-label": "Enter email",
-  "user.auth.reset-password.button-content": "Reset password",
-  "user.auth.pass-recovery-form.heading": "Set new password",
   "user.auth.pass-recovery-form.heading-user-name": "Set new password for {name}",
+  "user.auth.pass-recovery-form.heading": "Set new password",
+  "user.auth.reset-password.button-content": "Reset password",
+  "user.auth.reset-password.email-label": "Enter email",
+  "user.auth.reset-password.heading": "Reset your password",
+  "user.auth.sign-up.reset-password": "reset password form",
```
Ciunkos: Please use sentence case.

client/src/locales/en.json
```diff
@@ -102,12 +102,13 @@
...
+  "user.auth.sign-up.result-link": "sign in/sign up page",
```
Ciunkos: Please use sentence case.

client/src/locales/en.json
```diff
@@ -122,6 +123,7 @@
   "user.actions.password-recovery-success": "Password successfully updated. Please go to the {link} and sign in to app.",
   "user.actions.recovery-password-default-error": "Account does not exist or is not activated.",
   "user.actions.reset": "An email with instructions to reset password has been sent to {data} if it is associated with an EOC account. Your existing password has not been changed.",
+  "user.actions.token-revoked": "This reset link has expired. To reset your password please go to the {link}.",
```
Ciunkos: I would just leave the `Please try again` with no link in this case.

server/controllers/userAuth.js
```diff
@@ -465,8 +457,21 @@ const getUserName = (req, resp) => {
...
 
+const checkToken = (req, resp) => {
+  const {
+    params: { token }
+  } = req;
+
+  User.findOne({ resetToken: token })
+    .lean()
+    .exec()
+    .then(user => (user ? resp.send() : resp.sendStatus(400)))
```
Ciunkos: Please vary only the status code with `resp.sendStatus(user ? 200 : 400)`.

client/src/modules/user/AuthBox/components/LinkExpired.jsx
```diff
@@ -14,7 +16,24 @@ import { Routes } from 'common/constants/enums';
...
+  }
+
+  checkToken = () => {
+    const {
+      match: {
+        params: { token }
+      }
+    } = this.props;
+
+    checkToken(token)
```
Ciunkos: Token validation complicates this IMHO. Consider just returning a proper error message from the backend when consuming the token.

PR #321, opened 2019-09-10
[EOC-272] [BUG] Long name for the item is out of the box

adamklepacz:
Fixed
Zrzut ekranu 2019-09-10 o 09 25 58

### DIFF
--- a/client/src/modules/list/components/Items/ListArchivedItem/ListArchivedItem.scss
+++ b/client/src/modules/list/components/Items/ListArchivedItem/ListArchivedItem.scss
@@ -29,7 +29,6 @@
   }
 
   &__name {
-    border: 0;
     color: $text-dark;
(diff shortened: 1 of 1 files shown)


client/src/modules/list/components/Items/ListArchivedItem/ListArchivedItem.scss
```diff
@@ -29,8 +29,16 @@
...
 
   &__name {
-    font-weight: 500;
     color: $text-dark;
+    font-family: 'Montserrat', sans-serif;
+    font-size: 18px;
+    font-weight: 500;
+    line-height: 1.2;
+    margin-bottom: 4px;
+    max-width: 210px;
```
Ciunkos: Please also add `title={originalText}` attribute so the full name is visible on hover (uses native tooltip).

PR #322, opened 2019-09-10
[EOC-441] Changing list type. Member of the list gets no update about members when member box is open

alemikolo:


client/src/modules/list/components/ListHeader.jsx
```diff
@@ -57,23 +57,35 @@ class ListHeader extends PureComponent {
...
     const {
-      details: { name, description }
+      details: { name, description, type }
     } = this.props;
     const {
-      details: { name: previousName, description: previousDescription }
+      details: {
+        name: previousName,
+        description: previousDescription,
+        type: prevType
```
Ciunkos: Please use consistent naming - `previousType`.

PR #323, opened 2019-09-10
[EOC-470] Add additional info about account during recovery password

alemikolo:
Screenshot 2019-09-10_10-33-12-781

### DIFF
--- a/client/src/modules/user/AuthBox/components/PasswordRecoveryForm.jsx
+++ b/client/src/modules/user/AuthBox/components/PasswordRecoveryForm.jsx
@@ -6,7 +6,7 @@ import _debounce from 'lodash/debounce';
 import validator from 'validator';
 
 import { RouterMatchPropType, IntlPropType } from 'common/constants/propTypes';
-import { getUserName, updatePassword } from 'modules/user/model/actions';
+import { getAccountData, updatePassword } from 'modules/user/model/actions';
 import ValidationInput from './ValidationInput';
 import PendingButton from 'common/components/PendingButton';
 import {
@@ -20,6 +20,7 @@ class PasswordRecoveryForm extends PureComponent {
     super(props);
 
     this.state = {
+      email: null,
       errors: {
         comparePasswordsError: '',
         passwordError: '',
@@ -39,7 +40,7 @@ class PasswordRecoveryForm extends PureComponent {
   }
 
   componentDidMount() {
-    this.fetchUserName();
+    this.fetchAccountData();
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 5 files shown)


client/src/modules/user/AuthBox/components/PasswordRecoveryForm.jsx
```diff
@@ -181,10 +185,10 @@ class PasswordRecoveryForm extends PureComponent {
     return (
       <form className="pass-recovery-form" onSubmit={this.handleSubmit}>
         <h2 className="pass-recovery-form__heading">
```
Ciunkos: Please increase the line-height so the multiline text is more readable.

client/src/modules/user/AuthBox/components/PasswordRecoveryForm.jsx
```diff
@@ -181,10 +185,10 @@ class PasswordRecoveryForm extends PureComponent {
...
-          {userName ? (
+          {userName && email ? (
```
Ciunkos: Can the userName ever be an empty string `''`?

client/src/modules/user/model/actions.js
```diff
@@ -146,5 +146,5 @@ export const changePassword = (password, newPassword, newPasswordConfirm) =>
     newPasswordConfirm
   });
 
-export const getUserName = token =>
-  getJson(`/auth/user-name/${token}`).then(({ displayName }) => displayName);
+export const getAccountData = token =>
+  getJson(`/auth/account-data/${token}`).then(data => data);
```
Ciunkos: No need for `then` as it does nothing in this case.

```suggestion
  getJson(`/auth/account-data/${token}`);
```

server/routes/authorization.js
```diff
@@ -54,6 +54,6 @@ router.post(
 );
 router.get('/user-details', authorize, getUserDetails);
 router.post('/change-password', authorize, changePassword);
-router.get('/user-name/:token?', getUserName);
+router.get('/account-data/:token?', getAccountData);
```
Ciunkos: Consider renaming to `account-details` to match `user-details` route.

client/src/modules/user/model/actions.js
```diff
@@ -146,5 +146,5 @@ export const changePassword = (password, newPassword, newPasswordConfirm) =>
     newPasswordConfirm
   });
 
-export const getUserName = token =>
-  getJson(`/auth/user-name/${token}`).then(({ displayName }) => displayName);
+export const getAccountData = token =>
```
Ciunkos: Consider renaming `getAccountData` to `getAccountDetails`.

PR #327, opened 2019-09-10
[HOT-FIX] Display notification with ellipsis if text is too long

alemikolo:


client/src/modules/notification/components/Notification.jsx
```diff
@@ -4,9 +4,13 @@ import { FormattedMessage } from 'react-intl';
 
 import MessageBox from 'common/components/MessageBox';
 import { MessageType as NotificationType } from 'common/constants/enums';
+import { mapObject } from 'common/utils/helpers';
```
adamklepacz: why not lodash _map ??
https://lodash.com/docs/4.17.15#map

alemikolo: Because lodash's _map returns an array. We need an object. This helper (we use it on backend already) returns an object with all values modyfied by callback you pass as an argument

client/src/modules/notification/components/Notification.jsx
```diff
@@ -4,9 +4,13 @@ import { FormattedMessage } from 'react-intl';
...
 
 const Notification = ({ data, id, redirect, type }) => {
-  const values = typeof data === 'string' ? { data } : { ...data };
+  const values =
+    typeof data === 'string'
+      ? { data: <em>{data}</em> }
+      : { ...mapObject(value => <em>{value}</em>)(data) };
```
Ciunkos: This can be simplified to:
```javascript
const wrap = value => <em>{value}</em>;
const dataSource = typeof data === 'string' ? { data } : data;
const values = mapObject(wrap)(dataSource);
```

client/src/modules/notification/components/Notification.jsx
```diff
@@ -4,9 +4,12 @@ import { FormattedMessage } from 'react-intl';
 
 import MessageBox from 'common/components/MessageBox';
 import { MessageType as NotificationType } from 'common/constants/enums';
+import { mapObject } from 'common/utils/helpers';
 
 const Notification = ({ data, id, redirect, type }) => {
-  const values = typeof data === 'string' ? { data } : { ...data };
+  const wrap = value => <em>{value}</em>;
```
Ciunkos: This can be lifted up from the `Notification` function.

PR #332, opened 2019-09-12
[EOC-464] Joining and leaving rooms for all available resources on entering or exiting the app.

alemikolo:
1. On entering the app, on socket connection, the user joins all metadata rooms for resources he has access to.
2. To test if joining works I refactor updateCohort to emit data for new cohort meta-data room. 
3. To prevent dispatching an action twice for a user who performed this action I create redux middleware which prevents dispatching an action if the user's id is the same as the performer's id.
4. To clear unnecessary data that the user gets on views like 'about' or 'privacy policy' I check for changing the route in layout component and clear lists or cohorts data if that change occurs. I tried to not saving unnecessary data but logic was complicated. I had to compare the current route (we have no access to react-router outside of component) with event type to decide if that data has to be saved or not.
5. Leaving rooms on exiting an app is unnecessary because: "Upon disconnection, sockets leave all the channels they were part of automatically, and no special teardown is needed on your part." [docs](https://socket.io/docs/rooms-and-namespaces/)

### DIFF
--- a/client/src/app/model/middleware.js
+++ b/client/src/app/model/middleware.js
@@ -1,0 +1,17 @@
+export const preventDispatchForPerformer = store => next => action => {
+  if (action.payload) {
+    const {
+      payload: { performerId, ...rest },
+      type
+    } = action;
+    const { currentUser: user } = store.getState();
+
+    if (user && performerId === user.id) {
+      return;
+    }
+
+    return next({ type, payload: { ...rest } });
+  }
+
+  return next(action);
+};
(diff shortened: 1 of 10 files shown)


client/src/app/model/middleware.js
```diff
@@ -0,0 +1,23 @@
+export const preventDispatchForPerformer = store => next => action => {
+  if (typeof action === 'function') {
+    return next(action);
+  }
+  const { payload } = action;
+
+  if (typeof payload === 'object') {
```
adamklepacz: I think this if is unnecessary

alemikolo: For now, we have to keep it. We have actions that payload isn't an object eg. remove notification.

PR #335, opened 2019-09-12
[EOC-465] Joining leaving room on adding, removing or leaving in list

alemikolo:
To test if joining and leaving room works correctly on adding/removing the user or leaving list I refactored adding/removing user and leaving list methods also.


server/sockets/index.js
```diff
@@ -69,38 +70,54 @@ const socketListeners = socketInstance => {
...
+            break;
+          case Routes.COHORT:
+            socket.join(roomId);
+            cohortViewClients.set(userId, { socketId: socket.id, viewId });
+            break;
+          default:
+            break;
+        }
+
+        return;
```
marekrozmus: if here is just `return` you could `return` in every `case` and `default`

server/sockets/index.js
```diff
@@ -69,38 +70,54 @@ const socketListeners = socketInstance => {
...
+            break;
+          case Routes.COHORT:
+            socket.leave(roomId);
+            cohortViewClients.delete(userId);
+            break;
+          default:
+            break;
+        }
+
+        return;
```
marekrozmus: same here - see previous comment

client/src/common/utils/helpers.js
```diff
@@ -44,9 +44,11 @@ const fromEntries = convertedArray =>
 export const asyncTypes = key =>
   asyncPostfixes.map(postfix => [key, postfix].join('_'));
 
-export const enumerable = namespace => (...keys) =>
+export const enumerable = (namespace = null) => (...keys) =>
```
adamklepacz: It would be nice to add some jsDoc with example to this function cause it is pretty hard to understand

https://devdocs.io/jsdoc/howto-es2015-modules
I did something like this in some places in project so you can copy the pattern

client/src/modules/list/model/actions.js
```diff
@@ -433,14 +433,14 @@ export const addListViewer = (listId, email) => dispatch =>
...
-        const action = addViewerSuccess({ listId, ...json });
+    .then(viewer => {
+      if (viewer._id) {
+        const action = addViewerSuccess({ listId, viewer });
 
         dispatch(action);
         createNotificationWithTimeout(dispatch, NotificationType.SUCCESS, {
           notificationId: 'list.actions.add-viewer',
-          data: json.displayName
+          data: viewer.displayName
```
adamklepacz: when merging remember to change all data properties to objects
`data {name: displayName}`

client/src/sockets/enums.js
```diff
@@ -47,7 +48,6 @@ export const ListEvents = Object.freeze({
   REMOVE_MEMBER_SUCCESS: ListActionTypes.REMOVE_MEMBER_SUCCESS,
   REMOVE_WHEN_COHORT_UNAVAILABLE: 'list/REMOVE_WHEN_COHORT_UNAVAILABLE',
   REMOVE_OWNER_ROLE_SUCCESS: ListActionTypes.REMOVE_OWNER_ROLE_SUCCESS,
-  REMOVED_BY_SOMEONE: 'list/REMOVED_BY_SOMEONE',
```
adamklepacz: this one is not needed anymore??

server/common/utils/helpers.js
```diff
@@ -353,9 +353,11 @@ const responseWithCohortDetails = (doc, userId) => {
  */
 const isDefined = x => x !== undefined;
 
-const enumerable = namespace => (...keys) =>
+const enumerable = (namespace = null) => (...keys) =>
```
adamklepacz: same here, I would add some jsDoc comment

server/sockets/index.js
```diff
@@ -69,38 +70,57 @@ const socketListeners = socketInstance => {
...
-        case Routes.COHORT:
-          socket.join(roomId);
-          cohortViewClients.set(userId, { socketId: socket.id, viewId });
-          break;
-        default:
-          break;
+    socket.on(AppEvents.JOIN_ROOM, emittedData => {
+      // This check is for the old functionality to work.
+      // After refactoring this handler will be simplified to:
+      // socket.on('joinRoom', room => socket.join(room));
```
adamklepacz: when adding block comments user block comments. Just type in VSCode   -> `/**` and hit enter
```

/**
 * This check is for the old functionality to work.
 * After refactoring this handler will be simplified to:
 * switch (room) {	      // socket.on('joinRoom', room => socket.join(room));
 */
```

and please add sentence  " Switch statement will be removed"

server/sockets/index.js
```diff
@@ -69,38 +70,57 @@ const socketListeners = socketInstance => {
...
+      if (typeof emittedData === 'object') {
```
adamklepacz: are these checks necessary?

server/sockets/index.js
```diff
@@ -69,38 +70,57 @@ const socketListeners = socketInstance => {
...
-          listViewClients.delete(userId);
-          break;
-        case Routes.COHORT:
-          socket.leave(roomId);
-          cohortViewClients.delete(userId);
-          break;
-        default:
-          break;
+    socket.on(AppEvents.LEAVE_ROOM, emittedData => {
+      if (typeof emittedData === 'object') {
```
adamklepacz: are these checks necessary?

server/sockets/index.js
alemikolo: this checks will be removed there are for the old functionality to work. I wrote this in the comment

server/sockets/index.js
alemikolo: this checks will be removed there are for the old functionality to work. I wrote this in the comment

client/src/sockets/enums.js
alemikolo: jest it's not needed, all work is done in simple REMOVE_MEMBER_SUCCESS

client/src/common/utils/helpers.js
```diff
@@ -44,6 +44,14 @@ const fromEntries = convertedArray =>
 export const asyncTypes = key =>
   asyncPostfixes.map(postfix => [key, postfix].join('_'));
 
+/**
+ * Create object with key/value pairs.
+ * If the namespace is missing value is equal key
```
adamklepacz: equal to the key

PR #345, opened 2019-09-17
EOC-482 [BUG]Update members on list view when owner is removed

alemikolo:
Remove list owner controller wasn't associated with sockets and didn't emit socket events

### DIFF
--- a/server/controllers/list.js
+++ b/server/controllers/list.js
@@ -693,9 +693,9 @@ const removeOwner = (req, resp) => {
         userId: sanitizedUserId
       };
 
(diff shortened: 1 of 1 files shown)


server/controllers/list.js
```diff
@@ -687,6 +687,17 @@ const removeOwner = (req, resp) => {
         return resp.sendStatus(400);
       }
 
+      const data = {
+        listId,
+        performerId: currentUserId,
+        userId: sanitizedUserId
+      };
+
+      return returnPayload(socketActions.removeViewer(socketInstance)(data))(
```
marekrozmus: use `then` instead of `returnPayload`

PR #346, opened 2019-09-19
[EOC-374][EOC-468] Bug highlight item only when added cloned and restored/ Emit item data to proper rooms

alemikolo:
Animations are controlled by props stored in animations reducer.

I had to split 'update item' action, 'update item' controller and 'update item' socket` for separate methods to have separate redux's actions types and display animation properly. 

I had to create separate controllers and socket emitting methods for archiving, restoring, marking as done marking as unhandled and updating item. Because I had to create or refactor 3/4 of items' methods I decided to refactor the rest methods in the same way. Now this PR concerns tasks: **[EOC-374]**, **[EOC-468]**.

- All items controllers use async/await to avoid creating variables in the global controller's scope or passing them via 'returnPayload' method.
- Sending response is independent of socket's emission if emission failed response will be sent anyway (sockets are not the main functionality of the app).
- All controllers send data that is needed only. They pass to the sockets only data that is needed and socket send only data that was changed.
- Sockets don't use clientsMaps, they send data to rooms or to individual sockets obtained from DB.


client/src/modules/list/components/ArchivedItemsContainer.jsx
```diff
@@ -77,7 +81,9 @@ class ArchivedItemsContainer extends PureComponent {
             })}
           >
             <ItemsContainer
+              animate={animate}
               archived
+              disableAnimations={disableAnimationForArchivedItems}
```
marekrozmus: `onDisableAnimations'

client/src/modules/list/components/Items/ItemsList.jsx
```diff
@@ -103,7 +113,9 @@ class ItemsList extends PureComponent {
 }
 
 ItemsList.propTypes = {
+  animate: PropTypes.bool,
   archived: PropTypes.bool,
+  disableAnimations: PropTypes.func.isRequired,
```
marekrozmus: `onDisableAnimations` and sort properly (functions at the bottom)

client/src/modules/list/components/Items/model/actions.js
```diff
@@ -128,6 +129,36 @@ export const removeArchivedItems = payload => ({
   payload
 });
 
+export const disableAnimationForArchivedItems = () => ({
+  type: AnimationActionTypes.DISABLE_FOR_ARCHIVE_ITEMS
```
marekrozmus: `ARCHIVED`

PR #354, opened 2019-09-24
[EOC 443] Improvement refactor socket initialization and passing socket instance

adamklepacz:
The task is mainly related to require orders in index.js server file. As @Ciunkos said, reffering to the https://github.com/sandstreamdev/eoc/pull/289#discussion_r317961574 , import orderd should be not dependent on socketInstance initialization. 

Now it's refactored and it "works as nature always wanted 🐐 🌿 🌿".

### DIFF
--- a/server/controllers/cohort.js
+++ b/server/controllers/cohort.js
@@ -112,6 +112,7 @@ const updateCohortById = (req, resp) => {
     isArchived,
     name
   });
+  const socketInstance = io.getSocketInstance();
   let cohortActivity;
 
   if (name !== undefined && !validator.isLength(name, { min: 1, max: 32 })) {
@@ -146,10 +147,9 @@ const updateCohortById = (req, resp) => {
         data.description = description;
 
         return returnPayload(
-          socketActions.updateCohort(
-            io.getSocketInstance(),
-            allCohortsViewClients
-          )(data)
+          socketActions.updateCohort(socketInstance, allCohortsViewClients)(
+            data
+          )
         )(doc);
       }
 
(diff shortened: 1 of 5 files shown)


server/index.js
```diff
@@ -28,7 +20,13 @@ const mailerRouter = require('./routes/mailer');
 const activitiesRouter = require('./routes/activity');
 const unlockLocks = require('./common/utils/unlockLocks');
 
-/* eslint-enable import/order */
+const app = express();
+const server = http.Server(app);
+
+socket.init(server);
+const socketInstance = socket.getSocketInstance(server);
+socket.listen(socketInstance);
```
alemikolo: proposal: could we move line 27-28 to socket `index.js` file and initialize socket listeners in `init` function? , this index app will be cleaner?

server/controllers/cohort.js
```diff
@@ -293,7 +294,7 @@ const deleteCohortById = (req, resp) => {
       const { ownerIds: owners, name } = cohort;
       const data = { cohortId, owners };
 
-      socketActions.deleteCohort(socketInstance, allCohortsViewClients)(data);
+      socketActions.deleteCohort(io.getInstance(), allCohortsViewClients)(data);
```
Ciunkos: Please extract instance first to reduce diff and complexity.

```suggestion
      const socketInstance = io.getInstance();
      socketActions.deleteCohort(socketInstance, allCohortsViewClients)(data);
```

server/controllers/cohort.js
```diff
@@ -350,7 +351,7 @@ const removeMember = (req, resp) => {
     })
     .then(() => {
       socketActions.removeMember(
-        socketInstance,
+        io.getInstance(),
```
Ciunkos: Please apply the same pattern everywhere.

server/controllers/list.js
```diff
@@ -466,6 +466,7 @@ const updateListById = (req, resp) => {
     isDeleted,
     name
   });
+  const socketInstance = io.getInstance();
```
Ciunkos: Please move it closer to the usage.

PR #355, opened 2019-09-25
[EOC-490] Sort items lists and cohorts by date

adamklepacz:


server/controllers/list.js
```diff
@@ -130,7 +130,7 @@ const getListsMetaData = (req, resp) => {
     query.cohortId = sanitize(cohortId);
   }
 
-  List.find(query, '_id name created_at description items favIds type', {
+  List.find(query, '_id name createdAt description items favIds type', {
     sort: { created_at: -1 }
```
alemikolo: I think it is unneeded now because we always sort lists in selectors.

PR #356, opened 2019-09-25
[EOC-8] Notifications (email) - investigation when to send on which events

adamklepacz:
This is only a draft PR showing the possible flow of sending notifications to the user on the particular app events.
There is also a document related to this: 
https://drive.google.com/file/d/1DMtrHo5cpRcNmuiZ0F29-b5PVdVB5CL4/view

It's not gonna be merged

I have also sent email to Lukasz, to let him decide to which actions he would like to have emails.
List of actions you can find here: https://docs.google.com/document/d/1sH4axTOP8w_jCHLrHBD6L_i832nczdEQv_0Dvx3JA6M/edit?usp=sharing

### DIFF
--- a/server/controllers/list.js
+++ b/server/controllers/list.js
@@ -1029,7 +1029,7 @@ const removeMemberRole = (req, resp) => {
     });
 };
 
-const addViewer = (req, resp) => {
+const addViewer = (req, resp, next) => {
   const {
     user: { _id: currentUserId, idFromProvider }
   } = req;
@@ -1104,6 +1104,10 @@ const addViewer = (req, resp) => {
     .then(() => {
       if (user) {
         resp.send(userToSend);
+        // eslint-disable-next-line no-param-reassign
+        resp.locales = userToSend;
+
+        next();
 
         return fireAndForget(
           saveActivity(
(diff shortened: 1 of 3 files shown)


server/controllers/list.js
```diff
@@ -1104,6 +1104,10 @@ const addViewer = (req, resp) => {
     .then(() => {
       if (user) {
         resp.send(userToSend);
+        // eslint-disable-next-line no-param-reassign
+        resp.locales = userToSend;
```
alemikolo: (This is a proposal only I don't know which approach is better, we have to discuss this):
We can do this in the same way as in activities case. Instead of adding middleware to every route we can call methods eg. sendAddViewerEmailNotification(params) in the controller. We can call it in fireAndForgetHelper also.

adamklepacz: I think it is a matter of choice. 
I think using middleware is clean and elegant

Ciunkos: Personally, I would go for a single dispatch and handle notification as an effect of action - this way there is a single `dispatch(ActivityType.LIST_ADD_USER, payload)` call that broadcasts the event over sockets, handles DB writes and optionally sends notifications.
The best option but hardest to implement as you do not store raw events nor change authors is to watch the DB for changes, aggregate them and then diff them to get change lists to be sent as notifications.

adamklepacz:
Investigation marked as done

PR #357, opened 2019-09-26
[EOC-496] Change filter and sort options to search

adamklepacz:
It is now possible to serach using whatever field we want (name, authorName) etc.

chrome-capture (1)


client/src/modules/list/components/ItemsContainer.jsx
```diff
@@ -1,94 +1,48 @@
...
+import Filter from 'common/components/Filter';
+import { CloseIcon } from 'assets/images/icons';
 
-const SortOptionType = Object.freeze({
-  NAME: 'name',
-  DATE: 'createdAt',
-  AUTHOR: 'author',
-  VOTES: 'votes'
-});
+// FIXME: Remove sortbox and filterBox components and styles files
```
alemikolo: remove comment

client/src/modules/list/components/ItemsContainer.jsx
```diff
@@ -125,18 +78,15 @@ class ItemsContainer extends Component {
...
-            <SortBox
-              label={formatMessage({ id: 'list.items-container.sort-box' })}
-              onChange={this.onSortChange}
-              options={sortOptions}
-              sortBy={sortBy}
-              sortOrder={sortOrder}
+            <Filter
+              buttonContent={<CloseIcon />}
+              classes="items__filter"
+              fields={filterByFields}
```
alemikolo: hmm, `fields` prop is good because it is short but now I wonder if it is descriptive... Maybe sth like `fieldsToMatch`, `filterByFields`?

adamklepacz:
> I think `search input` is not enough different from the `add item input` or it is too close to it. Maybe a little bigger gaps between `search input`, `add input` and list title ('unhandled', 'done', 'archive') will resolve the matter.

Referring to this article
https://uxdesign.cc/best-ux-practices-for-search-inputs-c44dba565448
to distinguish search input from the basic input I have added magnifier icon on the left.
I have also added some more space under the search input. 

client/src/assets/images/icons/index.js
```diff
@@ -210,3 +210,12 @@ export const ArrowUpIcon = () => (
     />
   </Icon>
 );
+
+export const SearchIcon = () => (
+  <Icon data-icon="arrow-up" viewBox="0 0 512 512">
```
marekrozmus: data-icon is incorrect. What is this prop for anyway?

client/src/modules/list/components/ItemsContainer.scss
```diff
@@ -146,4 +143,69 @@
     font-size: 12px;
     line-height: 1.2;
   }
+
+  &__filter {
```
marekrozmus: why this style is here and not in `Filter.scss`?

adamklepacz: I've been thinking about adding this style to a separated file but @alemikolo said that we may want to have different styles for this component in different places.

@alemikolo  can you refer to that?

client/src/assets/images/icons/index.js
adamklepacz: it is here just for accessibility reasons. 
Change it to`dataIcon` or remove it anyway?

client/src/modules/list/components/ItemsContainer.scss
alemikolo: I thought that this component the component will only have the functionality implemented. Because we may want to style it in a different way in different places.

We use it to filter lists in `move the item to list`  panel, now its added to `itemsList` to filter items.  Łukasz said that he would like to have such kind of filter in activities panel.

In our desktop view in all those places, the filter should have at least slightly different styles...

Maybe it would be better to put filter styles in a separate file, and overwrite them only when it is necessary?

marekrozmus: Move to component css file all default styles and override in other places elements that needs that.

client/src/assets/images/icons/index.js
marekrozmus: "arrow-up" should be "search"

client/src/assets/images/icons/index.js
```diff
@@ -210,3 +210,12 @@ export const ArrowUpIcon = () => (
     />
   </Icon>
 );
+
+export const SearchIcon = () => (
+  <Icon data-icon="search-icon" viewBox="0 0 512 512">
```
Ciunkos: https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA24

client/src/common/components/Filter.jsx
```diff
@@ -1,7 +1,10 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import _debounce from 'lodash/debounce';
-import _trim from 'lodash/trim';
+import _trimStart from 'lodash/trimStart';
```
Ciunkos: Please trim only before performing the search and always leave the input as it is entered by the user (make sure filter "list 2   " ["list 1", "list 2"] yields ["list 2"]).

client/src/common/components/Filter.jsx
```diff
@@ -22,14 +25,29 @@ class Filter extends PureComponent {
     this.debouncedFilter.cancel();
   }
 
-  filter = query => {
-    const { options, onFilter } = this.props;
+  filterByFields = query => {
+    const { filterBy, options } = this.props;
+    const match = new RegExp(query, 'i');
```
Ciunkos: Please use case-invariant (toLowerCase()/toLocaleLowerCase) `String.includes` search as regexp constructor requires proper escaping (try "*" as a search query) and can lead to REDOS.

client/src/common/components/Filter.jsx
```diff
@@ -22,14 +25,29 @@ class Filter extends PureComponent {
...
+
+    return options.filter(option => {
+      let testString = '';
+
+      filterBy.forEach(field => {
```
Ciunkos: Wouldn't fields name be better? Consider `filterBy` -> `fields` rename.

client/src/modules/list/components/ItemsContainer.jsx
```diff
@@ -1,94 +1,46 @@
...
-    const {
-      currentUser: { id }
-    } = this.props;
+  updateItems = () => {
+    const { items } = this.props;
 
-    return filterBy === FilterOptionType.MY_ITEMS
-      ? items.filter(item => item.authorId === id)
-      : items;
+    this.setState({ items });
```
Ciunkos: Why not pure render filtering? Are there any performance issues present?

client/src/assets/images/icons/index.js
adamklepacz: Added as separated task

client/src/modules/list/components/ItemsContainer.jsx
adamklepacz: It was a customer's wish to replace the filter and sort components with search input

client/src/common/components/Filter.jsx
```diff
@@ -45,28 +64,17 @@ class Filter extends PureComponent {
...
-    const query = _trim(value);
-
-    this.setState(() => ({ query }), this.handleFilterOptions);
-  };
-
-  handleFilterOptions = () => {
-    const { query } = this.state;
 
-    this.debouncedFilter(query);
+    this.setState(() => ({ query: value }), () => this.debouncedFilter(value));
```
marekrozmus: Do you need function if not using previous state?
Isn't this enough? `this.setState({ query: value }, () => this.debouncedFilter(value));`

client/src/modules/list/components/ItemsContainer.jsx
Ciunkos: This is rather a technical question as this should not affect the UI. I think you should strive to minimize state and if possible, use pure filtering instead.

client/src/common/components/Filter.scss
```diff
@@ -0,0 +1,67 @@
...
+    background: 0;
+    border: 0;
+    margin: 0;
+    padding: 0;
+    min-width: 40px;
+    height: 40px;
+    cursor: pointer;
+
+    svg {
+      height: 14px;
```
Ciunkos: Please use constant-size helper as svgs tends to collapse within flex layouts.

client/src/common/components/Filter.scss
```diff
@@ -0,0 +1,67 @@
...
+
+      path {
+        fill: $text-normal;
+      }
+    }
+
+    &:hover {
+      svg {
+        height: 20px;
```
Ciunkos: Please use `transform: scale` to use GPU-accelerated transitions.

client/src/modules/list/components/Items/MoveToListPanel/MoveToListPanel.jsx
```diff
@@ -118,6 +118,7 @@ class MoveToListPanel extends PureComponent {
       intl: { formatMessage },
       lists
     } = this.props;
+    const filterByFields = ['name'];
```
Ciunkos: This can be moved outside the class context.

client/src/modules/list/components/ItemsContainer.jsx
```diff
@@ -112,11 +64,10 @@ class ItemsContainer extends Component {
...
       intl: { formatMessage },
       isMember,
-      items
+      items: itemsToSearch
     } = this.props;
-    const { filterBy, sortBy, sortOrder } = this.state;
-    const filteredList = this.filterItems(items, filterBy);
-    const sortedList = this.sortItems(filteredList, sortBy, sortOrder);
+    const { items } = this.state;
+    const filterByFields = ['authorName', 'name'];
```
Ciunkos: This can be moved outside the class context.

client/src/modules/list/components/ItemsContainer.jsx
adamklepacz: Sorry, but I don't understand what you mean by pure filtering?

client/src/common/components/Filter.jsx
```diff
@@ -22,14 +25,30 @@ class Filter extends PureComponent {
...
+        if (option[field]) {
+          testString += option[field];
+        }
+      });
+
+      return testString.toLowerCase().includes(normalizedQuery);
+    });
+  };
+
+  handleFilter = query => {
```
marekrozmus: `handleFilterChange`? or what does this method do exactly

adamklepacz: name changed

PR #358, opened 2019-09-26
[EOC-467] Socket emit list data to proper rooms

alemikolo:


server/common/utils/helpers.js
```diff
@@ -401,6 +401,38 @@ const fireAndForget = promise => promise.catch(err => console.error(err));
...
+  return {
+    ...responseWithList(list, userId),
+    cohortId: cohort ? cohort.cohortId : null,
+    cohortName: cohort ? cohort.name : null,
+    isArchived,
+    isGuest: !cohort || (cohort && !isMember(cohort, userId)),
+    isMember: isMember(list, userId),
+    isOwner: isOwner(list, userId),
+    items: _keyBy(items, '_id'),
+    members: _keyBy(members, '_id')
```
adamklepacz: just curious why keyby here? 
we were always using keyby on front end before dispatching data to the redux

alemikolo: We did it in action on fetch result. So I should implement this again in `listEventsController` helpers function. I didn't want to make this method any more complicated. So I send ready list object from backend

PR #366, opened 2019-10-01
[EOC-499] Add description to show/more less buttons

adamklepacz:
Zrzut ekranu 2019-10-1 o 13 30 51
Zrzut ekranu 2019-10-1 o 13 30 44


client/src/locales/en.json
```diff
@@ -61,6 +61,8 @@
   "common.app-name": "End of coffee",
   "common.try-again": "Please try again.",
   "common.something-went-wrong": "Something went wrong.",
+  "common.show-more": " Show more",
```
marekrozmus: why starts with space?

adamklepacz: it's a typo

PR #367, opened 2019-10-02
[EOC-511] Refactor save activity to pass object instead off list parameters

alemikolo:


server/controllers/comment.js
```diff
@@ -51,16 +51,17 @@ const addComment = (req, resp) => {
         displayName
       );
       const data = { itemId, listId, comment: commentToSend };
+      const activity = {
+        activityType: ActivityType.ITEM_ADD_COMMENT,
+        performerId: userId,
+        itemId: sanitizedItemId,
+        listId: sanitizedListId,
+        cohortId: list.cohortId,
+        editedUserId: null,
```
Ciunkos: Consider using the default values instead of explicit nulls.

server/controllers/activity.js
```diff
@@ -5,24 +5,8 @@ const List = require('../models/list.model');
...
-    activityType,
-    performerId,
-    cohortId,
-    editedUserId,
-    editedValue,
-    itemId,
-    listId
-  });
+const saveActivity = activity => {
+  const newActivity = new Activity({ ...activity });
```
marekrozmus: check if new object creation is needed - wht not passing only `activity`?

server/controllers/cohort.js
```diff
@@ -48,7 +48,13 @@ const createCohort = (req, resp) => {
...
         .send(responseWithCohort(doc, userId));
 
-      saveActivity(ActivityType.COHORT_ADD, userId, null, null, doc._id);
+      const activity = {
+        activityType: ActivityType.COHORT_ADD,
+        cohortId: doc._id,
+        performerId: userId
+      };
+
+      saveActivity(activity);
```
adamklepacz: without fireAndForget ?

PR #368, opened 2019-10-02
[EOC-514] 'X' sign should be visible in filter field only if user entered something

alemikolo:


client/src/common/components/Filter.jsx
```diff
@@ -71,6 +71,7 @@ class Filter extends PureComponent {
   render() {
     const { query } = this.state;
     const { buttonContent, clearFilterButton, placeholder } = this.props;
+    const isClearButtonVisible = clearFilterButton && buttonContent && query;
```
Ciunkos: Test case - `query = " "` 😉

alemikolo: When the user enters "  " the 'X' will appear. But as we discussed this is good solution because maybe the user accidentally entered that white-space and with the 'X' button visible the user can easily clear the input.

PR #369, opened 2019-10-02
[EOC-512] When entering sack search field for done is focused, should not be

adamklepacz:


client/src/common/components/Filter.jsx
```diff
@@ -98,6 +102,7 @@ Filter.defaultProps = {
 };
 
 Filter.propTypes = {
+  autofocus: PropTypes.bool,
```
alemikolo: proposal: add PropTypes default props and set autofocus to false. So you won't have to pass autofocus prop in every place. Only on move to item you pass:
```
 <Filter
    autofocus
```

adamklepacz: There is no need for default props if this is not required. 
I can just not pass this prop and it will be treated as falsy value :) 

I have removed autofocus from ItemsContainer.jsx file - check this out

PR #370, opened 2019-10-02
[EOC-516] Disable getting lists from archived cohorts by getAvailableLists controller

alemikolo:


server/controllers/list.js
```diff
@@ -1794,11 +1794,22 @@ const getAvailableLists = (req, resp) => {
...
+      .exec();
+
+    const listToSend = lists
+      .filter(
+        list => !list.cohortId || (list.cohortId && !list.cohortId.isArchived)
+      )
+      .map(({ _id, name }) => ({ _id, name }));
+
+    resp.send(listToSend);
+  } catch {
```
Ciunkos: Please at least log the error to ease the debugging.

server/controllers/list.js
```diff
@@ -1794,11 +1794,22 @@ const getAvailableLists = (req, resp) => {
...
-    .catch(() => resp.sendStatus(400));
+  try {
+    const lists = await List.find(query, 'name')
+      .lean()
+      .populate('cohortId', 'isArchived')
+      .exec();
+
+    const listToSend = lists
+      .filter(
+        list => !list.cohortId || (list.cohortId && !list.cohortId.isArchived)
```
Ciunkos: `list.cohortId.isArchived`? Shouldn't you populate the `cohort` field instead (based on the `cohortId` field)?

server/controllers/list.js
alemikolo: We don't console log errors. But I added a task: https://jira2.sanddev.com/browse/EOC-525,
to discuss this and create some logging system or at least console.log errors.

server/controllers/list.js
alemikolo: we have cohortId field in list schema. when we populate cohort data it is stored in cohortId.

PR #374, opened 2019-10-03
[EOC-509] Remove data-icon form icons components and add aria label

alemikolo:
I added aria-label and title props to the Icons components. But in almost all cases passing those props to the icon is unnecessary because:
1. We use icons, especially in buttons or links that already have their titles attributes which are used by screen readers. 
2. We use icons only for visual purposes next to titles, headers or labels so they have no functional significance.

I passed labels only in places where icon displays additional info which isn't displayed in another way eg. `lock list icon`.

Some buttons had no title attribute so I added it.


client/src/assets/images/icons/index.js
```diff
@@ -52,7 +52,7 @@ export const PlusIcon = () => (
 );
 
 export const RegularStarIcon = () => (
-  <Icon data-icon="star" viewBox="0 0 576 512">
+  <Icon aria-label="fot favourite" viewBox="0 0 576 512">
```
adamklepacz: fot favourite?

client/src/assets/images/icons/index.js
```diff
@@ -97,7 +97,7 @@ export const UserIcon = () => (
 );
 
 export const DotsIcon = () => (
-  <Icon data-icon="ellipsis-h" viewBox="0 0 512 512">
+  <Icon aria-label="ellipsis" rviewBox="0 0 512 512">
```
adamklepacz: `rviewBox`

typo

client/src/assets/images/icons/index.js
```diff
@@ -124,7 +124,7 @@ export const LogoutIcon = () => (
 );
 
 export const ChevronRight = () => (
-  <Icon data-icon="chevron-right" viewBox="0 0 320 512">
+  <Icon viewBox="0 0 320 512">
```
adamklepacz: no aria-label?

client/src/assets/images/icons/index.js
```diff
@@ -178,23 +178,21 @@ export const ListViewIcon = () => (
...
       fill="#000"
     />
-  </svg>
+  </Icon>
 );
 
 export const CheckIcon = () => (
-  <Icon data-icon="check" viewBox="0 0 512 512">
+  <Icon aria-label="ok" viewBox="0 0 512 512">
```
adamklepacz: maybe `check mark` would be more reasonable for people using aria labels?

client/src/assets/images/icons/index.js
```diff
@@ -212,25 +210,16 @@ export const ErrorIcon = () => (
 );
 
 export const AppIcon = () => (
-  <Icon data-icon="coffee" viewBox="0 0 640 512">
+  <Icon aria-label="eoc" viewBox="0 0 640 512">
```
adamklepacz: maybe 'aplication logo' will be more reasonable for people using aria labels?

client/src/assets/images/icons/index.js
```diff
@@ -25,7 +25,7 @@ export const CohortIcon = () => (
 );
 
 export const ListIcon = () => (
-  <Icon data-icon="clipboard-list" viewBox="0 0 384 512">
+  <Icon aria-label="list" viewBox="0 0 384 512">
```
Ciunkos: Please investigate https://stackoverflow.com/questions/22039910/what-is-aria-label-and-how-should-i-use-it

client/src/app/components/Layout/Layout.jsx
```diff
@@ -162,8 +162,8 @@ export class Layout extends PureComponent {
               title={formatMessage({
                 id:
                   viewType === ViewType.LIST
-                    ? 'app.layout.list-view'
-                    : 'app.layout.tile-view'
+                    ? 'app.layout.tile-view'
```
przemyslawzalewski: ??

alemikolo: There was a bug. Message-ids were reversed. Now it's ok. If there is a list view we want to change it to a tile view. So button title should be: 'change to tile view'

Ciunkos: Consider renaming locale key then, i.e. "app.layout.switch-to-tile-view".

PR #375, opened 2019-10-03
[EOC-520] Remove user account refactor controllers to check if display name exist

adamklepacz:
After a discussion with @Ciunkos  we decided to recognize if the user name is undefined, null or whatever only on front end and react to that by displaying the "Anonymous" string. 

The string has to be translated, that's why function formatName() takes formatMessage function from reac-intl library, as a second parameter.

### DIFF
--- a/client/src/common/components/Avatar.jsx
+++ b/client/src/common/components/Avatar.jsx
@@ -1,9 +1,12 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import { injectIntl } from 'react-intl';
 
 import { UserIcon } from 'assets/images/icons';
 import './Avatar.scss';
+import { formatName } from 'common/utils/helpers';
+import { IntlPropType } from 'common/constants/propTypes';
 
 class Avatar extends PureComponent {
   state = {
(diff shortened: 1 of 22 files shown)


client/src/common/components/Avatar.jsx
```diff
@@ -1,9 +1,12 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import { injectIntl } from 'react-intl';
```
alemikolo: hmm `Avatar` is a simple component to only renders an image. It gets a user name from props passed by its parent. Could we pass already `formattedName` to  `Avatar` and not inject additional libs/logic to it?

client/src/common/components/Members/index.jsx
```diff
@@ -128,24 +131,31 @@ class MembersBox extends PureComponent {
   };
 
   renderMemberList = () => {
-    const { members } = this.props;
+    const {
+      intl: { formatMessage },
+      members
```
alemikolo: will users who have deleted their accounts be members of the list or cohort? I think not. So usage of format message is not necessary for all members' components. We have to protect those places where the user name has to be populated: item, activities. On profile view (User who delete thier account won't display thier profile :-), on list/cohort members we won't display users who deleted their accounts. I somebody will delete thier account they will be removed from lists or cohorts.

PR #376, opened 2019-10-03
[EOC-520] Remove user account refactor controllers to check if displayName exist v2

adamklepacz:
After a discussion with @Ciunkos  we decided to recognize if the user name is undefined, null or whatever only on front end and react to that by displaying the "Anonymous" string.

The string has to be translated, that's why function formatName() takes formatMessage function from react-intl library, as a second parameter.

### DIFF
--- a/client/src/modules/list/components/Items/ListArchivedItem/ListArchivedItem.jsx
+++ b/client/src/modules/list/components/Items/ListArchivedItem/ListArchivedItem.jsx
@@ -33,15 +33,14 @@ class ListArchivedItem extends PureComponent {
     const {
       currentUser: { name: userName },
       data: { _id: itemId, done, name },
-      intl: { formatMessage },
+
       match: {
         params: { id: listId }
       },
       restoreItem
     } = this.props;
-    const formattedName = formatName(userName, formatMessage);
 
-    return restoreItem(listId, itemId, name, formattedName, done);
+    return restoreItem(listId, itemId, name, userName, done);
   };
 
   showConfirmation = () => this.setState({ isConfirmationVisible: true });
--- a/client/src/modules/list/components/Items/ListArchivedItem/ListArchivedItem.jsx
+++ b/client/src/modules/list/components/Items/ListArchivedItem/ListArchivedItem.jsx
@@ -18,6 +18,7 @@ import {
 import Confirmation from 'common/components/Confirmation';
 import { getCurrentUser } from 'modules/user/model/selectors';
 import './ListArchivedItem.scss';
+import { formatName } from 'common/utils/helpers';
 
 class ListArchivedItem extends PureComponent {
   constructor(props) {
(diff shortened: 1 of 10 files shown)


client/src/modules/list/components/Items/ListArchivedItem/ListArchivedItem.jsx
```diff
@@ -32,13 +33,15 @@ class ListArchivedItem extends PureComponent {
     const {
       currentUser: { name: userName },
       data: { _id: itemId, done, name },
+      intl: { formatMessage },
       match: {
         params: { id: listId }
       },
       restoreItem
     } = this.props;
+    const formattedName = formatName(userName, formatMessage);
```
alemikolo: do we have to call `formatName` on the current User name? current user can not be anonymous...

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -59,10 +65,10 @@ class ListItem extends PureComponent {
     if (!isMember) {
       return;
     }
-
+    const formattedName = formatName(editedBy, formatMessage);
```
alemikolo: `editedBy` is current user name - do we have to call `formatName` on the current User name? current user can not be anonymous...

client/src/modules/list/components/Items/ListItem/ListItem.jsx
```diff
@@ -133,14 +139,16 @@ class ListItem extends PureComponent {
       archiveItem,
       currentUser: { name: editedBy },
       data: { _id: itemId, name },
+      intl: { formatMessage },
       match: {
         params: { id: listId }
       }
     } = this.props;
+    const formattedName = formatName(editedBy, formatMessage);
```
alemikolo: the same here - current user can not be anonymous...

client/src/modules/list/components/Items/ListItemDescription/ListItemDescription.jsx
```diff
@@ -123,13 +127,15 @@ class ListItemDescription extends PureComponent {
...
       disabled,
+      intl: { formatMessage },
       itemId,
       match: {
         params: { id: listId }
       },
       name,
       updateListItem
     } = this.props;
+    const formattedName = formatName(userName, formatMessage);
```
alemikolo: do we have to call `formatName` on the current User name? current user can not be anonymous...

client/src/modules/list/components/Items/ListItemName/ListItemName.jsx
```diff
@@ -67,11 +74,12 @@ class ListItemName extends PureComponent {
     } = this.props;
     const isNameUpdated = updatedName !== name;
     const canBeUpdated = updatedName.trim().length > 1;
+    const formattedName = formatName(userName, formatMessage);
```
alemikolo: do we have to call `formatName` on the current User name? current user can not be anonymous...

PR #382, opened 2019-10-07
[EOC-495] Remove user account ui flow and re-authentication

adamklepacz:
Remove flow goes as images shows.
The user clicks on the DELETE ACCOUNT button in the Profile view.
A dialog is displayed, where he has to provide email, password and verification string. 
If the verification string is wrong, we display errors. If the data is correct, we process the request to the API. 

Right now, API is faking response with SUCCESS status, to display the success DIALOG  -> Account was deleted. 
Users can't lose the focus, disable this dialog.  One possible action is to CLICK on the OK button. 
After the user clicks the OK button, he is redirected to the SIGN IN page.

1. Click delete
Zrzut ekranu 2019-10-7 o 15 46 25
2.Dialog is displayed
Zrzut ekranu 2019-10-7 o 15 46 36
3.Wrong data causes errors
Zrzut ekranu 2019-10-7 o 15 46 49
4.Correct data. Account is deleted. User gets a message
Zrzut ekranu 2019-10-7 o 15 47 02
5. Redirection to SIGN in page
End_of_dsadasdaoffee_-_track_your_office_inventory


client/src/locales/en.json
```diff
@@ -49,6 +49,7 @@
   "common.member-details.leave": "Leave {data}",
   "common.members-form.email": "email",
   "common.members-form.add": "add",
+  "common.ok": "OK",
```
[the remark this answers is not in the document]
adamklepacz: but what if we will use within SPAN? 😄

client/src/modules/user/UserProfile/DeleteAccount.jsx
```diff
@@ -0,0 +1,169 @@
...
+        if (result) {
+          this.setState({
+            isAccountDeletedDialogVisible: true,
+            isDeleteDialogVisible: false,
+            pending: false
+          });
+        }
+      }
+
+      throw new Error();
```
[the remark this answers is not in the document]
adamklepacz: it throws error only when 
``` if (verificationText === verificationString) { ``` is false 
otherwise it goes inside the if and it never execute the error because deleteDialog is hidden after result is SUCCESS. 
but to be sure nothing unexpected happens I've added return within this if statement.

File: client/src/locales/en.json [hunk 3]
```diff
@@ -85,8 +85,8 @@
   "about.intro-heading": "{appName} Office inventory tracking app",
   "about.intro-subheading": "{appName} helps you track all the office inventory you need!",
   "user.anonymous": "Anonymous",
-  "user.account-deleted": "Your account has been successfully deleted.",
-  "user.account-deleted-message": "Your account has been successfully deleted. Click OK to continue.",
+  "user.account-deleted": "Your account has been successfully deleted. ",
```
alemikolo: White space at the end of the message. This message will be joined with another message so space is needed there will always be this space even if needed.
I'm not sure if this is correct but maybe try add space in component:
```
<FormatMessage />
{' '}
<FormatMessage />
```

client/src/locales/en.json
```diff
@@ -81,6 +85,12 @@
   "about.intro-heading": "{appName} Office inventory tracking app",
   "about.intro-subheading": "{appName} helps you track all the office inventory you need!",
   "user.anonymous": "Anonymous",
+  "user.account-deleted": "Your account has been successfully deleted. ",
```
marekrozmus: extra space at the end of line

client/src/locales/en.json
```diff
@@ -81,6 +85,12 @@
...
+  "user.account-deleted-message": "Click OK to continue.",
```
marekrozmus: `OK` should be taken from "common.ok"

client/src/common/components/AlertBox.scss
```diff
@@ -0,0 +1,59 @@
...
+    padding: four-by(1);
+    border-radius: 100%;
+    height: 10px;
+    width: 10px;
+    display: inline-block;
+    margin-right: four-by(2);
+  }
+
+  &--warning {
+    background: rgba($red, 0.1);
```
marekrozmus: these colors variables needs to be refactored and have meaningful names

client/src/modules/user/UserProfile/DeleteAccount.jsx
```diff
@@ -0,0 +1,172 @@
...
+            onPasswordChange={this.handlePasswordChange}
+            onSubmit={this.handleDeleteAccount}
+            onVerificationTextChange={this.handleVerificationText}
+            pending={pending}
+          />
+        )}
+        {isAccountDeletedDialogVisible && (
+          <Dialog
+            cancelLabel="common.ok"
+            onCancel={this.handleCancel}
```
marekrozmus: why `cancel` is used?

client/src/modules/user/UserProfile/DeleteForm.jsx
```diff
@@ -0,0 +1,67 @@
...
+      />
+    </label>
+    <input className="hidden" type="submit" />
+  </form>
+);
+
+DeleteForm.propTypes = {
+  error: PropTypes.bool,
+
+  intl: IntlPropType.isRequired,
```
marekrozmus: new line after this line, not before

client/src/modules/user/UserProfile/DeleteAccount.jsx
```diff
@@ -0,0 +1,172 @@
...
+  handleDeleteAccount = async event => {
+    this.setState({ pending: true });
+    event.preventDefault();
+
+    const {
+      intl: { formatMessage }
+    } = this.props;
+    const { verificationText } = this.state;
+    const verificationString = formatMessage({ id: 'delete-form.verify-text' });
+
```
marekrozmus:
```
if (verificationText !== verificationString) {
   this.setState({ isErrorVisible: true, pending: false });
   return;
}
```

client/src/locales/en.json
```diff
@@ -81,6 +85,12 @@
...
+  "user.delete-account-error": "Wrong data. Please try again.",
```
marekrozmus: Sth like: `Entered data is incorrect.`

client/src/modules/user/UserProfile/DeleteAccount.jsx
adamklepacz: This has to be refactored on a separated task

PR #383, opened 2019-10-08
[EOC-529] Sort all strings in en.json file

alemikolo:


Ciunkos:
What about a tool that sorts this kind of files on a pre-commit/pre-push checks? I can guarantee you that the sorting will be lost in a few commits so a tool to enforce this would be welcome :)

adamklepacz:
> What about a tool that sorts this kind of files on a pre-commit/pre-push checks? I can guarantee you that the sorting will be lost in a few commits so a tool to enforce this would be welcome :)

yes we can add this in the future, but in the next two weeks, we need to FOCUS to close all existing functionality and code. We can not implement anything extra without the Lukasz agreement.

@alemikolo you can add new IMPROVEMENT task on the bottom of the backlog

alemikolo:
But I won't add a new task until we end features that are the most important for Łukasz. Of course, I can propose such a task on preplanning in the next week.

PR #384, opened 2019-10-08
[EOC-527] Improvement refactor confirmlabel in dialog component to get formatted string and not string

adamklepacz:


client/src/common/components/Dialog.jsx
```diff
@@ -67,8 +66,8 @@ const Dialog = ({
 );
 
 Dialog.defaultProps = {
-  cancelLabel: 'common.button.cancel',
-  confirmLabel: 'common.button.confirm',
+  cancelLabel: 'cancel',
```
alemikolo: if someone does not pass those labels he will get English ones - they are hardcoded now...

PR #385, opened 2019-10-08
[Eoc-528] Validate item length name

alemikolo:


client/src/modules/list/components/Items/InputBar/InputBar.jsx
```diff
@@ -67,57 +65,43 @@ class InputBar extends Component {
...
     const newItem = {
       authorId: currentUser.id,
       name: itemName
     };
 
-    if (_trim(itemName)) {
+    if (!errorMessageId) {
       this.setState({ pending: true });
 
       return addItem(newItem, id).finally(() => {
```
adamklepacz: consider refactoring this to try catch block

alemikolo: Changing to `try-catch`  block is a good idea but I think we should avoid any refactoring unless it is necessary for the given situation

PR #386, opened 2019-10-10
[EOC-531] Profile page add email notifications frequency ( front and backend)

adamklepacz:
Zrzut ekranu 2019-10-10 o 12 50 22
Zrzut ekranu 2019-10-10 o 12 50 11


client/src/modules/user/UserProfile/EmailNotifications.jsx
```diff
@@ -0,0 +1,89 @@
...
+  render() {
+    const {
+      intl: { formatMessage }
+    } = this.props;
+    const { error, never, weekly } = this.state;
+
+    return (
+      <section className="email-notifications">
+        {error && (
+          <AlertBox type={MessageType.ERROR}>
```
alemikolo: Until now we displayed such errors in our notifications (excluding forms). I don't say that your solution is wrong, maybe it's even better but now it's patchy also

client/src/modules/user/UserProfile/EmailNotifications.jsx
```diff
@@ -0,0 +1,89 @@
...
+            <span>&nbsp;</span>
+            <FormattedMessage id="common.try-again" />
+          </AlertBox>
+        )}
+        <h2 className="email-notifications__heading">
+          <FormattedMessage id="email.notification.heading" />
+        </h2>
+        <ul className="email-notifications__list">
+          <li className="email-notifications__list-item">
+            <SwitchButton
```
alemikolo: Because  these options are separable I think it will be better to use `<radio />` or If we would have more options `<selecy />` tag

client/src/modules/user/model/actions.js
```diff
@@ -149,3 +149,6 @@ export const getAccountDetails = token =>
 
 export const deleteAccount = (email, password) =>
   deleteData('/auth', { email, password });
+
+export const saveEmailNotificationSettings = settings =>
+  postData('/auth/email-notification-settings', settings);
```
alemikolo: We always displayed notifications in actions (excluding forms, where we wanted to display errors inline)?

server/controllers/userAuth.js
```diff
@@ -448,6 +448,27 @@ const deleteAccount = async (req, resp) => {
...
 
+const updateEmailNotificationSettings = async (req, resp) => {
+  const { never, weekly } = req.body;
+  const { _id } = req.user;
+
+  try {
+    await User.findOneAndUpdate(
+      { _id },
+      {
+        emailNotificationSettings: {
```
alemikolo: I wonder if this should be called more general settings and inside we should save more specific "data"

server/controllers/userAuth.js
```diff
@@ -448,6 +448,27 @@ const deleteAccount = async (req, resp) => {
...
+          never,
+          weekly
```
alemikolo: Maybe, like in list type case, we should use some enums and store:
```
settings: {
  notificationFrequency: 'weekly'
}
```
If switches would be changed to radio/select we will have one value then, which could be a string instead of  boolean

server/migrations/20191010101148-0.0.9.js
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable no-param-reassign */
+const { runAsyncTasks } = require('../common/utils');
+
+const up = async db => {
+  const addEmailNotificationSettings = async () =>
+    db
+      .collection('users')
+      .updateMany(
+        { emailNotificationSettings: { $exists: false } },
```
alemikolo: if my proposals would be implemented, remember to change the migrations

server/models/user.model.js
```diff
@@ -18,7 +18,11 @@ const UserSchema = new Schema(
     resetTokenExpirationDate: { type: Date },
     signUpHash: { type: String },
     signUpHashExpirationDate: { type: Date },
-    surname: { type: String }
+    surname: { type: String },
+    emailNotificationSettings: {
```
alemikolo:
```
settings: {
 notificationFrequency : { type: String, default: ematilFrequencyType.NEVER }
}
```

client/src/modules/user/UserProfile/EmailNotifications.jsx
```diff
@@ -0,0 +1,122 @@
...
+    this.setState({ error: false });
+
+    try {
+      const result = await saveEmailNotificationSettings(notificationFrequency);
+
+      if (result) {
+        this.showSuccessMessage();
+      }
+    } catch {
+      this.setState({ error: true });
```
alemikolo: we display such errors in our notifications (excluding forms where we want to display errors inline). I don't say that your solution is wrong, maybe it's even better but now it's patchy also

client/src/modules/user/UserProfile/SelectDays.jsx
```diff
@@ -0,0 +1,43 @@
...
+import { injectIntl } from 'react-intl';
+
+import { NotificationFrequency } from 'common/constants/enums';
+import { IntlPropType } from 'common/constants/propTypes';
+import './SelectDays.scss';
+
+const SelectDays = ({ intl: { formatMessage }, onChange }) => (
+  <div className="select-days">
+    <select className="select-days__select" onChange={onChange}>
+      <option value={NotificationFrequency.MONDAY}>
```
alemikolo: Could we use some loop (map etc) to render select options automatically?

client/src/modules/user/UserProfile/EmailNotifications.jsx
```diff
@@ -0,0 +1,122 @@
+import React, { PureComponent } from 'react';
+import { FormattedMessage } from 'react-intl';
+
+import AlertBox from 'common/components/AlertBox';
+import { saveEmailNotificationSettings } from '../model/actions';
+import { NotificationFrequency, MessageType } from 'common/constants/enums';
+import SelectDays from './SelectDays';
+import './EmailNotification.scss';
+
+// TODO: Fetch settings when profile view is displayed
```
alemikolo: Hmm I think in this task, all features connected to notification settings should be done... (front, backend, getting and saving)  The method saving notification settings in DB is created, why fetching setting is missing? I think when we get user details we should send only one prop more?

server/models/user.model.js
```diff
@@ -18,7 +18,9 @@ const UserSchema = new Schema(
     resetTokenExpirationDate: { type: Date },
     signUpHash: { type: String },
     signUpHashExpirationDate: { type: Date },
-    surname: { type: String }
+    surname: { type: String },
+    emailNotificationsFrequency: { type: String, default: '' },
```
alemikolo: alphabetic order

client/src/common/constants/enums.js
```diff
@@ -65,3 +65,15 @@ export const PasswordValidationValues = Object.freeze({
...
+export const NotificationFrequency = Object.freeze({
+  FRIDAY: 'notification/FRIDAY',
+  MONDAY: 'notification/MONDAY',
+  NEVER: 'notification/NEVER',
+  SATURDAY: 'notification/SATURDAY',
+  SUNDAY: 'notification/SUNDAY',
+  THURSDAY: 'notification/THURSDAY',
+  TUESDAY: 'notification/TUESDAY',
+  WEDNESDAY: 'notification/WEDNESDAY',
+  WEEKLY: 'notifications/WEEKLY'
```
alemikolo: why weekly is here? Is it needed?

client/src/modules/user/model/actions.js
```diff
@@ -149,3 +149,6 @@ export const getAccountDetails = token =>
 
 export const deleteAccount = (email, password) =>
   deleteData('/auth', { email, password });
+
+export const saveEmailNotificationSettings = notificationFrequency =>
+  postData('/auth/email-notification-settings', { notificationFrequency });
```
alemikolo: We always displayed notification here (except of auth forms). Why there are placed now in profile view? This is not consistently

alemikolo:
Can you upload new screenshots and remove the outdated ones?

client/src/modules/user/UserProfile/EmailNotifications.jsx
```diff
@@ -0,0 +1,122 @@
...
+
+class EmailNotifications extends PureComponent {
+  state = {
+    success: false,
+    error: false,
+    notificationFrequency: NotificationFrequency.MONDAY,
+    areDaysVisible: true
```
Ciunkos: No need for that state. Please just vary on `notificationFrequency !== NotificationFrequency.NEVER`

server/controllers/userAuth.js
```diff
@@ -448,6 +448,24 @@ const deleteAccount = async (req, resp) => {
...
 
+const updateEmailNotificationSettings = async (req, resp) => {
+  const { _id } = req.user;
+  const { notificationFrequency } = req.body;
+
+  try {
+    await User.findOneAndUpdate(
+      { _id },
+      {
+        emailNotificationsFrequency: notificationFrequency
```
Ciunkos: Please be consistent - please use `emailNotificationFrequency` or `notificationFrequency` but pick one throught server/frontent.

server/controllers/userAuth.js
```diff
@@ -448,6 +448,24 @@ const deleteAccount = async (req, resp) => {
...
+      }
+    ).exec();
+
+    resp.send();
+  } catch {
+    resp.send(400);
```
Ciunkos: Please log the error to ease the debugging.

server/migrations/20191010101148-0.0.9.js
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable no-param-reassign */
+const { runAsyncTasks } = require('../common/utils');
+
+const up = async db => {
+  const addEmailNotificationSettings = async () =>
+    db.collection('users').updateMany(
+      {
+        emailNotificationsFrequency: { $exists: false },
+        emailReportSentAt: { $exists: false }
```
Ciunkos: Is this needed?

server/migrations/20191010101148-0.0.9.js
```diff
@@ -0,0 +1,38 @@
...
+      },
+      {
+        $set: {
+          emailNotificationsFrequency: '',
```
Ciunkos: This is not a valid value. Valid values are one of the enumerable - please assign the default here.

server/models/user.model.js
Ciunkos: Please assign proper default here - `NEVER`.

server/models/user.model.js
```diff
@@ -18,7 +18,9 @@ const UserSchema = new Schema(
...
+    emailReportSentAt: { type: Date, default: '' }
```
Ciunkos: Please consider renaming to `lastEmailNotificationSentAt`. Please do not assign empty string as the default - the type is `Date`. As this is not required, not need for default.

client/src/common/constants/enums.js
adamklepacz: yes it is needed to recognize that someone chose weekly and DAYS select needs to be visible.

It is not saved in the DB. it is only for select days visibility

client/src/common/constants/enums.js
```diff
@@ -65,3 +65,15 @@ export const PasswordValidationValues = Object.freeze({
   MIN: 4,
   MAX: 32
 });
+
+export const NotificationFrequency = Object.freeze({
```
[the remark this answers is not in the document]
adamklepacz: I can't. Eslint says "cycle error" and enumerable function doesn't work in this file

server/controllers/userAuth.js
adamklepacz: we decided to not log the errors. There is a task in Jira for coding the error logging mechanism

server/migrations/20191010101148-0.0.9.js
adamklepacz: it will be in the coming task

PR #387, opened 2019-10-13
[EOC-521] Remove user account backend

alemikolo:
**When the user account is removed:**
- controller checks if there are some lists/cohorts when the user is the only owner if so lists of this resources are sent to the client to help the user solve this situation,
- in the next step, user data is removed, his email is hashed and save instead of email,
- all lists/cohorts in which this user is the only viewer/member (there are no other users there) are marked as deleted,
- user is removed from other lists/cohorts (this information is sent via sockets),
- logout event is sent to all user's sockets (devices)
- all users sessions are destroyed,
- User is redirected to account-deleted route

Screenshot 2019-10-14_09-53-38-912

Screenshot 2019-10-13_18-52-29-824

I left comments in `removeAccount` controller to make it easier to understand.

**There were some problems with the sockets.**
- (I didn't make deep investigation because of no time.) When a user has multiple tabs open he doesn't receive data on all of them. It is because there are problems with sharing socket connection between tabs (the new socket is created for a new tab - https://stackoverflow.com/a/41024344). So when we send data to individual sockets (we do this quite often) data is sent only to one/two tabs. There is no problem when we sent data to the rooms.
- Of course, this problem regards only tab but no different devices.
- I use `localstorage` to save some data when the account is deleted and create some listener for `localstorage` changes. Because then all tabs can react to those changes. So if the user deletes the account he will be redirected to account-deleted route on all devices on all tabs.


server/controllers/userAuth.js
```diff
@@ -428,24 +435,147 @@ const checkToken = (req, resp) => {
...
+
+      if (cohortsOnlyOwner.length >= 1 || listsOnlyOwner.length >= 1) {
+        const data = { cohorts: cohortsOnlyOwner, lists: listsOnlyOwner };
+
+        return resp
+          .status(400)
+          .send({ reason: BadRequestReason.REQUIREMENTS, data });
+      }
+
+      // delete account details
```
marekrozmus: extract all those commented blocks to separate functions

server/common/utils/userUtils.js
```diff
@@ -125,11 +128,154 @@ const responseWithUserData = user => {
...
+  if (cohorts.length > 0) {
+    errors.cohorts = cohorts;
+  }
+
+  return errors;
+};
+
+const deleteAccountDetails = async (userModel, user) => {
+  const { _id: userId, createdAt, email } = user;
+  const hashedEmail = bcrypt.hashSync(email, 12);
```
Ciunkos: Please extract `const BCRYPT_SALT_ROUNDS = 12`.

server/common/utils/userUtils.js
```diff
@@ -125,11 +128,154 @@ const responseWithUserData = user => {
...
+  const removedUser = new User({
+    _id: userId,
+    createdAt,
+    email: hashedEmail,
+    updatedAt: new Date()
+  });
+
+  await userModel.deleteOne({ email });
```
Ciunkos: This should be a transaction.

server/common/utils/userUtils.js
```diff
@@ -125,11 +128,154 @@ const responseWithUserData = user => {
...
+        isDeleted: true
+      }
+    )
+    .exec();
+
+const removeUserFromLists = io => async (listModel, displayName, userId) => {
+  const lists = await listModel.find({ viewersIds: userId });
+  const listPromises = [];
+
+  lists.forEach(async list => {
```
Ciunkos: Flatmap this shit - `Promise.all(lists.flatMap(list => ... [list.save(), leaveList(io)(data)]))`

server/common/utils/userUtils.js
```diff
@@ -125,11 +128,154 @@ const responseWithUserData = user => {
...
+    };
+
+    cohortPromises.push(cohort.save(), leaveCohort(io)(data));
+  });
+
+  await Promise.all(cohortPromises);
+};
+
+const destroyUserSessions = async (store, userId) => {
+  const regexp = new RegExp(userId);
```
Ciunkos: Please investigate if regexes can be avoided.

Ciunkos: Otherwise, please escape it.

client/src/modules/user/UserProfile/DeleteDialog.jsx
```diff
@@ -52,6 +56,20 @@ const DeleteDialog = ({
...
 DeleteDialog.propTypes = {
   error: PropTypes.bool,
+  errorData: PropTypes.shape({
+    cohorts: PropTypes.arrayOf(
+      PropTypes.shape({
+        _id: PropTypes.string.isRequired,
+        name: PropTypes.string.isRequired
+      })
+    ),
+    lists: PropTypes.arrayOf(
```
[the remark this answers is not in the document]
alemikolo: This is specific for this one case - removing account an use only in this one component. This 
 aren't ordinary errors but a list of issues that the user has to solve before he will be able to delete his account. If there will be another case that we will need those data we create separate `propType` for that.

client/src/modules/user/UserProfile/ResourcePanel.jsx
```diff
@@ -0,0 +1,80 @@
...
+          resources={lists}
+          route={Routes.LIST}
+        />
+      )}
+    </div>
+  );
+};
+
+ResourcePanel.propTypes = {
+  errorData: PropTypes.shape({
```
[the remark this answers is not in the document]
alemikolo: https://github.com/sandstreamdev/eoc/pull/387#discussion_r334467234

server/common/utils/userUtils.js
alemikolo: We know that, but to enable transactions in Mongo we need replicaSet first

server/common/utils/userUtils.js
alemikolo: we have node 10. and flatMap was added to node 11

client/src/modules/user/model/actions.js
```diff
@@ -9,7 +9,11 @@ import { MessageType as NotificationType } from 'common/constants/enums';
 import { createNotificationWithTimeout } from 'modules/notification/model/actions';
 import { ValidationException } from 'common/exceptions/ValidationException';
 import history from 'common/utils/history';
-import { asyncTypes, enumerable } from 'common/utils/helpers';
+import {
+  accountDeletedRoute,
```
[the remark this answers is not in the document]
alemikolo: we have 4 methods which return specific routes and all of them are placed in one place - in our helper's file. This one is used "localy" in user's actions file and in sockets events handlers.

client/src/modules/user/UserProfile/ResourcePanel.jsx
```diff
@@ -0,0 +1,80 @@
...
+      {cohorts.length > 0 && (
+        <ResourceLinks
+          message={getMessage(cohortContext)}
+          resources={cohorts}
+          route={Routes.COHORT}
+        />
+      )}
+      {lists.length > 0 && (
+        <ResourceLinks
+          message={getMessage(listContext)}
```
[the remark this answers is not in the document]
alemikolo: Yes, but `<ResourceLinks />` is a simple component now, that only receives and displays data without any additional logic and it is not related to external dependencies. 

Both `<ResourcePanel />` and `<ResourceLinks />` need the same data which should be computed twice if I would move computing message to `<ResourceLinks />`

client/src/modules/user/UserProfile/DeleteAccount.jsx
```diff
@@ -47,14 +52,19 @@ class DeleteAccount extends Component {
...
+    } catch (err) {
+      let errorData;
+
+      if (err instanceof RequirementsException) {
+        const { data } = err;
+
+        errorData = data;
+      }
+
+      this.setState({ errorData, isErrorVisible: true, pending: false });
```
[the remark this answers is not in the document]
alemikolo: `errorData` is an object that stores issues (lists and cohorts) that the user has to solve before he will be able to remove his account.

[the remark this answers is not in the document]
Ciunkos: Yet, still, it is an error (error = errorData) in this case. No need for a separate field. All the code has to check if error is present with a simple `if (error)` checks.

client/src/modules/user/model/actions.js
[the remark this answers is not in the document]
Ciunkos: Consider moving them to `routes.js`.

client/src/sockets/enums.js
```diff
@@ -79,4 +80,8 @@ export const CommentEvents = Object.freeze({
 
 export const AppEvents = enumerable()('JOIN_ROOM', 'LEAVE_ROOM');
 
+export const UserEvents = Object.freeze({
```
Ciunkos: Please use `enumerable` helper.

server/common/utils/userUtils.js
Ciunkos: That should be still possible as @kborucinski had implemented transactions already.

server/common/utils/userUtils.js
Ciunkos: Catch some ponyfill then and consider using it -> `const flatMap = f => xs => xs.reduce((ys, y) => ys.concat(f(y)), []);` (included in std https://github.com/sandstreamdev/std/blob/master/array/flatMap.js)

client/src/common/utils/localStorage.js
```diff
@@ -1,18 +1,42 @@
...
+};
+
+export const loadAccountData = () => {
   try {
-    const serializedState = localStorage.getItem('settings');
+    const serializedState = localStorage.getItem(storageKeys.ACCOUNT);
 
     return serializedState === null ? undefined : JSON.parse(serializedState);
   } catch {
     // Ignore write errors
```
Ciunkos: Please keep the single logic for load/save and the just:
```javascript
export const loadAccountData = load(storageKeys.ACCOUNT);
export const saveAccountData = save(storageKeys.ACCOUNT);
export const loadSettingsData = load(storageKeys.SETTINGS);
export const saveSettingsData = save(storageKeys.SETTINGS);
```

where load and save are generalized:

```javascript
export const load = storageKey => () => { ... }
export const save = storageKey => value => { ... }
```

client/src/modules/user/UserProfile/DeleteDialog.jsx
```diff
@@ -34,13 +36,15 @@ const DeleteDialog = ({
       <AlertBox type={MessageType.ERROR}>
         <FormattedMessage id="user.delete-account-warning" />
       </AlertBox>
-      {error && (
+      {error && !errorData && (
         <ErrorMessage
           message={formatMessage({ id: 'user.delete-account-error' })}
         />
       )}
+      {errorData && <ResourcePanel errorData={errorData} />}
```
Ciunkos: Please wrap them both within `{error && <>{...}</>}` section.

client/src/modules/user/UserProfile/ResourceLinks.jsx
```diff
@@ -0,0 +1,37 @@
...
+        </li>
+      ))}
+    </ul>
+  </div>
+);
+
+ResourceLinks.propTypes = {
+  message: PropTypes.string.isRequired,
+  resources: PropTypes.arrayOf(
+    PropTypes.shape({
```
Ciunkos: This type can be reused.

client/src/modules/user/model/actions.js
```diff
@@ -36,6 +44,12 @@ const fetchUserDetailsFailure = () => ({
   type: AuthorizationActionTypes.FETCH_FAILURE
 });
 
+export const removeUserData = () => dispatch => {
+  dispatch(logoutSuccess());
+  history.replace(accountDeletedRoute());
```
Ciunkos: Consider leaving history manipulation as the last step.

server/controllers/userAuth.js
```diff
@@ -428,24 +453,52 @@ const checkToken = (req, resp) => {
...
+
+      await destroyUserSessions(req.sessionStore, userId);
+
+      req.logout();
+      resp.clearCookie('connect.sid');
+
+      return resp.send();
+    }
+    throw new Error();
+  } catch {
```
Ciunkos: Please at least log the error.

server/controllers/userAuth.js
```diff
@@ -428,24 +453,52 @@ const checkToken = (req, resp) => {
...
+
+      await removeUserFromLists(socketInstance)(List, displayName, userId);
+
+      await deleteUserCohorts(Cohort, userId);
+
+      await removeUserFromCohorts(socketInstance)(Cohort, displayName, userId);
+
+      await logoutOtherSessions(socketInstance)(userId);
+
+      await destroyUserSessions(req.sessionStore, userId);
```
adamklepacz: remove empty lines between all these awaits

client/src/modules/user/UserProfile/DeleteDialog.jsx
alemikolo: Not every error comes with errorData.

client/src/modules/user/UserProfile/DeleteDialog.jsx
```diff
@@ -34,13 +36,15 @@ const DeleteDialog = ({
       <AlertBox type={MessageType.ERROR}>
         <FormattedMessage id="user.delete-account-warning" />
       </AlertBox>
-      {error && (
+      {error && !errorData && (
```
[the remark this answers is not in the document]
alemikolo: change `errorData` to `onlyOwnerResources`
```
const authorizationError = error && !onlyOwnerResources;
// it means that you entered wrong credential etc...
```
```
{authorizationError && (
          <ErrorMessage
            message={formatMessage({ id: 'user.delete-account-error' })}
          />
        )}
        {onlyOwnerResources && <ResourcePanel resources={onlyOwnerResources} />}
        <DeleteForm
         (...)
        />
```

client/src/modules/user/UserProfile/DeleteAccount.jsx
```diff
@@ -47,14 +52,19 @@ class DeleteAccount extends Component {
       const result = await deleteAccount(trimmedEmail, password);
 
       if (result) {
-        this.setState({
-          isAccountDeletedDialogVisible: true,
-          isDeleteDialogVisible: false,
-          pending: false
-        });
+        saveAccountData('deleted');
```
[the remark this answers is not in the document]
alemikolo: nothing, this key-value pair `ACCOUNT: DELETED` i save to local storage to allow other browse tabs with eoc to redirect to account-deleted route. I moved it to new enum

PR #389, opened 2019-10-14
[EOC-531] Profile page add email notifications frequency

adamklepacz:
This PR refers to https://github.com/sandstreamdev/eoc/pulls?q=is%3Apr+is%3Aclosed
I have implemented all your comments on this branch. (you can see there which one gets updated and which one I left with just an answer).

I needed to open a new PR instead of a "RE-OPENING" old one, because github somehow doesn't see my last few commits on the old PR. 

Zrzut ekranu 2019-10-15 o 10 29 40
Zrzut ekranu 2019-10-15 o 10 29 35


client/src/locales/en.json
```diff
@@ -82,6 +89,11 @@
   "delete-form.email": "Your email",
   "delete-form.verify-message": "To verify, type {text} below",
   "delete-form.verify-text": "delete my account",
+  "email-notifications.save-settings-failure": "Changing notification frequency failed.",
+  "email-notifications.save-settings-success": "Email notification frequency saved.",
+  "email.notification.heading": "Email notifications frequency",
```
marekrozmus: Change "Email notifications frequency" to "Send summary report". Also "email.notification...." to "email.report... ".

adamklepacz:
> "Send now" button is missing

This button will be implemented on EOC-533 as we decided on a meeting last time. 
One button and data sent via email as JSON.strigify() object

client/src/modules/user/UserProfile/EmailNotifications.jsx
```diff
@@ -0,0 +1,150 @@
...
+  },
+  { message: 'common.friday', value: EmailNotificationsFrequency.FRIDAY },
+  {
+    message: 'common.saturday',
+    value: EmailNotificationsFrequency.SATURDAY
+  },
+  { message: 'common.sunday', value: EmailNotificationsFrequency.SUNDAY }
+];
+
+class EmailNotifications extends PureComponent {
```
alemikolo: `SummaryReportSettings`? This component is not an email notification it allows the user to set up sending reports frequency...

PS. Or if we stay with notification `NotificationsSettings`

client/src/modules/user/UserProfile/EmailNotifications.jsx
```diff
@@ -0,0 +1,150 @@
...
+      <section className="email-notifications">
+        <h2 className="email-notifications__heading">
+          <FormattedMessage id="email.notification.heading" />
+        </h2>
+        <div className="email-notifications__body">
+          <label className="email-notifications__label">
+            <FormattedMessage id="email.notification.never" />
+            <input
+              checked={
+                emailNotificationsFrequency ===
```
alemikolo: create booleans:
```
const sendNever = emailNotificationsFrequency ===  EmailNotificationsFrequency.NEVER;
const sendWeekly = emailNotificationsFrequency !==  EmailNotificationsFrequency.NEVER;
```
and use them :
```
checked={sendNever}
(...)
checked={sendWeekly}
```
or
```
checked={sendNever}
(...)
checked={!sendNever}
```

client/src/modules/user/UserProfile/UserProfile.jsx
```diff
@@ -197,6 +199,7 @@ class UserProfile extends PureComponent {
           {this.renderPersonalInfo()}
           {this.renderContactInfo()}
           {this.renderAccountInfo()}
+          <EmailNotifications user={currentUser} />
```
alemikolo: `EmailNotifications` is a class component, is connected with redux etc. Because it is not a simple "render" component you can get user data inside this component and not pass it from the parent

client/src/modules/user/model/actions.js
```diff
@@ -149,3 +149,26 @@ export const getAccountDetails = token =>
 
 export const deleteAccount = (email, password) =>
   deleteData('/auth', { email, password });
+
+export const saveEmailNotificationSettings = emailNotificationsFrequency => async dispatch => {
```
alemikolo: this is comment only - `reportFrequency` is more adequate, sound shorter and better

client/src/common/constants/enums.js
```diff
@@ -65,3 +65,15 @@ export const PasswordValidationValues = Object.freeze({
   MIN: 4,
   MAX: 32
 });
+
+export const EmailNotificationsFrequency = Object.freeze({
```
alemikolo: why not create by enumerable?

server/migrations/20191010101148-0.0.9.js
```diff
@@ -0,0 +1,44 @@
...
+  const addEmailNotificationSettings = async () =>
+    db.collection('users').updateMany(
+      {
+        emailNotificationsFrequency: { $exists: false },
+        lastEmailNotificationSentAt: { $exists: false }
+      },
+      {
+        $set: {
+          emailNotificationsFrequency: EmailNotificationsFrequency.NEVER,
+          lastEmailNotificationSentAt: ''
```
alemikolo: type of lastEmailNotificationSentAt is Date. When there is no value we should save an empty string or null? OR maybe it is unnecessary to set it now maybe at the first sending we should save it to user doc?

server/models/user.model.js
```diff
@@ -9,6 +11,11 @@ const UserSchema = new Schema(
     avatarUrl: { type: String },
     displayName: { type: String, required: true },
     email: { type: String, required: true },
+    emailNotificationsFrequency: {
+      type: String,
+      default: EmailNotificationsFrequency.NEVER
+    },
+    lastEmailNotificationSentAt: { type: Date },
```
alemikolo: alphabetic order

client/src/common/constants/enums.js
adamklepacz: enumerable doesn't work here 'cycle error' from eslint is displayed.

client/src/modules/user/UserProfile/UserProfile.jsx
adamklepacz: yes I know, but i will have to repeat all fetch logic inside EmailNotification component 
So all code below will be repeated two times. So I think passing user data is a better fit for this situation.


```
import { getCurrentUser } from 'modules/user/model/selectors';

handleFetchUserDetails = () => {
    const { fetchUserDetails } = this.props;

    this.pendingPromise = makeAbortablePromise(fetchUserDetails());
    this.pendingPromise.promise
      .then(() => this.setState({ pending: false }))
      .catch(err => {
        if (!(err instanceof AbortPromiseException)) {
          this.setState({ pending: false });
        }
      });
  };



UserProfile.propTypes = {

  fetchUserDetails: PropTypes.func.isRequired
};

const mapStateToProps = state => ({
  currentUser: getCurrentUser(state)
});

export default connect(
  mapStateToProps,
  { fetchUserDetails }
)(UserProfile);

**Comment by adamklepacz** [comment]
@marekrozmus  @alemikolo  PR is ready for review. I have updated screenshots

File: client/src/locales/en.json [hunk 9]
```diff
@@ -82,6 +89,11 @@
   "delete-form.email": "Your email",
   "delete-form.verify-message": "To verify, type {text} below",
   "delete-form.verify-text": "delete my account",
+  "email-reports.save-settings-failure": "Changing email reports frequency failed.",
+  "email-reports.save-settings-success": "Email reports frequency saved.",
+  "email.reports.heading": "Send summary reports",
```
marekrozmus: singular - "report"

client/src/locales/en.json
```diff
@@ -82,6 +89,11 @@
   "delete-form.email": "Your email",
   "delete-form.verify-message": "To verify, type {text} below",
   "delete-form.verify-text": "delete my account",
+  "email-reports.save-settings-failure": "Changing email reports frequency failed.",
+  "email-reports.save-settings-success": "Email reports frequency saved.",
```
marekrozmus: use common notification like "Settings saved successfully"

PR #397, opened 2019-10-16
Update items sorting in weekly reports email

marekrozmus:


server/common/utils/helpers.js
```diff
@@ -492,6 +492,23 @@ const formatItems = items => list =>
     };
   });
 
+const timestampFromDateString = date =>
+  Math.round(new Date(date).getTime() / 1000);
+
+const reportDataSortFunction = (a, b) => {
```
alemikolo: Is the word 'Function' necessary?

Ciunkos: "Comparator" postfix would be better for this one.

server/common/utils/helpers.js
```diff
@@ -492,6 +492,23 @@ const formatItems = items => list =>
...
+  const aTimestamp = timestampFromDateString(a.requestedAt);
+  const bTimestamp = timestampFromDateString(b.requestedAt);
+  if (aTimestamp === bTimestamp) {
+    return 0;
+  }
+
+  if (aTimestamp < bTimestamp) {
```
Ciunkos: Consider using ternary here, `return aTimestamp < bTimestamp ? 1 : -1`.

server/common/utils/helpers.js
```diff
@@ -492,6 +492,19 @@ const formatItems = items => list =>
     };
   });
 
+const timestampFromDateString = date =>
+  Math.round(new Date(date).getTime() / 1000);
+
+const reportDataSortComparator = (a, b) => {
```
Ciunkos: Consider `reportDateComparator`.

PR #398, opened 2019-10-17
[EOC-552] Remove account that was created with google login

adamklepacz:
Remove flow goes the same as for account created by local strategy. 
If no password is set, no password input is displayed.

Backend logic is checking if the user has password - then it is local strategy.
It also checks if the user has no password and has idFromProvider property which means the account was created by different strategies, in our case via Google Strategy and then it also allows user to remove an account.

### DIFF
--- a/client/src/modules/user/UserProfile/DeleteAccount.jsx
+++ b/client/src/modules/user/UserProfile/DeleteAccount.jsx
@@ -4,6 +4,7 @@ import { injectIntl, FormattedMessage } from 'react-intl';
 import { connect } from 'react-redux';
 import _flowRight from 'lodash/flowRight';
 import _trim from 'lodash/trim';
+import isEmail from 'validator/lib/isEmail';
 
 import { IntlPropType, UserPropType } from 'common/constants/propTypes';
 import {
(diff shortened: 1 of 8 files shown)


client/src/modules/user/UserProfile/DeleteAccount.jsx
```diff
@@ -4,48 +4,61 @@ import { injectIntl, FormattedMessage } from 'react-intl';
...
+      intl: { formatMessage }
     } = this.props;
-    const { verificationText } = this.state;
+    const { email, password, verificationText } = this.state;
     const verificationString = formatMessage({
       id: 'user.delete-form.verify-text'
     });
+    const isEmailValid = isEmail(email) && email === userEmail;
+    const isVerificationStringValid = verificationText === verificationString;
+    const isPasswordValid = isPasswordSet ? validatePassword(password) : true;
```
alemikolo: password validation??? We can't validate password on authorization forms

server/config/auth.js
```diff
@@ -20,6 +20,7 @@ const StrategyType = Object.freeze({
 
 // Use GoogleStrategy to authenticate user
 passport.use(
+  StrategyType.GOOGLE,
```
alemikolo: this change brings nothing now

PR #399, opened 2019-10-17
Update invitation email template and content

marekrozmus:
chrome_2019-10-17_10-47-54
chrome_2019-10-17_10-48-16

### DIFF
--- a/client/src/modules/cohort/index.jsx
+++ b/client/src/modules/cohort/index.jsx
@@ -40,7 +40,11 @@ import Preloader from 'common/components/Preloader';
 import Breadcrumbs from 'common/components/Breadcrumbs';
 import { getCurrentUser } from 'modules/user/model/selectors';
 import { AbortPromiseException } from 'common/exceptions/AbortPromiseException';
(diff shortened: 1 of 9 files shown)


client/src/modules/cohort/index.jsx
```diff
@@ -429,6 +429,10 @@ class Cohort extends PureComponent {
               />
               <div className="cohort__details">
                 <MembersBox
+                  inviteData={{
+                    name,
+                    url: `cohort/${cohortId}`
```
adamklepacz: you can use here our: `routeGenerator(Routes.COHORT, cohortId)` helper from `helpers.js file`

client/src/modules/list/index.jsx
```diff
@@ -377,6 +377,10 @@ class List extends Component {
                   </button>
                   {isMembersBoxVisible && (
                     <MembersBox
+                      inviteData={{
+                        name,
+                        url: `sack/${listId}`
```
adamklepacz: you can use here our: `routeGenerator(Routes.LIST, listId)` helper from `helpers.js file`

PR #400, opened 2019-10-17
[EOC-551] sendgrid sending email notifications implementation

alemikolo:
For this functionality to work, we have to install `agenda` and to add field HOST to our env file


server/tasks/jobs/sendReports.js
```diff
@@ -0,0 +1,44 @@
...
+        { emailReportsFrequency: reportDay },
+        'displayName email'
+      )
+        .lean()
+        .exec();
+      const reportsData = await Promise.all(
+        users.map(async user => getItemsForReport(List, user))
+      );
+
+      await Promise.all(
```
marekrozmus: check `reportsData` results

server/tasks/jobs/sendReports.js
```diff
@@ -0,0 +1,51 @@
...
+    try {
+      const users = await User.find(
+        { emailReportsFrequency: reportDay },
+        'displayName email'
+      )
+        .lean()
+        .exec();
+
+      await Promise.all(
+        users.map(async user => {
```
marekrozmus: change map to for

Jenkinsfile
```diff
@@ -46,6 +46,7 @@ pipeline {
         GOOGLE_CLIENT_SECRET = credentials('GOOGLE_CLIENT_SECRET')
         EXPRESS_SESSION_KEY = credentials('EXPRESS_SESSION_KEY')
         SENDGRID_API_KEY = credentials('SENDGRID_API_KEY')
+        HOST = credentials('HOST')
```
adamklepacz: sort

server/common/utils/helpers.js
```diff
@@ -536,6 +536,46 @@ const prepareRequestedItems = lists => userId => {
...
+
+  if (ownerLists) {
+    data.todos = prepareTodosItems(ownerLists);
+  }
+
+  if (viewerLists) {
+    data.requests = prepareRequestedItems(viewerLists)(userId);
+  }
+
+  reportData.data = data;
```
adamklepacz: wouldnt' be easier to add displayName and receiver to data object on line 542 instead of creating second object, that contains `data` object?  Complicated.

`line 542:  const data = { requests: [], todos: [], displayName, receiver};`
`line 574  remove this line`
`line 576 return data`

server/controllers/userAuth.js
```diff
@@ -511,42 +508,13 @@ const deleteAccount = async (req, resp) => {
...
-      .exec();
-
-    if (ownerLists) {
-      data.todos = prepareTodosItems(ownerLists);
-    }
-
-    if (viewerLists) {
-      data.requests = prepareRequestedItems(viewerLists)(userId);
-    }
+    const reportData = await getItemsForReport(List, user);
```
adamklepacz: unnecessary List argument, you can simplify this by passing just `getItemsForReport(user) `
and within ` helper.js`  file you can also have access to` List.model ` by requiring at the top of the file

server/index.js
```diff
@@ -59,5 +60,7 @@ socket.init(server);
 
 const PORT = 8080;
 
+runAgenda();
```
adamklepacz: what if agenda function throws an error? does it crash the app? 
don't we need any error handler here or inside `runAgenda` function even with  `catch(noOp)`??

server/routes/authorization.js
```diff
@@ -30,7 +30,7 @@ const {
 const {
   sendResetPasswordLink,
   sendSignUpConfirmationLink,
-  sendReport
+  sendReportOnDemand
```
adamklepacz: sort

server/tasks/agenda.js
```diff
@@ -0,0 +1,21 @@
+const Agenda = require('agenda');
+
+const { DB_URL } = require('../common/variables');
+const sendReports = require('./jobs/sendReports');
+const { enumerable } = require('../common/utils/helpers');
+
+const connectionOpts = {
```
adamklepacz: rename to `connectionOptions`

server/tasks/jobs/sendReports.js
```diff
@@ -0,0 +1,49 @@
...
+        try {
+          const report = await getItemsForReport(List, user);
+
+          if (report) {
+            await sendReport(HOST, report);
+          }
+        } catch {
+          // Ignore errors
+        }
+      });
```
adamklepacz: Check if you can get rid of this nested try catch to simplify this a little bit by using our helper fireAndForget()


```
[].forEach(async user => {
  const report = await fireAndForget(getItemsForReport(List, user));

  if (report) {
    fireAndForget(sendReport(HOST, report));
  }
});

```

alemikolo: This doesn't solve the problem because there is no problem with `sendReport` but with `getItemsForReport`. If there is no nested `try-catch` and the `getItemsForReport` function fails the whole forEach will fail. Of course the last `catch` catches the error but we won't send reports to the rest of users. 

We can't add `fireAndForget` to `getItemsForReport` because we wait for the result. 

But I added `fireAndForget(sendReport(HOST, report))` to have additional protection against sending errors.

server/controllers/userAuth.js
alemikolo: Should helpers file be connected with models? I know that this "helper" is closely associated with List but thought that helpers should rely on the params passed in the place of use? What do you think?

adamklepacz: okay you're right. I agree with point 1.

server/controllers/userAuth.js
```diff
@@ -511,42 +508,13 @@ const deleteAccount = async (req, resp) => {
...
 
     /* eslint-disable-next-line no-param-reassign */
-    resp.locales = {
-      data
-    };
+    resp.locales = reportData;
```
adamklepacz: resp.locals

server/mailer/index.js
```diff
@@ -106,24 +106,28 @@ const sendResetPasswordLink = (req, resp) => {
...
 
+  return SendGridMail.send(message);
+};
+
+const sendReportOnDemand = async (req, resp) => {
+  const host = req.get('host');
+
   try {
-    const result = await SendGridMail.send(message);
+    const result = await sendReport(host, resp.locales);
```
adamklepacz: reps.locals

PR #402, opened 2019-10-17
[EOC-544] Add dialogs animations fade in fade out

adamklepacz:


client/src/common/constants/variables.js
```diff
@@ -7,3 +7,4 @@ export const PROJECT_NAME = 'EOC';
 export const REDIRECT_TIMEOUT = 10000;
 export const DISPLAY_LIMIT = 3;
 export const USER_ANONYMOUS = 'user.anonymous';
+export const DIALOG_FADE_TIME = 250;
```
alemikolo: TRANSITION_TIME? ANIMATION_TIME? we will use it in many places

PR #411, opened 2019-10-21
[EOC-417] Proposal add notification additional info when leaving saved data

adamklepacz:
End_of_Coffee_-_tradsack_your_office_inventory

### DIFF
--- a/client/src/common/utils/events.js
+++ b/client/src/common/utils/events.js
@@ -4,9 +4,11 @@ export const attachBeforeUnloadEvent = callback =>
 export const removeBeforeUnloadEvent = callback =>
   window.removeEventListener('beforeunload', callback);
 
-export const handleWindowBeforeUnload = event => {
-  event.preventDefault();
-  // Chrome requires returnValue to be set
-  // eslint-disable-next-line no-param-reassign
-  event.returnValue = '';
+export const handleWindowBeforeUnload = event => isDirty => {
+  if (isDirty) {
+    event.preventDefault();
+    // Chrome requires returnValue to be set
+    // eslint-disable-next-line no-param-reassign
+    event.returnValue = '';
+  }
 };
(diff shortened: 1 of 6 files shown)


client/src/common/utils/events.js
```diff
@@ -0,0 +1,12 @@
+export const attachBeforeUnloadEvent = callback =>
+  window.addEventListener('beforeunload', callback);
```
alemikolo: hmm, if we use sth like this:
```
export const attachBeforeUnloadEvent = (callback = handleWindowBeforeUnload) =>
  window.addEventListener('beforeunload', callback);
```
there will be node need to import `handleWindowBeforeUnload` in ever file where we want to add/remove onBeforeUnload event, but we will be able to pass different callback;

client/src/modules/cohort/components/CohortDescription.jsx
```diff
@@ -43,10 +48,24 @@ class CohortDescription extends PureComponent {
     const {
       details: { description: previousDescription }
     } = previousProps;
+    const { descriptionInputValue } = this.state;
+    const dataHasChanged = previousDescription !== descriptionInputValue;
```
alemikolo: inputValueHasChanged?

client/src/modules/cohort/components/CohortName.jsx
```diff
@@ -50,11 +55,25 @@ class CohortName extends PureComponent {
     const {
       details: { name: previousName }
     } = previousProps;
-    const hasNameBeenChanged = name !== previousName;
+    const { nameInputValue } = this.state;
+    const nameHasChanged = name !== previousName;
+    const dataHasChanged = previousName !== nameInputValue;
```
alemikolo: inputValueHasChanged?

client/src/common/utils/events.js
adamklepacz: @Ciunkos  @marekrozmus  what do you think?

client/src/modules/cohort/components/CohortDescription.jsx
adamklepacz: I think data has changed is understandable enough.

client/src/modules/cohort/components/CohortName.jsx
adamklepacz: I think data has changed is understandable enough.

PR #412, opened 2019-10-21
[EOC-556] Restoring cohort/list doesn't send lists via socket

adamklepacz:


server/controllers/list.js
```diff
@@ -505,6 +505,7 @@ const restoreList = async (req, resp) => {
       .populate('viewersIds', 'avatarUrl displayName _id')
       .populate('items.authorId', 'displayName')
       .populate('items.editedBy', 'displayName')
+      .populate('cohortId', 'name memberIds ownerIds')
```
alemikolo: this line is unneeded because you get for the cohort in a separate query.

adamklepacz: You're wrong I need them.
without this population line it doesn't fix anything

PR #414, opened 2019-10-22
[EOC-561] Remove account that was created with google login (add common dialogs and flows)

adamklepacz:
Zrzut ekranu 2019-10-22 o 11 36 46
Zrzut ekranu 2019-10-22 o 11 35 54

### DIFF
--- a/client/src/modules/user/UserProfile/DeleteAccount.jsx
+++ b/client/src/modules/user/UserProfile/DeleteAccount.jsx
@@ -5,10 +5,7 @@ import { connect } from 'react-redux';
 import _flowRight from 'lodash/flowRight';
 import _trim from 'lodash/trim';
 
-import {
-  IntlPropType,
-  ResourcesDataPropType
-} from 'common/constants/propTypes';
+import { IntlPropType } from 'common/constants/propTypes';
 import {
   checkIfDataLeft,
   deleteAccount,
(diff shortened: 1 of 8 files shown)


server/controllers/userAuth.js
```diff
@@ -465,6 +465,24 @@ const checkToken = (req, resp) => {
     .catch(() => resp.sendStatus(400));
 };
 
+const checkIfDataLeft = async (req, resp) => {
+  const { _id: userId } = req.user;
+
+  try {
+    const errors = await checkIfUserIsTheOnlyOwner(List, Cohort, userId);
```
alemikolo: maybe this `checkIfUserIsTheOnlyOwner` is no longer needed? Before, deleteAccount controller was very complicated and log so I extract all its parts to separate helpers. Now. this is separate controller so all its logic can be written here

adamklepacz: Let leave it as it is. Another layer of polymorphism is expected, especially for these kinds of checks. We never know if we gonna check this information in some other places. Even if not, it will be much easier to test and reason about.

client/src/modules/user/UserProfile/DeleteAccount.jsx
```diff
@@ -130,12 +162,22 @@ class DeleteAccount extends Component {
...
           onEmailChange={this.handleEmailChange}
-          onlyOwnerResources={onlyOwnerResources}
           onPasswordChange={this.handlePasswordChange}
           onSubmit={this.handleDeleteAccount}
           onVerificationTextChange={this.handleVerificationText}
           pending={pending}
         />
+        <Dialog
+          cancelLabel={formatMessage({ id: 'common.button.cancel' })}
+          isVisible={isSelectionDialogVisible}
```
marekrozmus: change naming for this dialog - it is about ownership transfer so maybe OwnershipTransferInfoDialog or sth similar :) - it has nothing to do with selection.

PR #430, opened 2019-10-30
[EOC-577] Slow network support

adamklepacz:


client/src/modules/list/index.jsx
```diff
@@ -181,15 +168,27 @@ class List extends Component {
...
       fetchListData,
       match: {
         params: { id }
       }
     } = this.props;
+    this.setState({ pendingForDetails: true });
 
-    return fetchListData(id);
+    try {
+      const result = await fetchListData(id);
```
alemikolo: There is no need for using result here :-)
```
try {
  await fetchListData(id);
  this.setState({ pendingForDetails: false });
  this.handleBreadcrumbs();
} catch (error) {
(...)
}
```

PR #432, opened 2019-10-30
[EOC-584] Library disappears on click in Safari

alemikolo:


alemikolo:
> I think you should rever to sorting in the selector.
> You can never trust reducers regarding to object order.

For performance reasons, it is better to remove sorting from the selector.  This is a huge list of objects. Each time you download a license, the library selector will sort the entire list...

adamklepacz:
Referring to this post on StackOveflow  https://stackoverflow.com/questions/46228556/how-is-array-sort-performance-affected-by-the-initial-ordering-of-the-input

sorting of 35 000 strings takes about **6 seconds** using native array.sort() method.
In my opinion if you want to keep this in redux as an object, you should still .sort() it in the selector. 

You can even make some performance tests using console.time() and check how much time it takes to use native .sort() or lodash _sort()

adamklepacz:
So are you gonna sort this in selector or not?

PR #439, opened 2019-10-31
[EOC-586] Fix safari bug related with new comment Textarea

adamklepacz:
I needed to unify the comment field to use default placeholder cause safari treats ::before ::after elements placing differently and it was an issue with missing top border on iOS safari. 

Zrzut ekranu 2019-10-31 o 08 43 29
Zrzut ekranu 2019-10-31 o 08 43 17

### DIFF
--- a/client/src/common/components/Forms/Textarea.jsx
+++ b/client/src/common/components/Forms/Textarea.jsx
@@ -1,6 +1,5 @@
 import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
-import classNames from 'classnames';
 
 import './Textarea.scss';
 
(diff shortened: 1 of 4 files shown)


client/src/common/components/Forms/Textarea.jsx
```diff
@@ -29,44 +36,31 @@ class Textarea extends PureComponent {
...
-        )}
         <textarea
           className="ss-textarea__textarea"
           disabled={disabled}
           name={placeholder}
-          onBlur={this.handleBlur}
           onChange={this.handleOnChange}
-          onFocus={this.handleFocus}
-          placeholder={placeholder}
+          placeholder="Add comment"
```
alemikolo: use react-intl

PR #453, opened 2019-11-05
[EOC-615] Fix text alignments in item's confirmation messages

alemikolo:
I didn't remove `<em>` because then we won't be able to use an ellipsis for very long items' names... 

The issue is fixed by adding vertical alignment in CSS.

Screenshot 2019-11-05_11-25-50-927


client/src/common/components/Confirmation.scss
```diff
@@ -26,6 +26,7 @@
     font-style: normal;
     text-overflow: ellipsis;
     white-space: nowrap;
+    vertical-align: bottom;
```
adamklepacz: please check if 
` vertical-align: baseline;`
will display this correctly. If so, remove this line as 'baseline' is a default option.

My suggestion is to remove this property because modifying this property causes a lot of o problems in past in ActivitiesBox component.

alemikolo: `DeleteForm` component was unused. It was used before when we had an old flow for removing an account. Now we send an email so form and this component is unnecessary.

alemikolo: This was my first choice: `vertical-align: baseline`. But it didn't solve the problem. I tried to remove `<em>` tag add `white-space` property etc... But only `vertical-align: bottom` solved the problem.

alemikolo:
Delete Form used when we had an old flow for deleting an account. Now we send an email so it is unneeded.

PR #484, opened 2019-11-11
[EOC-601] Profile view increase delay for showing preloaded animation

adamklepacz:
The issue wasn't related to delay value. Data wasn't cleared after the profile view was left. 
So I've added a method to clear unnecessary data when profile view is left.  Now preloaders work as expected on every network connection speed. Even 1Kb/s .


client/src/modules/user/model/reducer.js
```diff
@@ -21,6 +21,15 @@ const currentUser = (state = null, action) => {
 
       return state;
     }
+    case AuthorizationActionTypes.LEAVE_PROFILE_VIEW_SUCCESS: {
```
alemikolo: We have already action for this - a common action  `clearMetaDataSuccess` maybe we can use it. It Is dispatched on every route change and some reducers react when this action is dispatched. If you use this action only reducer file would be changed. `currentUser` reducer may react on it and remove unnecessary data.

adamklepacz: you're right

PR #521, opened 2019-11-28
[EO-616] RODO add posibility to edit display name from profile view

adamklepacz:
Zrzut ekranu 2019-11-28 o 14 43 59
Zrzut ekranu 2019-11-28 o 14 44 07
Zrzut ekranu 2019-11-28 o 14 44 11
Zrzut ekranu 2019-11-28 o 14 44 26
Zrzut ekranu 2019-11-28 o 14 44 31


adamklepacz:
> Why it isn't done visually like any other edit filed an application? Is this check mark always visible?

We've discussed this with Aleksander. 
I've implemented this as everywhere, so it was just a plain text that can be clicked to edit. But there was few issues with this solution:
1) User can not see that this field can be edited, 
2) Displaying input after clicking need few extra methods like handleBlur, handleFocus, handleClickOutside within UserProfile component which creates a mess in source code of this compoenent,
3) There was a problem with input when validation error was displayed. When validation error was visible and user clicks outside of the input he calls "blur" event on this input so border of the input disappears. But in the handleOutside() method I was checking for an error, if an error exists I was forcing this input to focus() once again to not let the updateName() action run. 
Then there was a situation when the input border was blinking on every outside click. 
4) Without the "check button" submitting this input on mobile would be not intuitive cause "ENTER" key on mobile is not always displayed, so I would have to add extra SUBMIT button only for mobile which causes another complication of code. (apart from the fact, that we have this solution in other places)


I know that this is inconsistent with other places in our app, anyway I've been talking about this many times we should modify those edit field within sack and cohort to this kind of solution with "submit" button always visible or at least some "edit icon" visible, to show the user that those fields are editable. According to this https://webapphuddle.com/inline-edit-design/ article there should be always "edit icon" visible for the user or input should be always visible.

client/src/modules/user/UserProfile/UserProfile.jsx
```diff
@@ -1,29 +1,46 @@
-import React, { PureComponent } from 'react';
+import React, { PureComponent, createRef } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { FormattedDate, FormattedMessage, FormattedTime } from 'react-intl';
 import classNames from 'classnames';
+import isEmpty from 'validator/lib/isEmpty';
+import isLength from 'validator/lib/isLength';
+import _trim from 'lodash/trim';
```
alemikolo: can we use js trim instead of `lodash`? Is `lodash` trim better than the native js method? Przemo said that if possible, it is better to use native methods.

client/src/modules/user/UserProfile/UserProfile.jsx
```diff
@@ -56,10 +73,61 @@ class UserProfile extends PureComponent {
...
+
+  handleNameUpdate = () => {
+    const { errorMessageId, isNameUpdated, userName } = this.state;
+    const {
+      currentUser: { id: userId },
+      updateName
+    } = this.props;
+
+    if (isNameUpdated && !errorMessageId) {
+      updateName(userName, userId);
```
alemikolo: isn't this an async action? Shouldn't we use `then... catch` or `async...await` with  `try-catch` and wait for the response and handle errors?  Now, the state is changed before we get response

client/src/modules/user/UserProfile/UserProfile.scss
```diff
@@ -118,4 +118,92 @@
     margin: 0;
     border-bottom: 1px solid $border-color;
   }
+
+  &__name-item {
+    padding: 24px 8px;
```
alemikolo: we use `four-by()` for this

client/src/modules/user/UserProfile/UserProfile.scss
```diff
@@ -118,4 +118,92 @@
...
+
+    @include breakpoint($medium) {
+      padding: 20px 8px;
```
alemikolo: we use `four-by()` for this

client/src/modules/user/UserProfile/UserProfile.scss
```diff
@@ -118,4 +118,92 @@
...
+    form {
+      width: 100%;
+
+      @include breakpoint($small) {
+        width: auto;
+      }
+    }
+
+    svg {
+      height: 16px;
```
alemikolo: can we use `four-by()` here?

client/src/modules/user/UserProfile/UserProfile.scss
```diff
@@ -118,4 +118,92 @@
...
+  }
+
+  &__name-form {
+    display: flex;
+    align-items: stretch;
+  }
+
+  .error-message {
+    display: block;
+    margin-top: 5px;
```
alemikolo: shouldn't we use multiples of 4?

server/controllers/userAuth.js
```diff
@@ -347,6 +347,26 @@ const updatePassword = (req, resp) => {
     });
 };
 
+const updateName = async (req, resp) => {
+  const { userId, updatedName } = req.body;
+  const { _id } = req.user;
+
+  if (userId !== _id.toString()) {
+    return resp.sendStatus(400);
```
alemikolo: Shouldn't we validate the name on backend also?

PR #540, opened 2019-12-03
[EOC-616] RODO add posibility to edit display name from profile view

adamklepacz:
Zrzut ekranu 2019-12-3 o 11 42 45
Zrzut ekranu 2019-12-3 o 11 43 09
Zrzut ekranu 2019-12-3 o 11 43 46
Zrzut ekranu 2019-12-3 o 11 44 24


client/src/modules/user/UserProfile/UserProfileName.jsx
```diff
@@ -0,0 +1,164 @@
...
+    setName(value);
+    setIsDirty(true);
+  };
+
+  useEffect(() => {
+    validateName();
+  }, [currentName]);
+
+  useEffect(() => {
+    const handleClickOutside = event => {
```
alemikolo: please check if this handler is added when input is visible and focused, and if it is removed on blur

client/src/modules/user/UserProfile/UserProfileName.jsx
```diff
@@ -0,0 +1,164 @@
...
+      if (clickedOutsideInput && clickedOutsideSpan) {
+        handleNameUpdate(event);
+      }
+    };
+
+    if (isFormVisible) {
+      document.addEventListener('click', handleClickOutside);
+    }
+
+    return () => {
```
marekrozmus: move that return to if above

PR #552, opened 2019-12-09
[HOT_FIX] Fix sign in via login and password, fix small pending issue

alemikolo:
I fixed the issue reported by Mariusz. Additionally, I fixed a few small issues related with disabling sign in/sign up buttons.

Change that fixe the main issue is in `client/src/modules/user/model/actions.js` file.

### DIFF
--- a/client/src/modules/user/AuthBox/components/SignInForm.jsx
+++ b/client/src/modules/user/AuthBox/components/SignInForm.jsx
@@ -137,9 +137,9 @@ class SignInForm extends PureComponent {
 
         if (err instanceof UnauthorizedException) {
           newState.signInErrorId = 'user.actions.sign-in.invalid-credentials';
+        } else {
+          newState.signInErrorId = 'common.something-went-wrong';
         }
-
-        newState.signInErrorId = 'common.something-went-wrong';
 
         this.setState(newState);
       }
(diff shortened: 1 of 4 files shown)


client/src/modules/user/AuthBox/components/SignInForm.jsx
```diff
@@ -130,21 +131,19 @@ class SignInForm extends PureComponent {
...
-          }
+    return this.pendingPromise.promise.catch(err => {
+      if (!(err instanceof AbortPromiseException)) {
+        const newState = { pending: false };
 
-          this.setState(newState);
+        if (err instanceof UnauthorizedException) {
+          newState.signInErrorId = 'user.actions.sign-in.invalid-credentials';
+        } else {
+          newState.signInErrorId = 'common.something-went-wrong';
```
adamklepacz: you don't need to nest it here. 

just put it next to each other, also else is not necessary. Cause if the error is not an instance of AbortPromiseException and UnauthoriedException, which means it is ANY other error we need to handle this someway. 

```
 if (!(err instanceof AbortPromiseException)) {
    //....
  }

 if (err instanceof UnauthorizedException) {
   //....
 }

 newState.signInErrorId = 'common.something-went-wrong';
```

alemikolo: This is old logic, only first then was removed... You're right - else is not necessary. But all changes that affect state should be performed  only when an error is not `instanceof AbortPromiseException`. So Id leaves it here. If your sign action fails and you click cancel before this async returns the result you'll get an error: can't change setState on unmounted component...

